PR #14, opened 2025-05-26
chore: add msw

peppescg:
Add MSW, msw-endpoint and fixtures

`Sample config for msw-endpoint`
```

function mockMCPServers() {
  server.use(
    http.get(mswEndpoint("/api/v1beta/servers"), () => {
      return HttpResponse.json(fakeMCPs);
    }),
  );
}
```

### DIFF
--- a/src/mocks/handlers.ts
+++ b/src/mocks/handlers.ts
@@ -6,6 +6,7 @@ import {
   createServerResponseFixture,
   getServerByName,
 } from "./fixtures";
+import { mswEndpoint } from "./msw-endpoint";
 
 export const handlers = [
   http.get("/health", () => {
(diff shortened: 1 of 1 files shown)


kantord:
I have a question. What is the point of the `mswEndpoint` helper? I was looking at the changes, but I cannot find any current usage of it either :thinking: 

could it be reused for some of the current examples?

peppescg:
> I have a question. What is the point of the `mswEndpoint` helper? I was looking at the changes, but I cannot find any current usage of it either 🤔
> 
> could it be reused for some of the current examples?

basically we used it in codegate for mocking a specific response of an endpoint in a test case, it is not implemented cause we don't have components that are calling thv APIs and we don't have tests on that. The idea is to add the all the configuration that we will need in future, and build the UI once we completed the requirements overview. So after tomorrow I would expect we can work on UI stuff too

lxmcgvrn:
> I have a question. What is the point of the `mswEndpoint` helper? I was looking at the changes, but I cannot find any current usage of it either 🤔
> 
> could it be reused for some of the current examples?

It is a utility copied from the Codegate UI repo — basically, it infers a [template literal type](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) based on a path in the OpenAPI spec, so you get type-hints & type-checking. 

We added this after we had some bugs where we were mocking non-existent endpoints (they had been renamed, I think), and it was messing with tests.

src/mocks/handlers.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "./fixtures";
+
+export const handlers = [
+  http.get("/health", () => {
+    return new HttpResponse({
+      status: 204,
+    });
+  }),
+
+  http.get("*/api/v1beta/version", () => {
```
lxmcgvrn: Can we use the `mswEndpoint` helper here? (throughout entire file)

peppescg: right. I added it later but I forgot that I can replace the handlers path

PR #15, opened 2025-05-28
ci: add knip

kantord:
fixes https://github.com/StacklokLabs/toolhive-react/issues/7

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -1,21 +1,21 @@
 export default {
-  "entry": ["src/main.tsx", "src/index.css"],
-  "project": ["src/**"],
-  "ignore": [
+  entry: ["src/main.tsx", "src/index.css"],
+  project: ["src/**"],
+  ignore: [
     "src/common/api/generated/**",
     "src/common/hooks/use-servers.ts",
-    "src/common/lib/utils.ts"
+    "src/common/lib/utils.ts",
   ],
-  "ignoreDependencies": [
+  ignoreDependencies: [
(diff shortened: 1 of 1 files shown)


knip.json
```diff
@@ -0,0 +1,17 @@
+{
+  "entry": ["src/main.tsx"],
+  "project": ["src/**"],
+  "ignore": [
+    "src/common/api/generated/**",
+    "src/common/hooks/use-servers.ts",
```
kantord: note, I had to ignore some files and dependencies that seemed to be legitimately unused, because I didn't want to remove anything that someone just added with a purpose :smile_cat: 

basically all of the ignores, except for the generated files, should be removed once we have some actual code here that would be used somewhere :smiley_cat:

knip.json
```diff
@@ -0,0 +1,18 @@
...
+  "entry": ["src/main.tsx", "src/index.css"],
+  "project": ["src/**"],
+  "ignore": [
+    "src/common/api/generated/**",
+    "src/common/hooks/use-servers.ts",
+    "src/common/lib/utils.ts"
+  ],
+  "ignoreDependencies": [
+    "tailwindcss",
+    "tw-animate-css",
```
kantord: these 2 tailwind dependencies had to be ignored because they are only imported in CSS, and knip does not support CSS entry files

kantord: never mind, it is possible to use a trivial custom compiler for this, instead of relying on an ignore :rocket:

PR #16, opened 2025-05-28
chore: bootstrap tanstack router

### LINKED ISSUE #1: [Task]: Add router

alex-mcgovern:
- configures tanstack router with data fetching using react-query
- adds test utils for testing tanstack router routes
- fixes some nits in MSW configuration to unblock tests
- closes #1

### DIFF
--- a/openapi-ts.config.ts
+++ b/openapi-ts.config.ts
@@ -16,9 +16,5 @@ export default defineConfig({
       name: "@hey-api/typescript",
       enums: false,
     },
-    {
-      asClass: false, // default
-      name: "@hey-api/sdk",
-    },
   ],
 });
--- a/vitest.setup.ts
+++ b/vitest.setup.ts
@@ -14,7 +14,7 @@ afterEach(() => {
 
 beforeAll(() => {
   server.listen({
-    onUnhandledRequest: "warn",
+    onUnhandledRequest: "error",
   });
   client.setConfig({
     baseUrl: "https://foo.bar.com",


src/app/routes/__root.tsx
```diff
@@ -0,0 +1,15 @@
+import type { QueryClient } from "@tanstack/react-query";
```
peppescg, line 1: `__root.tsx` I suppose this is a standard from tanstack?

lxmcgvrn: It is — we're using [file based routing](https://tanstack.com/router/v1/docs/framework/react/routing/file-based-routing)

src/app/__tests__/app.test.tsx
```diff
@@ -1,9 +1,18 @@
...
+import { MOCK_MCP_SERVERS } from "@/common/mocks/fixtures/servers";
+
+const router = createTestRouter(Index);
+
+it("should render list of MCP servers", async () => {
+  renderRoute(router);
+  await waitFor(() => {
+    for (const mcpServer of MOCK_MCP_SERVERS) {
+      expect(screen.queryByText(mcpServer.name)).toBeInTheDocument();
+    }
```
peppescg, line 16: I would prefer to use `it.each` but I think it is for now, it is only a placeholder

lxmcgvrn: It is a placeholder, but will keep it in mind if touching this again 👍🏻

PR #17, opened 2025-05-28
feat: all mcp servers

alex-mcgovern:
- adds the cards to display MCP servers to the index of the app
- TODOs
  - the `query?all=true` param on the `GET /api/v1beta/servers?all=true` is undocumented, so we currently have a `@ts-expect-error` on this - https://github.com/stacklok/toolhive/issues/497
  - the fields in `#/components/schemas/runtime.ContainerInfo` in the OpenAPI JSON are incorrectly cased, so I have manually amended them (seemed the lowest friction option). To that end, I have also split the `generate-client` npm script into `fetch-openapi`, `generate-client` & `generate-client:nofetch` https://github.com/stacklok/toolhive/issues/496
  - there is a very weird bug (I think in the `@hey-api/client-fetch` implementation) when a server response is missing a `Content-Type: application/json` header that is causing the response from `GET /api/v1beta/servers?all=true` to be "stringified". https://github.com/stacklok/toolhive/issues/495

Screenshot 2025-05-28 at 4 07 47 PM

### DIFF
--- a/src/common/mocks/handlers.ts
+++ b/src/common/mocks/handlers.ts
@@ -21,7 +21,9 @@ export const handlers = [
   }),
 
   http.get(mswEndpoint("/api/v1beta/servers"), () => {
-    return HttpResponse.json(serverListFixture);
+    // TODO: Don't stringify after
+    // https://github.com/stacklok/toolhive/issues/495 is resolved
+    return HttpResponse.json(JSON.stringify(serverListFixture));
   }),
 
   http.post(mswEndpoint("/api/v1beta/servers"), async ({ request }) => {


src/app/routes/index.tsx
```diff
@@ -1,24 +1,30 @@
...
 
 export function Index() {
-  const serversQuery = useSuspenseQuery(getApiV1BetaServersOptions());
-  const servers = serversQuery.data.servers;
+  const serversQuery = useSuspenseQuery(
+    // @ts-expect-error - TODO: the `all=true` query param is not documented
+    getApiV1BetaServersOptions({ query: { all: true } }),
+  );
+  // TODO: Why is the JSON stringified?!
+  const parsed: V1ServerListResponse = JSON.parse(serversQuery.data as string);
```
lxmcgvrn, line 22: There is a very weird bug (I think in the `@hey-api/client-fetch` implementation that is causing the response from `GET /api/v1beta/servers` to be "stringified".

A cURL or a naked fetch request from the browser console do not return a string, so I am pretty sure it's a bug in an upstream dependency.

PR #50, opened 2025-05-30
feat: initial lightweight registry page

alex-mcgovern:


src/app/route-tree.gen.ts
```diff
@@ -11,10 +11,17 @@
...
+import { Route as RegistryImport } from "./routes/registry";
 import { Route as IndexImport } from "./routes/index";
 
 // Create/Update Routes
 
+const RegistryRoute = RegistryImport.update({
+  id: "/registry",
+  path: "/registry",
+  getParentRoute: () => rootRoute,
+} as any);
```
peppescg, line 23: we can type the route?

lxmcgvrn: this is auto-generated — I don't even read it :)

samuv:
I think this can be closed, we can iterate from what we have merged in main

PR #58, opened 2025-06-04
experiment: API client through IPC calls

kantord:


samuv:
@kantord can we close this?

kantord:
I suppose, although we might want to implement it sometime

PR #61, opened 2025-06-05
feat(system-tray): add system tray

samuv:
As  per title, very raw initial implementation of the system tray

Screenshot 2025-06-05 at 11 08 20

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -26,7 +26,7 @@ const config: ForgeConfig = {
      * <app>/Contents/Resources/bin/ (macOS)
      * <app>/resources/bin/       (Win/Linux)
      */
-    extraResource: ["bin/"],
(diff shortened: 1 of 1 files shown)


src/system-tray.ts
```diff
@@ -0,0 +1,90 @@
...
+  BrowserWindow,
+} from "electron";
+import path from "node:path";
+import { existsSync } from "node:fs";
+
+export const initTray = ({
+  toolHiveIsRunning,
+}: {
+  toolHiveIsRunning: boolean;
+}) => {
```
peppescg, line 16: nit: does it makes sense to have object with only one param?

samuv: I have a feeling this is going to end up with more logic than expected

PR #65, opened 2025-06-05
feat(auto-launch): add settings action to run the application at OS startup

samuv:
This is the first raw implementation for the auto launch feature

Wording for toast and other fine tuning would be handle in a separated effort

https://github.com/user-attachments/assets/a2ae8d2b-672c-4844-8cda-9ab9f1dce743


src/common/components/settings/settings-dropdown.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  useEffect(() => {
+    async function loadAutoLaunchStatus() {
+      try {
+        if (window.electronAPI) {
+          const status = await window.electronAPI.getAutoLaunchStatus();
+          setIsAutoLaunchEnabled(status);
+        }
+      } catch (error) {
+        console.error("Failed to load auto-launch status:", error);
```
peppescg, line 25: maybe we can a toast error message here?

samuv: I think this is an internal issue, it should happen if the preload script fails to write the electronAPI object to the window

PR #80, opened 2025-06-06
refactor: structure folder grouping into main, rerender, preload folders

peppescg:
Refactored the structure folder:

- `main/`: Contains the code for the Electron main process.
- `preload/`: Contains the preload scripts for the Electron renderer process.
- `renderer/`: Contains the React application for the renderer process. This is where the UI components live.

https://github.com/user-attachments/assets/fafc7e36-204b-491e-8a3d-52cce1c381e5


peppescg:
> Overall makes sense, a couple of idiosyncrasies to my eye:
> 
> * Would we have a slightly easier time configuring tooling (linting, test coverage etc.), if we still had a `src` folder at the root? (e.g. `src/renderer/`, `src/main/` etc)
> * Why have a `src/` folder in `renderer/` but not not in `main/` or `preload/`?
> * What is the benefit of moving config files (`vite.renderer.config.ts` & `openapi-ts.config.ts`) into the `renderer/` directory?
> 
> Screenshot 2025-06-09 at 10 39 39 AM

@alex-mcgovern basically the best practice is use renderer folder for the UI.  iI took inspiration from podman desktop or other oss electron based project.

> Would we have a slightly easier time configuring tooling (linting, test coverage etc.), if we still had a src folder at the root? (e.g. src/renderer/, src/main/ etc)
> Why have a src/ folder in renderer/ but not not in main/ or preload/?

I guess it makes sense to add src/ into `main` and `preload` too 🤔 to be consistent with `renderer`

> What is the benefit of moving config files (vite.renderer.config.ts & openapi-ts.config.ts) into the renderer/ directory?
basically you have different vite config per main/renderer/preload, you can see the vite config in each folders. The openapi config is there cause it is used only in the renderer.

I am going to add src in main and preload too 👍

PR #90, opened 2025-06-09
chore: configure prettier and format

samuv:
This applies a Prettier configuration with a few plugins that are handy for Tailwind. We’ve already used them in previous projects, and I'm also formatting the entire codebase with the new config

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -9,6 +9,6 @@
     "prettier-plugin-classnames",
     "prettier-plugin-merge"
   ],
-  "tailwindFunctions": ["tv", "twMerge", "composeTailwindRenderProps"],
-  "customFunctions": ["tv", "twMerge", "composeTailwindRenderProps"]
+  "tailwindFunctions": ["tv", "twMerge", "cn"],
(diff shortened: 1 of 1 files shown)


.prettierrc
```diff
@@ -1 +1,14 @@
...
+  "singleQuote": true,
+  "tabWidth": 2,
+  "useTabs": false,
+  "plugins": [
+    "prettier-plugin-tailwindcss",
+    "prettier-plugin-classnames",
+    "prettier-plugin-merge"
+  ],
+  "tailwindFunctions": ["tv", "twMerge", "composeTailwindRenderProps"],
+  "customFunctions": ["tv", "twMerge", "composeTailwindRenderProps"]
```
lxmcgvrn: `composeTailwindRenderProps` is from the React Aria starter, so not needed.
`cn` is a function used for composing classnames so prob is required.

```suggestion
  "tailwindFunctions": ["tv", "twMerge", "cn"],
  "customFunctions": ["tv", "twMerge", "cn"]
```

PR #92, opened 2025-06-09
chore: add csp for production env

peppescg:
Add csp only for production env

**Tested it adding some script style and link to html**
Screenshot 2025-06-09 at 15 29 38

### DIFF
--- a/main/src/csp.ts
+++ b/main/src/csp.ts
@@ -13,6 +13,7 @@ const getCspMap = (port: number) => ({
   'manifest-src': "'self'",
   'media-src': "'self' blob: data:",
   'worker-src': "'self'",
+  'child-src': "'none'",
 })
 
 export const getCspString = (port: number) =>


main/src/csp.ts
```diff
@@ -0,0 +1,20 @@
+const cspMap = {
+  'default-src': "'self'",
+  'script-src': "'self'",
+  'style-src': "'self' 'unsafe-inline'",
+  'img-src': "'self' data: blob:",
+  'font-src': "'self' data:",
+  'connect-src': "'self' http://localhost:*",
```
lxmcgvrn: suggestion: we know the port that the thv server is running on in advance — a low-hanging security improvement might be to remove the wildcard and restrict it to this port only.

lxmcgvrn: Here is a PR onto this branch that does this: https://github.com/StacklokLabs/toolhive-react/pull/101

main/src/csp.ts
```diff
@@ -0,0 +1,20 @@
+const cspMap = {
```
lxmcgvrn: I think we can be extra restrictive by setting [`child-src: 'none'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/child-src)

According to the MDN docs, this restricts:
* nested content (`<frame>` & `<iframe>`)
* [web workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API)

We definitely don't want nested content, and web workers in Electron are an [opt-in feature](https://www.electronjs.org/docs/latest/tutorial/multithreading) for the renderer process that use Node.js APIs — so we probably don't want to enable those either, due to the security implications.

Here is a PR that does this: https://github.com/StacklokLabs/toolhive-react/pull/102

peppescg: @alex-mcgovern cause the #102 has conflicts after merge the previous pr I added directly here the change

PR #96, opened 2025-06-09
chore: upgrade thv bin and openapi

peppescg:
After upgrade thv to v0.0.45, there are some breaking changes to endpoint url

**Tested all the fetch and mutations.**

https://github.com/user-attachments/assets/b35f3ba4-2940-492b-87f9-db056a8d56e9

### DIFF
--- a/scripts/update-api.ts
+++ b/scripts/update-api.ts
@@ -3,7 +3,7 @@ import { writeFile } from 'node:fs/promises'
 import * as path from 'node:path'
 import { TOOLHIVE_VERSION } from '../utils/constants'
 ;(async () => {
-  const url = `https://raw.githubusercontent.com/stacklok/toolhive/refs/tags/v${TOOLHIVE_VERSION}/docs/server/swagger.json`
+  const url = `https://raw.githubusercontent.com/stacklok/toolhive/refs/tags/${TOOLHIVE_VERSION}/docs/server/swagger.json`
   const dest = path.resolve('./renderer/src/common/api/openapi.json')
 
   console.log(`Fetching OpenAPI spec from: ${url}`)


kantord:
@peppescg  should this file also be updated? :thinking: 

https://github.com/StacklokLabs/toolhive-react/blob/main/utils/constants.ts#L1


so that the new version is correctly fetched when you run the build? :thinking:

scripts/update-api.ts
```diff
@@ -3,7 +3,7 @@ import { writeFile } from 'node:fs/promises'
 import * as path from 'node:path'
 import { TOOLHIVE_VERSION } from '../utils/constants'
 ;(async () => {
-  const url = `https://raw.githubusercontent.com/stacklok/toolhive/${TOOLHIVE_VERSION}/docs/server/swagger.json`
+  const url = `https://raw.githubusercontent.com/stacklok/toolhive/refs/tags/v${TOOLHIVE_VERSION}/docs/server/swagger.json`
```
lxmcgvrn: There seems to be an additional `v` in the constructed URL.

```suggestion
  const url = `https://raw.githubusercontent.com/stacklok/toolhive/refs/tags/${TOOLHIVE_VERSION}/docs/server/swagger.json`
```

peppescg: it is based on tag not from main, for that reason I added the v
you can see the url 
https://github.com/stacklok/toolhive/blob/v0.0.45/docs/server/swagger.json

peppescg:
> @peppescg should this file also be updated? 🤔
> 
> https://github.com/StacklokLabs/toolhive-react/blob/main/utils/constants.ts#L1
> 
> so that the new version is correctly fetched when you run the build? 🤔

Right, I am using the env var, and I forgot to update it. Thanks.
Updated @kantord

scripts/update-api.ts
lxmcgvrn: Yes but in the constants file it is also prefixed with a v?

So it tries to fetch `vv0.0.45`

```shell
Fetching OpenAPI spec from: https://raw.githubusercontent.com/stacklok/toolhive/refs/tags/vv0.0.45/docs/server/swagger.json
Error: Failed to fetch OpenAPI spec: 404 Not Found
```

lxmcgvrn: Maybe we should consider sanitizing/normalizing it to prevent errors

peppescg: > Yes but in the constants file it is also prefixed with a v?

oh shit, you're right, cause I used the env var I didn't see it...let me fix it.
Fixed

PR #97, opened 2025-06-09
feat(registry): fine tuning and added validation for the form

samuv:
Just simple fine tuning and validation for registry form

https://github.com/user-attachments/assets/56408014-4375-495d-8193-b981c424481e


renderer/src/features/registry-servers/components/form-catalog-creation.tsx
```diff
@@ -52,8 +47,52 @@ export function FormCatalogCreation({
   onOpenChange,
   onSubmit,
 }: FormCatalogCreationProps) {
+  // Create dynamic schema based on server's env vars
+  const formSchema = useMemo(() => {
```
lxmcgvrn, line 51: nit: I don't think the `useMemo` is warranted here :)

samuv: yeah, I used key for the rerendering, I’ll move it outside the component, I’ll refactor it in the next PR

PR #120, opened 2025-06-10
move "remove" option to server card

kantord:
fixes https://github.com/StacklokLabs/toolhive-react/issues/100

addresses the delete part from here: https://github.com/StacklokLabs/toolhive-studio/issues/26

also adds a new hook `useConfirm` which can be used anywhere to show confirm dialogs in a similar way to the `confirm()` that is built into the browser engine, but using react components and shadcn. This was basically ported from the Minder codebase where it worked really well. 

removes the description route as well

https://github.com/user-attachments/assets/7a50591c-61a4-4396-875c-c4227e7657d3

### DIFF
--- a/renderer/src/common/mocks/fixtures/registry_server.ts
+++ b/renderer/src/common/mocks/fixtures/registry_server.ts
@@ -1,0 +1,33 @@
+export const registryServerFixture = {
+  server: {
+    name: 'time',
+    image: 'mcp/time:latest',
+    description:
+      'MCP server for time info and IANA timezone conversions with auto system timezone detection.',
+    transport: 'stdio',
+    permissions: {
+      network: { outbound: { allow_transport: ['tcp'], allow_port: [443] } },
+    },
+    tools: ['get_current_time', 'convert_time'],
+    env_vars: [],
+    args: [],
+    metadata: {
+      stars: 52153,
+      pulls: 5883,
+      last_updated: '2025-06-09T00:23:00Z',
+    },
+    repository_url: 'https://github.com/modelcontextprotocol/servers',
+    tags: [
+      'auto',
+      'available',
+      'configuration',
+      'conversions',
+      'convert_time',
+      'customization',
+      'details',
+      'detection',
+      'example',
+      'examples',
+    ],
+  },
+}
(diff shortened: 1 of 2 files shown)


peppescg:
@kantord mm are you sure we want to remove the detail page? I was thinking to remove the tabs, showing only the mcp server URL, the GH button and the delete 🤔

kantord:
> @kantord mm are you sure we want to remove the detail page? I was thinking to remove the tabs, showing only the mcp server URL, the GH button and the delete 🤔

oh, that was my understanding. let's check with Alex or James.

In any case, I think just for that, we don't need another page. we can put the GH button in the dropdown menu instead.

main/src/main.ts
```diff
@@ -144,6 +145,16 @@ const createWindow = () => {
...
+  // Handle external URLs
+  mainWindow.webContents.setWindowOpenHandler(({ url }) => {
+    // Open external URLs in the default browser
+    if (url.startsWith('http://') || url.startsWith('https://')) {
+      shell.openExternal(url)
+      return { action: 'deny' }
+    }
+    return { action: 'allow' }
+  })
+
```
peppescg, line 157: out of curiosity, in the detail page the link to github worked without this. I am missing something 🤔 ?

kantord: It was sort-of working, because it was opening the page, but it was opening it in Electron, basically opening a new electron window. (At least on Linux, but I'm assuming it's the same behavior across OSs)

This way it actually opens it in the browser, using OS appropriate mechanism the same way any native app would (so using your default browser)

renderer/src/features/mcp-servers/components/card-mcp-server.tsx
```diff
@@ -59,28 +75,126 @@ export function CardMcpServer({
...
+      try {
+        const { data } = await getApiV1BetaRegistryByNameServersByServerName({
+          path: {
+            name: 'default',
+            serverName: name,
+          },
+        })
+
+        // @ts-expect-error - wtf
+        return JSON.parse(data ?? {})
```
peppescg, line 102: don't worry about this, I fixed in toolhive repository and here https://github.com/StacklokLabs/toolhive-studio/pull/125, as soon as we have a new toolhive release, I will merge mine. I am going to rebase main after you merge this and fix it

main/src/main.ts
kantord: https://www.reddit.com/r/electronjs/comments/1al2w67/opening_url_in_default_browser_macos/

peppescg: oh weird in mac it was opening it in the browser 😆

PR #134, opened 2025-06-12
feat(windows): adjust build, tray and action for win32 support

samuv:
This PR introduces Windows build support and relative UI/UX improvements to enhance the application experience.

Key Changes:
- Added Windows builds to release workflow for cross-platform distribution
- Enhanced system tray status management for better user experience
- Improved UI by hiding application menu bar for a cleaner interface
- Fixed Vite renderer configuration issues
- Added missing autoprefixer dependency
- Normalized line endings and formatted configuration files

Screenshot 2025-06-12 090347

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -63,7 +63,7 @@ const config: ForgeConfig = {
       noMsi: true, // Don't create MSI installer
       authors: 'Stacklok Labs',
       description: 'ToolHive Studio - Development Environment',
-      exe: 'ToolHive Studio.exe',
+      exe: 'toolhive-studio.exe',
     }),
     new MakerDMG({}, ['darwin']),
     new MakerZIP({}, ['darwin']),


main/src/system-tray.ts
```diff
@@ -57,19 +112,25 @@ function setupTrayMenu(tray: Tray, toolHiveIsRunning: boolean) {
...
         if (mainWindow) {
+          if (mainWindow.isMinimized()) {
+            mainWindow.restore()
+          }
           mainWindow.show()
           mainWindow.focus()
+          // On Windows, bring window to front
+          if (process.platform === 'win32') {
+            mainWindow.setAlwaysOnTop(true)
+            mainWindow.setAlwaysOnTop(false)
```
peppescg, line 125: 🤔

peppescg: it seems weird but I supposed that it because of win is weird 😆

peppescg:
sorry I merged the renovate PRs 😆

main/src/system-tray.ts
kantord: well, I don't think we _should_ need this

samuv: I will revisit this in a separate effort, it seems to work at the moment 😄

samuv: but the icon doesn’t render very well in the tray, we probably need to generate a better .ico file

kantord: makes sense

PR #180, opened 2025-06-16
chore: tidy up depedencies for hey-api and regenerate

samuv:
After merging [this](https://github.com/StacklokLabs/toolhive-studio/pull/174), hey-api no longer requires the plugin of fetc-api as a dependency, so we removed it. I also encountered a lot of any types in the generated files, so I added a custom ESLint rule to skip those during validation


samuv:
waiting next release, with this change the fetch-api core is generated and we have a type error there

PR #187, opened 2025-06-16
fix: improve fresh install experience

kantord:
This PR:

- Improves installation experience on Linux: now you don't need to have Docker manually installed, it will be installed automatically (.deb or .rpm based distros). This could be made more sophisticated and should still be used in combination with runtime checks (not implemented in this PR) but it's a good iteration (https://stacklok.slack.com/archives/C091GRYAG49/p1750081930467509)
- Fixes a bug where fresh installations were crashing because we were not casting `null` to `[]` properly. This bug was also happening when you deleted the last server. (https://stacklok.slack.com/archives/C091GRYAG49/p1750084048974649?thread_ts=1750081930.467509&cid=C091GRYAG49)

### DIFF
--- a/renderer/src/routes/index.tsx
+++ b/renderer/src/routes/index.tsx
@@ -23,12 +23,11 @@ export const Route = createFileRoute('/')({
 })
 
 export function Index() {
-  const {
-    data: { workloads = [] },
-  } = useSuspenseQuery({
+  const { data } = useSuspenseQuery({
     // @ts-expect-error - https://github.com/stacklok/toolhive/issues/497
     ...getApiV1BetaWorkloadsOptions({ query: { all: true } }),
   })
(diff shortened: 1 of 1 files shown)


renderer/src/routes/index.tsx
```diff
@@ -23,12 +23,11 @@ export const Route = createFileRoute('/')({
 })
 
 export function Index() {
-  const {
-    data: { workloads = [] },
-  } = useSuspenseQuery({
+  const { data } = useSuspenseQuery({
```
samuv, line 26: does this mean that without a running server, the response data is without `workloads`?

kantord: the problem is the opposite, it was returning workloads as `null` instead of empty array

PR #193, opened 2025-06-17
chore: don't format generated files

alex-mcgovern:


kantord:
though they could be formatted using a hook after generation? :thinking:

samuv:
FYI: I was trying to tidy up the dependencies for hey-api with the latest version and encountered a few issues.

https://github.com/StacklokLabs/toolhive-studio/pull/180

I think the output still needs some fine-tuning (like fixing any types and unused variables), but ideally formatting, linting, and type-checking should work

PR #194, opened 2025-06-17
feat: pass secrets when running workloads from registry

alex-mcgovern:
- [x] "inline" save secrets to store on form submission
- [x] map newly saved secrets to the secrets store keys on form submission
- [x] handle the distinction between previously saved secrets & newly entered secrets
- [x] empty states when secrets/env vars are not required
- [ ] fetch secret keys from API (once endpoint is in place)
- [ ] Integration tests at the form level
- [ ] Change the "run from registry" modal to match the "custom server" modal (in terms of size & appearance)

https://github.com/user-attachments/assets/69c24b29-4d07-4505-b438-653a5bf77163

### DIFF
--- a/preload/src/preload.ts
+++ b/preload/src/preload.ts
@@ -16,8 +16,7 @@ contextBridge.exposeInMainWorld('electronAPI', {
   quitApp: () => ipcRenderer.invoke('quit-app'),
 
   // ToolHive port
-  // getToolhivePort: () => ipcRenderer.invoke('get-toolhive-port'),
-  getToolhivePort: () => 8080,
+  getToolhivePort: () => ipcRenderer.invoke('get-toolhive-port'),
 
   // Theme management
   darkMode: {


renderer/src/common/components/secrets/form-combobox-secrets-store.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+type ConstrainedFieldValues = FieldValues & {
+  secrets: {
+    name: string
+    value: {
+      secret: string
+      isFromStore: boolean
+    }
+  }[]
+}
```
lxmcgvrn, line 42: I will come back and use the real retrieved secret values in a follow-up PR.

utils/fetch-thv.ts
```diff
@@ -146,17 +146,17 @@ async function downloadAndExtractBinary(
...
+//   console.log('🔄 Regenerating API client...')
+//   try {
+//     await execFileAsync('pnpm', ['run', 'generate-client'], {
+//       cwd: path.resolve(__dirname, '..'),
+//     })
+//     console.log('✅ API client regenerated successfully')
+//   } catch (error) {
+//     console.warn('⚠️  Failed to regenerate API client:', error)
+//   }
+// }
```
lxmcgvrn, line 159: Temporary until a new release has been cut.

renderer/src/features/registry-servers/components/form-catalog-creation.tsx
```diff
@@ -17,28 +17,191 @@ import {
...
+                  onChange={(e) =>
+                    field.onChange({
+                      secret: e.target.value,
+                      isFromStore: false,
+                    })
+                  }
+                  className="rounded-tr-none rounded-br-none border-r-0 font-mono focus-visible:z-10"
+                  autoComplete="off"
+                  data-1p-ignore
+                  type="password"
```
peppescg, line 123: One thing that I added in secrets was the eye icon to show/hide  password type field, I guess it could makes sense in these use cases

lxmcgvrn: I did think about it, but it's sort of fiddly as when you select a secret from the store, we show a "fake" value as a series of `••••••••` dots (because we don't have access to it) — so I think the logic around whether you could toggle it would become confusing to users.

lxmcgvrn:
> LGTM 🚀 only a suggestion for password type field that we did for secrets too

Will play around with this in a follow up 👍🏻

Issue #19, opened 2025-05-29
[Task] In-app updater for ToolHive


kantord:
As a user
I need to be able to update the application in a straightforward way

samuv:
this is on hold because we need to have a public repo for using the auto updater from electron ([doc](https://www.electronforge.io/advanced/auto-update#open-source-apps-update.electronjs.org))

PR #195, opened 2025-06-17
chore: add GUI for update-on-restart system

kantord:
For details, see discussion: https://stacklok.slack.com/archives/C091GRYAG49/p1749820407691189

This is needed for #19, although some aspects of the auto-update (main process configuration) is still missing. Until those are properly implemented and tested, this is disabled by a "feature flag".

The PR also adds and extremely trivial implementation for feature flags that will also enable us to merge other GUI components that cannot be displayed to users just yet.

See screenshot for the current implementation:

Screenshot_select-area_20250617144619

**note** I don't _love_ the dismiss button and it's design there. It is just how it looks out of the box from sonner, as it is not really designed to show multiple buttons at the same time. to customize the button layout/spacing, it seems like we have to implement a custom renderer from scratch. I decided to not put a lot of time into this for now. We might want to address it in general for all notifications, but it's not priority specifically for this PR, as this would only be seen by each MVP user once (after they reboot the computer or click the restart button, they will see the updated version) 

Also the issues about the positioning of the notification are not in the scope of this PR

### DIFF
--- a/renderer/src/feature-flags/index.ts
+++ b/renderer/src/feature-flags/index.ts
@@ -1,5 +1,7 @@
-import features from './toggles.json'
+import features from './toggles.json' assert { type: 'json' }
 
-export function isFeatureEnabled(feature: string): boolean {
+type FeatureName = keyof typeof features
+
+export function isFeatureEnabled(feature: FeatureName): boolean {
   return features[feature] === true
 }


renderer/src/feature-flags/index.ts
```diff
@@ -0,0 +1,7 @@
+import features from './toggles.json' assert { type: 'json' }
+
+type FeatureName = keyof typeof features
+
+export function isFeatureEnabled(feature: FeatureName): boolean {
```
peppescg, line 5: why the feature flag is not leverage cookies or LS 🤔 ?

kantord: what is LS? :thinking: 

we could leverage cookies, but I just needed something extremely simple. for my current use cases, manually editing the json is enough or even more convenient for cookies.

if we need something more complete in the future, we can add some library or even copy sth from Minder

renderer/src/route-tree.gen.ts
```diff
@@ -8,99 +8,52 @@
 // You should NOT make any changes in this file as it will be overwritten.
 // Additionally, you should also exclude this file from your linter and/or formatter to prevent it from being checked or modified.
 
-// Import Routes
```
samuv, line 11: Why was this file changed?

PR #196, opened 2025-06-17
chore: add logs

kantord:
https://github.com/user-attachments/assets/f1c36c8a-3c73-4e47-a840-fa4a3aef584b

note:

- the API is not yet ready, for this, only mock data is used
- for the same reason there is no refresh button yet, as it would be a lot of useless logic to try to "mock" a refresh logic
- **this feature is disabled using a feature flag** to be enabled when the API is ready and wired up with the component
- figma: https://www.figma.com/design/KuQGsgrNzRCGmSTl520Hwh/Toolhive?node-id=291-1724&p=f&m=dev


renderer/src/features/mcp-servers/components/__tests__/card-mcp-server.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const moreOptionsButton = screen.getByRole('button', {
+    name: /more options/i,
+  })
+
+  await userEvent.click(moreOptionsButton)
+
+  await waitFor(() => {
+    expect(screen.getByRole('menuitem', { name: /logs/i })).toBeVisible()
+  })
+})
```
lxmcgvrn, line 32: thought: All of the setup for the router-level test feels unnecessary for a simple check that a menu item is visible, I would have thought a simple RTL `render` call would be appropriate here?

kantord: yeah, let me think about this :thinking:  I guess the original plan here was to actually test navigation, not just the visibility. testing the visibility alone actually seems a little bit useless

PR #226, opened 2025-06-19
fix(clients): use proper client name for client_type

peppescg:
Screenshot 2025-06-19 at 12 43 19

___
Fixed the workload status removing the restarting status (keep in mind that there is an [issue](https://github.com/stacklok/toolhive/issues/780) on BE openAPI generation)

### DIFF
--- a/renderer/src/features/clients/components/card-client.tsx
+++ b/renderer/src/features/clients/components/card-client.tsx
@@ -11,7 +11,7 @@ import { Switch } from '@/common/components/ui/switch'
 const CLIENT_TYPE_LABEL_MAP = {
   'roo-code': 'Roo Code',
   cline: 'Cline',
-  'vscode-insider': 'VS CodeI nsider',
+  'vscode-insider': 'VS Code Insider',
   vscode: 'VS Code',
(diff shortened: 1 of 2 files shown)


renderer/src/features/clients/components/card-client.tsx
```diff
@@ -7,12 +7,24 @@ import {
 } from '@/common/components/ui/card'
 import { Switch } from '@/common/components/ui/switch'
 
+// There is an issue with openAPI generator in BE, similar issue https://github.com/stacklok/toolhive/issues/780
+const CLIENT_TYPE_LABEL_MAP = {
+  'roo-code': 'Roo Code',
+  cline: 'Cline',
+  'vscode-insider': 'VS Code Insider',
+  vscode: 'VS Code',
```
lxmcgvrn, line 15: Not related to your change, just has bothered me for a while — shouldn't the client say "GitHub Copilot"?

peppescg: 🤔 I guess that in terms of auto-discovery it is vs code, also if the mpc will be available to copilot

PR #227, opened 2025-06-19
feat: orchestrate secret encryption when running server from registry

alex-mcgovern:
> [!NOTE]  
> This is a small subset of the code required to wire up secrets for running MCP servers from the registry. It is presented in isolation for review & feedback. It is not currently connected to anything, so could be safely merged.

Adds "orchestration" code to the "onSubmit" handler when running an MCP server from the registry:
- Secrets are pre-processed & suffixed with an incremented number (compared against an up-to-date list fetched from the server) to prevent naming collisions. The result of this, is that if a user adds the same secret (defined by the name used when running the server, they will end up with duplicates like this:
```shell
❯ ./bin/thv secret list
A new version of ToolHive is available: v0.0.47
Currently running: v0.0.47-42-gda7a604
Available secrets:
  - CONFLUENCE_API_TOKEN
  - CONFLUENCE_API_TOKEN_2
```
- Secrets are saved 1-by-1 in the secrets store over the API (A small, random delay is used so that the user has visibility into the system status)
- The logic for actually running the server, and polling for it's status remains largely unchanged
  
The only meaningful thing to review in this PR is this hook, `useRunFromRegistry`: https://github.com/StacklokLabs/toolhive-studio/blob/fc6262650b6c4eb7f1d43fdf5e0c13150c45cf41/renderer/src/features/registry-servers/hooks/use-run-from-registry.tsx#L1-L577

https://github.com/user-attachments/assets/683e3a4f-554c-49d8-8d58-4e9aa11a7cff

### DIFF
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -24,10 +24,7 @@
     "paths": {
       "@/*": ["./renderer/src/*"]
     },
-    "types": [
-      "vitest/globals",
-      "@testing-library/jest-dom/vitest",
-    ]
+    "types": ["vitest/globals", "@testing-library/jest-dom/vitest"]
   },
   "include": ["forge.env.d.ts", "renderer/src", "preload/src"],
   "exclude": ["node_modules"]
--- a/tsconfig.node.json
+++ b/tsconfig.node.json
@@ -17,7 +17,7 @@
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
     "noUncheckedSideEffectImports": true,
-    "noUncheckedIndexedAccess": true,
+    "noUncheckedIndexedAccess": true
   },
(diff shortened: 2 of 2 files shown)


renderer/src/features/registry-servers/hooks/use-run-from-registry.tsx
```diff
@@ -0,0 +1,577 @@
...
+    }
+    return acc
+  }, [])
+}
+
+/**
+ * Takes all of the secrets from the form and saves them serially to the
+ * secret store. Accepts a `toastId`, which it uses to provide feedback on the
+ * progress of the operation.
+ * // NOTE 1: `thv` appears to hold a lock on the OS keychain while retrieving
```
lxmcgvrn: I actually need to validate this assumption, I saw some errors which led me to this conclusion, but haven't delved into the source code yet.

PR #236, opened 2025-06-19
fix(app): os app title

peppescg:
Screenshot 2025-06-19 at 17 12 42


samuv:
it seems that we need to add a `executableName` with `toolhive-studio` in the forge.config inside `packagerConfig`, spaces are ok only in macos 🤷🏼

peppescg:
> it seems that we need to add a `executableName` with `toolhive-studio` in the forge.config inside `packagerConfig`, spaces are ok only in macos 🤷🏼

yep I pushed it now

Issue #232, opened 2025-06-19
[Bug] Toast notifications collide with important UI elements


alex-mcgovern:
### Task Title

[Task] Toast notifications collide with important UI elements

### Description

Toast notifications (the popups that provide feedback on system state) often collide with important UI elements.

We have tried positioning them on the bottom right — but we often have important actions here (e.g. "form submit" buttons)

We have tried positioning them on the top right — but this will collide with the settings menu, and possibly even with the Windows controls for closing/minimising the window.

### Priority

Medium

### Task Type

Feature

lxmcgvrn:
We discussed this and we think the current implementation is **good enough**™️

PR #237, opened 2025-06-19
fix: scale down root font size by 10%

alex-mcgovern:
Changes the root font-size to 14.4px.
As everything is calculated in rem units, this decreases the size of UI elements across-the-board.
This isn't a panacea for all of our UI issues (see #232) but it should help.

### Before

Screenshot 2025-06-19 at 4 08 50 PM

### After

Screenshot 2025-06-19 at 4 09 13 PM

### DIFF
--- a/renderer/src/index.css
+++ b/renderer/src/index.css
@@ -42,6 +42,7 @@
 }
 
 :root {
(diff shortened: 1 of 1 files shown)


lxmcgvrn:
I will come back and align the titleBar controls on mac in a separate pass.

PR #246, opened 2025-06-20
fix(servers): remove state filters and adjust title size

peppescg:


samuv:
now that we don't have filters, I wonder if `add tool` should be different from secrets `add secret`? it's seems to float around a little bit
Screenshot 2025-06-20 at 14 42 40
Screenshot 2025-06-20 at 14 42 30

PR #265, opened 2025-06-23
feat: add auto-launch on tray system and menu app

peppescg:
Handle and sync auto-launch on:
- menu
- system Tray
- UI settings

https://github.com/user-attachments/assets/7a76a157-d052-4d7a-8da9-4a44d2cc9bc1

### DIFF
--- a/renderer/src/common/components/settings/settings-dropdown.tsx
+++ b/renderer/src/common/components/settings/settings-dropdown.tsx
@@ -1,5 +1,4 @@
 import { SettingsIcon, Check } from 'lucide-react'
-import { useState } from 'react'
 import {
   DropdownMenu,
   DropdownMenuContent,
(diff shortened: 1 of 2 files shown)


renderer/src/common/components/settings/settings-dropdown.tsx
```diff
@@ -15,22 +15,25 @@ export function SettingsDropdown({ className }: { className?: string }) {
...
     }
+  }
 
-    loadAutoLaunchStatus()
-  }, [])
+  const handleDropdownOpenChange = (open: boolean) => {
+    if (open) {
+      loadAutoLaunchStatus()
+    }
+  }
```
lxmcgvrn: nit/suggestion: We could use react-query to simplify the state management here, e.g.

```tsx
useQuery({
  queryKey: ['auto-launch-status'],
  queryFn: window.electronAPI.getAutoLaunchStatus
})
```

Nothing wrong with today's approach, I just think the `setIsLoading` approach has bit me in the butt before.

peppescg: yep definitely better approach 👍

PR #291, opened 2025-06-24
feat(secrets): handle setup secret provider

peppescg:
https://github.com/user-attachments/assets/be9dd8fb-1586-4a9c-add9-e98980d0f18d

### DIFF
--- a/renderer/src/routes/__root.tsx
+++ b/renderer/src/routes/__root.tsx
@@ -11,6 +11,28 @@ import {
 } from '@tanstack/react-router'
 import { TanStackRouterDevtools } from '@tanstack/react-router-devtools'
 import { Toaster } from '@/common/components/ui/sonner'
+import {
+  postApiV1BetaSecretsOptions,
+  getApiV1BetaSecretsDefaultOptions,
+} from '@/common/api/generated/@tanstack/react-query.gen'
+
+async function setupSecretProvider(queryClient: QueryClient) {
+  const createEncryptedProvider = () =>
+    queryClient.ensureQueryData(
+      postApiV1BetaSecretsOptions({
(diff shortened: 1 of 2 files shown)


samuv:
> 1. fetching secrets inline

I forgot about this, yes it would be better to fix it right away then

lxmcgvrn:
> I will put a remainder to revisit this , maybe it would make sense to bring the logic of setup in main


I think at the least, we need to move this logic to the index route

peppescg:
I agree on moving it at index for now...I am not entirely sure it would be in main 🤔 cause I don't see any advantage, the running/stopping server is a different use case IMO involving the entire before the bootstrap. With Secrets I guess it makes sense to have at bootstrap level.

I am going to tackle that right now, in this way we can release

renderer/src/routes/__root.tsx
```diff
@@ -11,6 +11,28 @@ import {
...
+
+  return queryClient
+    .ensureQueryData(getApiV1BetaSecretsDefaultOptions())
+    .then(async (res) => {
+      if (res?.provider_type !== 'encrypted') {
+        await createEncryptedProvider()
+      }
+    })
+    .catch(createEncryptedProvider)
+}
```
lxmcgvrn, line 35: nit: I think this would be better done with the fetch client, if the user nuked the secrets file, or reset the provider with the `thv` CLI, I worry we could get out-of-sync issues here, if our react-query cache has an out-of-date value. It might not be an issue though, so approving anyway.

peppescg: I don't think we can have issue with react-query 🤔 but we can add `staleTime: 0` for the query. I would add it if there will be an edge case.

PR #323, opened 2025-06-25
fix: fine tune labels

peppescg:
Refactor labels for registry, and add aria-label on secret form

Replace input with labels in registry form

Screenshot 2025-06-25 at 17 17 31

### DIFF
--- a/renderer/src/features/registry-servers/components/form-run-from-registry.tsx
+++ b/renderer/src/features/registry-servers/components/form-run-from-registry.tsx
@@ -79,6 +79,7 @@ function SecretRow({
             <div className="relative">
               <FormControl>
                 <Label
+                  htmlFor={`secrets.${index}.value`}
                   className={cn(
                     'text-muted-foreground !border-input h-full items-center font-mono !ring-0',
                     secret.required ? 'pr-8' : ''
@@ -160,6 +161,7 @@ function EnvVarRow({
             <div className="relative">
               <FormControl>
                 <Label
+                  htmlFor={`envVar.${index}.value`}
                   className={cn(
                     `text-muted-foreground !border-input flex h-full items-center gap-1 font-mono
                     !ring-0`


renderer/src/features/secrets/components/dialog-form-secret.tsx
```diff
@@ -127,7 +127,11 @@ function SecretForm({ form, isEditMode, onSubmit, onCancel }: SecretFormProps) {
                 <FormItem>
                   <FormLabel>Secret name</FormLabel>
                   <FormControl>
-                    <Input placeholder="Name" {...field} />
+                    <Input
+                      placeholder="Name"
+                      {...field}
+                      aria-label="Secret key"
```
lxmcgvrn: nit: Do we need this aria-label? The `FormItem` should be taking care of associating the `FormLabel` with the `Input` when you spread in `{...field}`. It also differs from the actual label.

peppescg: yep I don't need it

lxmcgvrn: Yeah under the hood it will associate the label with the form element — when I test things I usually do `screen.getByLabelText` which helps validate that it's working correctly — but either way it's a small nit.

peppescg: fixing it

renderer/src/features/registry-servers/components/form-run-from-registry.tsx
```diff
@@ -74,23 +74,20 @@ function SecretRow({
...
+        render={() => (
           <FormItem>
             <div className="relative">
               <FormControl>
-                <Input
-                  {...field}
-                  autoComplete="off"
-                  tabIndex={-1}
+                <Label
                   className={cn(
```
lxmcgvrn, line 83: nit: I think this `Label` won't be associated with the correct form field — to be honest, I would probably try to collapse this down to a single form field, rather than having a separate one containing only a `Label` — but if you don't want to do that, the change below would fix the nit:

```suggestion
                  htmlFor={`secrets.${index}.value`}
                  className={cn(
```

The same thing applies to the changes in the `EnvVarRow` component too.

peppescg: right also the env var one

PR #328, opened 2025-06-25
fix: show useful error message when OS keyring is not configured

kantord:
fixes: https://github.com/StacklokLabs/toolhive-studio/issues/320

more context: https://stacklok.slack.com/archives/C091GRYAG49/p1750862007547619?thread_ts=1750853359.556549&cid=C091GRYAG49

**Screenshots of this special error screen:**

Screenshot_select-area_20250625190221
Screenshot_select-area_20250625190340

**The old error screen should be unchanged**

Screenshot_select-area_20250625190515
Screenshot_select-area_20250625190553

### DIFF
--- a/renderer/src/common/components/error/base-error-screen.tsx
+++ b/renderer/src/common/components/error/base-error-screen.tsx
@@ -27,15 +27,17 @@ export function BaseErrorScreen({
   }
 
   return (
-    <div className="flex h-[calc(100vh-3rem)] items-center justify-center p-4">
-      <Card className="flex h-full max-h-[600px] w-full max-w-md flex-col shadow-none">
+    <div className="flex h-[calc(100vh-3rem)] items-center justify-center px-4">
+      <Card className="flex max-h-[min(600px,_100%)] w-full max-w-md flex-col shadow-none">
         <CardHeader className="text-center">
           <div className="mb-4 flex justify-center">{icon}</div>
           <CardTitle className="text-xl font-semibold">{title}</CardTitle>
         </CardHeader>
-        <CardContent className="text-muted-foreground flex-1 space-y-4 overflow-y-auto pr-4">
+
+        <CardContent className="text-muted-foreground min-h-0 flex-1 space-y-4 overflow-y-auto pr-4">
           {children}
         </CardContent>
+
         <CardFooter>
           <Button onClick={handleReload} className="w-full">
             <RefreshCw className="mr-2 size-4" />


peppescg:
Why we have the windows and mac errors 🤔 
what about showing this only for linux, having POST error on provider setup, matching the error message ( if the API cannot add other info for better matching), we can leverage the fact that we know the OS too

kantord:
> Why we have the windows and mac errors 🤔 what about showing this only for linux, having POST error on provider setup, matching the error message ( if the API cannot add other info for better matching), we can leverage the fact that we know the OS too

well, in theory the error should only happen on Linux. if you want, I can add this conditional logic :man_shrugging:

kantord:
> Elegant solution, nice job. Let's hope they don't change the wording of that error message underneath us 🤞

yeah I suppose this is the weakest part - it would be nice to have some specific way to handle this, like error codes that act like "a magic string" or some structured error output. Though in the latter case, we also need to make sure that the structured information about the error is properly transferred

samuv:
> > Why we have the windows and mac errors 🤔 what about showing this only for linux, having POST error on provider setup, matching the error message ( if the API cannot add other info for better matching), we can leverage the fact that we know the OS too
> 
> well, in theory the error should only happen on Linux. if you want, I can add this conditional logic 🤷‍♂️

In electronAPI, we have the boolean flags isMac, isWindows, and isLinux, so it should be straightforward to display only the relevant one

kantord:
https://github.com/user-attachments/assets/4df358e1-7420-4173-95fa-c14849d5c26f

@alex-mcgovern, do you like this solution?

kantord:
I added the logic to show it only on Linux - but keep in mind, that apart from manual testing on Linux (custom errors is shown) and the `vitest` test cases using mocks, there is no way to test the behavior on Windows and Mac OS, since we don't know a simple/reasonable way to simulate a bug related to OS keyrigns on those platforms.

So it is a little bit YOLO, we have to trust that it works well on other platforms - still, based on the code it should be fine

lxmcgvrn:
> Record_select-area_20250626125700.mp4 
> @alex-mcgovern, do you like this solution?

It's much better 👍🏻

kantord:
https://github.com/user-attachments/assets/58eda467-3500-4d67-b1b2-a68d97aaf401

tweaked the layout a bit to make sure that it's always properly centered, does not grow very big and does not cause any vertical scroll outside of the card. So, when the text of the card is long and the window of the card is small, there is an internal scroll in the card instead of scrolling the entire page.

Also made it less verbose.

kantord:
Screenshot_select-area_20250626135142


generic error screen seems intact by the refactoring and layout fixes :+1:

PR #334, opened 2025-06-26
feat: allow telemetry opt-out

### LINKED ISSUE #218: [Task] Sentry needs to be opt-out by the user
### Task Title

Sentry opt-in/opt-out by the user

### Description

As a User
I want to disable all the telemetry of the application

### Acceptance Criteria

- [x] opt-in/opt-out sentry tracking

### Priority

Medium

### Task Type

Feature

### Additional Context

_No response_

alex-mcgovern:
Simple opt-out for telemetry, currently only controls Sentry.

* uses [electron store](https://www.npmjs.com/package/electron-store) for storing preferences, this seems like a battle tested options — has options like migrations & discarding invalid config, so can't really go wrong
* IPC handlers to update the store
* react query to manage the client-side state of the users preference

https://github.com/user-attachments/assets/72ac4730-6ca2-4101-915e-32b5187ddd1f

closes #218


main/src/main.ts
```diff
@@ -19,8 +19,15 @@ import net from 'node:net'
...
 
+import Store from 'electron-store'
+
+const store = new Store<{
+  isTelemetryEnabled: boolean
+}>({ defaults: { isTelemetryEnabled: true } })
+
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
+  enabled: store.get('isTelemetryEnabled', true),
```
peppescg: do we need to do that also for renderer init?

lxmcgvrn: I don't believe we do, the render sentry instance just connects to the one in main over IPC

lxmcgvrn: I will verify though

peppescg: I mean here
```
// Sentry setup
Sentry.init({
  // Adds request headers and IP for users, for more info visit:
  // https://docs.sentry.io/platforms/javascript/guides/electron/configuration/options/#sendDefaultPii
  sendDefaultPii: true,
  integrations: [
    Sentry.browserTracingIntegration(),
    Sentry.replayIntegration(),
  ],
  // Set tracesSampleRate to 1.0 to capture 100%
  // of transactions for performance monitoring.
  // We recommend adjusting this value in production
  // Learn more at
  // https://docs.sentry.io/platforms/javascript/configuration/options/#traces-sample-rate
  tracesSampleRate: 1.0,
  // Capture Replay for 10% of all sessions,
  // plus for 100% of sessions with an error
  // Learn more at
  // https://docs.sentry.io/platforms/javascript/session-replay/configuration/#general-integration-configuration
  replaysSessionSampleRate: 0.1,
  replaysOnErrorSampleRate: 1.0,
})
```

cause IMO the rendere is a different tracking, for UI only

lxmcgvrn: Yeah I found a better approach in Github issues for Sentry: https://github.com/StacklokLabs/toolhive-studio/pull/334/commits/804588d51baabce7e86a7aa42a0c466db592fd86

Doesn't actually disable anything, just doesn't send anything

lxmcgvrn: I still think nothing would have happened in the previous iteration, as only the Sentry instance in `main` holds the DSN, the Sentry instance in `renderer` just proxies everything to `main`

peppescg (dismissed):
🚀

peppescg:
one thing that I would like to check, is if there is a possibility to stop only the traces but not the error tracking

lxmcgvrn:
Not merging this for now after discussion with @samuv — we don't see this as essential for the initial open-source release, and he would like to see it featured less prominently in the UI (Error tracking will be important in the initial release phase).

PR #343, opened 2025-06-26
fix(card): replace card style

peppescg:
Screenshot 2025-06-26 at 15 02 31
Screenshot 2025-06-26 at 15 02 25
Screenshot 2025-06-26 at 15 02 18
Screenshot 2025-06-26 at 15 02 11
Screenshot 2025-06-26 at 15 02 03
Screenshot 2025-06-26 at 15 01 57

### DIFF
--- a/renderer/src/routes/__root.tsx
+++ b/renderer/src/routes/__root.tsx
@@ -37,7 +37,7 @@ async function setupSecretProvider(queryClient: QueryClient) {
 function RootComponent() {
   const matches = useMatches()
   const isShutdownRoute = matches.some((match) => match.routeId === '/shutdown')
-  throw new Error('test')
+
   return (
     <>
       {!isShutdownRoute && <TopNav />}


kantord:
sorry, I caused you some conflicts. let me know if you need help resolving them

peppescg:
> tidy job — should we roll the padding/border/shadow changes into the `ui` component instead? Ensure ongoing consistency throughout the app?

yep I also suggested this in the thread 😆 ,but I was thinking that maybe the changes are not definitive and the card could change 🤔 and also I asked to James that IMO the error card should have a different padding so I am waiting a feedback on that.

do you think it is better to change only the error card at this point? also if those padding border are pretty small changes?

peppescg:
> sorry, I caused you some conflicts. let me know if you need help resolving them

no worries mate 👍 I am only waiting a feedback from James before pushing it

lxmcgvrn:
> > tidy job — should we roll the padding/border/shadow changes into the `ui` component instead? Ensure ongoing consistency throughout the app?
> 
> yep I also suggested this in the thread 😆 ,but I was thinking that maybe the changes are not definitive and the card could change 🤔 and also I asked to James that IMO the error card should have a different padding so I am waiting a feedback on that.
> 
> do you think it is better to change only the error card at this point? also if those padding border are pretty small changes?

In my opinion, it's more future-proof to make these styles the "norm" and then the error card the exception, rather than the other way around.

For example, I found a nit with use of the `Switch` component that would be mitigated with this approach (this is the PR, if interested: https://github.com/StacklokLabs/toolhive-studio/pull/345).

peppescg:
> > > tidy job — should we roll the padding/border/shadow changes into the `ui` component instead? Ensure ongoing consistency throughout the app?
> > 
> > 
> > yep I also suggested this in the thread 😆 ,but I was thinking that maybe the changes are not definitive and the card could change 🤔 and also I asked to James that IMO the error card should have a different padding so I am waiting a feedback on that.
> > do you think it is better to change only the error card at this point? also if those padding border are pretty small changes?


> In my opinion, it's more future-proof to make these styles the "norm" and then the error card the exception, rather than the other way around.

yep i agree, ok let me move the changes at ui level

PR #344, opened 2025-06-26
fix: layout shift when scrollbar appears

### LINKED ISSUE #231: [Bug] Layout shift when scrollbar appears
### Task Title

[Task] Layout shift when scrollbar appears

### Description

As per title, there is a layout shift when the scrollbar appears. This normally happens when navigating from a page that requires scrolling to see all content to 1 that is not (or vice versa)

### Acceptance Criteria

_No response_

### Priority

Medium

### Task Type

Feature

### Additional Context

_No response_

### LINKED ISSUE #269: [Bug] Scroll position is not restored upon navigation to a new route
### Bug Description

https://github.com/user-attachments/assets/79b61c25-d7dd-4dd5-8a25-d958f5f98c3a

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Priority

Medium

### Environment

_No response_

### Additional Context

_No response_

alex-mcgovern:
- adds `scrollbar-gutter: stable both-edges` to the `<main>` element — so the scrollbar gutter is always taken into account when calculating layout
- compensates for the extra gutter by reducing the horizontal padding 
- makes the main element a self contained element with a fixed height and it's own scroll — so the scrollbar no longer extends into the app navigation — this means removing the sticky positioning & blur from the nav
- fairly confident there are no weird side effects from the fixed height `<main>` element

closes #231 
closes #269 — this seems to be gone now too

https://github.com/user-attachments/assets/d11e60c9-37a3-445a-8101-277d748a6779

### DIFF
--- a/renderer/src/common/components/layout/top-nav.tsx
+++ b/renderer/src/common/components/layout/top-nav.tsx
@@ -37,8 +37,7 @@ function TopNavContainer(props: HTMLProps<HTMLElement>) {
       {...props}
       className={twMerge(
         props.className,
-        'sticky top-0 z-50',
-        'bg-muted/50 backdrop-blur-2xl',
+        'bg-muted/50',
         'border-mid h-16 border-b',
         'px-6',
         'grid grid-cols-[auto_1fr_auto] items-center gap-7',
--- a/renderer/src/common/components/layout/main.tsx
+++ b/renderer/src/common/components/layout/main.tsx
@@ -6,9 +6,14 @@ export function Main(props: { children: ReactNode; className?: string }) {
     <main
       {...props}
       className={twMerge(
-        'flex min-h-0 w-full flex-1 flex-col px-6 py-4',
+        'flex min-h-0 w-full flex-1 flex-col',
+        'max-h-[calc(100dvh-calc(var(--spacing)_*_16))] overflow-y-auto',
+        'px-3 py-4',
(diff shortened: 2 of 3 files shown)


renderer/src/common/components/layout/top-nav.tsx
```diff
@@ -37,8 +37,7 @@ function TopNavContainer(props: HTMLProps<HTMLElement>) {
       {...props}
       className={twMerge(
         props.className,
-        'sticky top-0 z-50',
-        'bg-muted/50 backdrop-blur-2xl',
+        'bg-muted/50',
```
peppescg, line 40: you removed `backdrop-blur-2xl`

lxmcgvrn: Yes, see the point in the description:
> - makes the main element a self contained element with a fixed height and it's own scroll — so the scrollbar no longer extends into the app navigation — this means removing the sticky positioning & blur from the nav

I guess I should cc James before we merge this

kantord: maybe this would break something else, but I guess we could re-add the blur if we add a negative margin to the main (margin-top: -whatever the navbar height is) and a padding of the same amount.

but maybe this will break the scroller calculation, I don't remember

PR #347, opened 2025-06-26
feat: allow viewing logs

kantord:
fixes https://github.com/StacklokLabs/toolhive-studio/issues/40

### DIFF
--- a/renderer/src/common/components/refresh-button.tsx
+++ b/renderer/src/common/components/refresh-button.tsx
@@ -1,7 +1,7 @@
 import { RefreshCwIcon } from 'lucide-react'
 import { cn } from '../lib/utils'
 import { Button } from './ui/button'
-import { useState, useEffect, ButtonHTMLAttributes } from 'react'
+import { useState, useEffect, type ButtonHTMLAttributes } from 'react'
 
 /**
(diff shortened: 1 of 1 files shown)


renderer/src/features/mcp-servers/sub-pages/logs-page.tsx
```diff
@@ -4,29 +4,28 @@ import { ChevronLeft } from 'lucide-react'
...
-    '[2024-03-20 10:00:04] INFO: Database connection established',
-    '[2024-03-20 10:00:05] INFO: Starting API server...',
-    '[2024-03-20 10:00:06] INFO: API server started on port 8080',
-    `[2024-03-20 10:00:07] INFO: Server ${serverName} is ready to accept connections`,
-    '[2024-03-20 10:00:08] INFO: Health check passed',
-    '[2024-03-20 10:00:09] INFO: Monitoring system initialized',
-  ]
+  const { data: logs, refetch } = useSuspenseQuery({
+    ...getApiV1BetaWorkloadsByNameLogsOptions({ path: { name: serverName } }),
+  })
```
peppescg, line 17: nit: I guess having the loader in the root would be better in our case cause we don't have a spinner component or experience, in this way the data are loaded before the first render of the page

PR #351, opened 2025-06-26
feat: add container engine detection functionality

samuv:
As per the title, this is the initial implementation adding container engine detection functionality. The main process part that checks Docker and Podman will be replaced once we land https://github.com/stacklok/toolhive/issues/699.

Screenshot 2025-06-26 at 18 53 02

https://github.com/user-attachments/assets/35706839-4247-45dd-aa1e-c587343c01db

related to https://github.com/StacklokLabs/toolhive-studio/issues/352

### DIFF
--- a/renderer/src/common/components/error/connection-refused-error.tsx
+++ b/renderer/src/common/components/error/connection-refused-error.tsx
@@ -1,9 +1,10 @@
-import { IllustrationPackage } from '../illustrations/illustration-package'
 import { ExternalLink, RefreshCw, AlertCircle } from 'lucide-react'
-import { BaseErrorScreen } from './base-error-screen'
 import type { ReactNode } from 'react'
 import { useQuery, useMutation } from '@tanstack/react-query'
 import { useEffect } from 'react'
+import { Button } from '@/common/components/ui/button'
+import { BaseErrorScreen } from './base-error-screen'
+import { IllustrationPackage } from '../illustrations/illustration-package'
 import { withMinimumDelay } from './utils'
 
 interface ExternalLinkButtonProps {
@@ -13,21 +14,12 @@ interface ExternalLinkButtonProps {
 }
 
 const ExternalLinkButton = ({ href, children }: ExternalLinkButtonProps) => (
-  <a
-    href={href}
-    target="_blank"
-    rel="noopener noreferrer"
-    className="border-input bg-background ring-offset-background hover:bg-accent
-      hover:text-accent-foreground focus-visible:ring-ring inline-flex w-full
-      items-center justify-between rounded-md border px-3 py-2 text-sm
-      transition-colors focus-visible:ring-2 focus-visible:ring-offset-2
-      focus-visible:outline-none"
-  >
-    <span className="flex items-center">
-      <ExternalLink className="mr-2 size-4" />
+  <Button asChild variant="outline" className="w-full justify-start">
+    <a href={href} target="_blank" rel="noopener noreferrer">
+      <ExternalLink />
       {children}
-    </span>
-  </a>
+    </a>
+  </Button>
 )
 
 export function ConnectionRefusedError() {


renderer/src/common/components/error/connection-refused-error.tsx
```diff
@@ -0,0 +1,155 @@
...
+  // If engines are available, show generic error
+  if (showGenericError) {
+    return (
+      <BaseErrorScreen
+        title="Connection Refused"
+        icon={<AlertCircle className="text-destructive size-12" />}
+      >
+        <p>
+          We're having trouble connecting to ToolHive. This is unexpected and
+          may indicate a service issue.
```
peppescg, line 82: I would add a text saying contact us or something similar with the link to discord channel

renderer/src/common/components/error/connection-refused-error.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <li>Start the container engine</li>
+          <li>Click "Check Again" to continue</li>
+        </ol>
+      </div>
+
+      <div className="space-y-2 pb-4">
+        <Button
+          variant="outline"
+          size="sm"
+          onClick={handleOpenDockerDocs}
```
lxmcgvrn: nit: `window.open` feels unnecessary.

What about:

```ts
<Button asChild>
  <a href="https://foo.bar" target="_blank">
    foo
  </a>
</Button>
```

renderer/src/common/components/error/connection-refused-error.tsx
```diff
@@ -0,0 +1,155 @@
...
+        }
+      } catch (restartError) {
+        console.error('Error restarting ToolHive:', restartError)
+        setIsRestarting(false)
+        setShowGenericError(true)
+      }
+    }, 1200)
+  }
+
+  useEffect(() => {
```
lxmcgvrn, line 65: thought: use of `useEffect` feels a little unecessary, could you just use react-query for the state management? (Yes I know it's a fancy `useEffect` under the the hood)

I'm not sure how that fits with your `withMinimumDelay` fn

e.g.
```ts
const { data: status } = useQuery({
  queryFn: window.electronAPI.checkContainerEngine,
  queryKey: [
    "check-container-engine",
    {
      // ...any other variables that are dependencies and should trigger a refresh
    }
  ]
})

File: renderer/src/common/components/error/index.tsx [hunk 4]
```diff
@@ -16,6 +17,18 @@ export function Error({ error }: ErrorProps = {}) {
...
+  if (
+    error?.toString().includes('ECONNREFUSED') ||
+    error?.toString().includes('Connection refused') ||
+    error?.toString().includes('connect ECONNREFUSED') ||
+    error?.toString().includes('ENOTFOUND') ||
+    error?.toString().includes('Network Error') ||
+    error?.message?.includes('fetch')
+  ) {
+    return <ConnectionRefusedError />
+  }
```
lxmcgvrn, line 30: concern: this feels very hacky. It's good enough for now, but I really want them to work on https://github.com/stacklok/toolhive/issues/699

lxmcgvrn:
Missed that it was still at draft stage, sorry :)

renderer/src/common/components/error/connection-refused-error.tsx
samuv: it makes sense I have added the discord link

renderer/src/common/components/error/connection-refused-error.tsx
samuv: yeah, at first I thought I had to use a specific Electron internal for external links, but that wasn’t the case 😄

I change it

renderer/src/common/components/error/connection-refused-error.tsx
samuv: I gave it a try, but we still need a useEffect to trigger a Toolhive restart

renderer/src/common/components/error/index.tsx
samuv: Yeah, but it’s very specific because it’s related to the route loader errors, and we know it’s an internal API. It’s just temporary in any case

renderer/src/common/components/error/connection-refused-error.tsx
```diff
@@ -0,0 +1,157 @@
...
+const ExternalLinkButton = ({ href, children }: ExternalLinkButtonProps) => (
+  <a
+    href={href}
+    target="_blank"
+    rel="noopener noreferrer"
+    className="border-input bg-background ring-offset-background hover:bg-accent
+      hover:text-accent-foreground focus-visible:ring-ring inline-flex w-full
+      items-center justify-between rounded-md border px-3 py-2 text-sm
+      transition-colors focus-visible:ring-2 focus-visible:ring-offset-2
+      focus-visible:outline-none"
```
lxmcgvrn: tip: looks like you're recreating a button with these styles, you can import the `buttonVariants` function from the button module:

```ts
import { buttonVariants } from '../ui/button'
```

```suggestion
    className={buttonVariants({ variant: "default" })}
```

samuv: Yeah, I cleaned up. I forgot I could use `asChild` to keep it more consistent. The style was Button outline with the icon

lxmcgvrn: Yeah `asChild` is a nice solution as well

PR #355, opened 2025-06-27
feat: improve logs page

kantord:
some touch ups for the visual layout of the logs page. This time it seems like it's not breaking other pages:

https://github.com/user-attachments/assets/cb14efc9-3cae-4e93-881b-3f6ca614cc4e

It also added a minimal highlighting feature using the `<mark />` tag so that it's easy to see why something was included. I didn't add any custom styling, the default yellow looks good to me, but I'm open to changing it.


renderer/src/features/mcp-servers/sub-pages/logs-page.tsx
```diff
@@ -42,23 +68,29 @@ export function LogsPage() {
...
+        <h1 className="m-0 mb-0 p-0 text-3xl font-bold">{serverName}</h1>
+        <Separator />
+        <div className="flex justify-between">
+          <Input
+            className="mb-4 w-full max-w-[250px]"
+            placeholder="Search log"
+            value={search}
+            onChange={(e) => setSearch(e.target.value)}
+            aria-label="Search log"
+          />
```
lxmcgvrn, line 80: I abstracted the search inputs to make the width/icons consistent — you might want to import this here?

```suggestion
          <InputSearch
            onChange={(v) => setSearch(v)}
            value={search}
            placeholder="Filter..."
          />
```

renderer/src/features/mcp-servers/sub-pages/logs-page.tsx
```diff
@@ -8,6 +8,32 @@ import { useSuspenseQuery } from '@tanstack/react-query'
...
+      <mark role="mark" aria-label={innerText} key={idx}>
+        {innerText}
+      </mark>
+    )
+    start = idx + queryLen
+  }
+  if (start < text.length) out.push(text.slice(start))
+
+  return out
+}
```
lxmcgvrn, line 35: really nice touch 👍🏻

renderer/src/features/mcp-servers/sub-pages/logs-page.tsx
```diff
@@ -8,6 +8,32 @@ import { useSuspenseQuery } from '@tanstack/react-query'
...
+
+  const lowerText = text.toLowerCase()
+  const lowerQuery = query.toLowerCase()
+  const queryLen = query.length
+
+  const out: React.ReactNode[] = []
+  let start = 0
+  let idx: number
+
+  while ((idx = lowerText.indexOf(lowerQuery, start)) !== -1) {
```
peppescg, line 22: cool 👍 

nit: I probably would use `const matches = text.split(new RegExp(`(${query})`, 'gi'))` and `matches.map...` replacing the while, but it is not so relevant

kantord:
I will address all the feedback, just wanted to merge a good baseline before the demo

kantord:
> Looks good, although I see some weird text wrapping from some log lines. It looks like this is being returned from the API that way, I wonder how common this is?
> 
> Screenshot 2025-06-27 at 1 14 36 PM

seems like yeah, it is being returned from the API that way, and already printed by the MCP that way :thinking:  not sure if there is anything we can do

PR #364, opened 2025-06-27
feat: view transitions

alex-mcgovern:
As already mentioned, spent some time playing with the View Transition API

One important thing to note is that this functionality is injected via a custom `Link` wrapper, [`LinkViewTransition`](https://github.com/StacklokLabs/toolhive-studio/blob/70ff9397a7648cd20fee28410a9186a40a69d431/renderer/src/common/components/link-view-transition.tsx) — any links that don't use this component won't have view transitions applied (sometimes this is desirable).

All of the config is in CSS variables, so can be fine tuned from here: https://github.com/StacklokLabs/toolhive-studio/blob/70ff9397a7648cd20fee28410a9186a40a69d431/renderer/src/styles/view-transitions.css#L1-L8

https://github.com/user-attachments/assets/c5535244-1b9f-4c8e-b450-a5de2ac8aa67

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "vitest-fail-on-console": "^0.7.1"
   },
   "dependencies": {
+    "@fontsource-variable/inter": "^5.2.6",
     "@fontsource/atkinson-hyperlegible": "^5.2.6",
     "@fontsource/space-mono": "^5.2.7",
     "@hey-api/client-fetch": "^0.13.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,9 @@ importers:
 
   .:
     dependencies:
+      '@fontsource-variable/inter':
+        specifier: ^5.2.6
+        version: 5.2.6
       '@fontsource/atkinson-hyperlegible':
         specifier: ^5.2.6
         version: 5.2.6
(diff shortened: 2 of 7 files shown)


renderer/src/features/mcp-servers/lib/orchestrate-run-custom-server.tsx
```diff
@@ -326,6 +326,7 @@ export async function orchestrateRunCustomServer({
               to="/"
               search={{ newServerName: data.name }}
               onClick={() => toast.dismiss(toastID)}
+              viewTransition={{ types: ['slide-left'] }}
```
lxmcgvrn, line 329: The way `@tanstack/router`'s `Link` is typed prevents use of the `LinkViewTransition` component here — this is a bit of a workaround.

renderer/src/features/registry-servers/lib/orchestrate-run-registry-server.tsx
```diff
@@ -326,6 +326,7 @@ export async function orchestrateRunRegistryServer({
               to="/"
               search={{ newServerName: data.serverName }}
               onClick={() => toast.dismiss(toastID)}
+              viewTransition={{ types: ['slide-left'] }}
```
lxmcgvrn, line 329: The way `@tanstack/router`'s `Link` is typed prevents use of the `LinkViewTransition` component here — this is a bit of a workaround.

renderer/src/common/components/link-view-transition.tsx
```diff
@@ -0,0 +1,45 @@
+import type { FileRouteTypes } from '@/route-tree.gen'
+import { Link, useRouterState } from '@tanstack/react-router'
+import { forwardRef, type ComponentProps } from 'react'
+
+type Route = FileRouteTypes['fullPaths']
+
+const ORDERED_ROUTES: Route[] = ['/', '/registry', '/clients', '/secrets']
```
peppescg, line 7: it is possible to leverage tanstack router foe retrieving the routes
```
const { routesByPath } = useRouter()
const routes = Object.keys(routesByPath) 
```

This will get
```{
    "routesByPath": [
        "/",
        "/clients",
        "/registry",
        "/secrets",
        "/shutdown",
        "/logs/$serverName"
    ]
}
```

Of course you need to use the hook so you need to use it inside the component

lxmcgvrn: Yes but I want them in a specific order, they are still type-safe against the generated routes via the `FileRouteTypes['fullPaths']` type — this is imported from `renderer/src/route-tree.gen.ts`

PR #371, opened 2025-06-30
fix: inter font not loaded

alex-mcgovern:
### Before

Screenshot 2025-06-30 at 10 12 27 AM

### After

Screenshot 2025-06-30 at 10 11 58 AM


kantord:
This is weird. I could swear that it worked on my machine :smiley_cat:  

looks good to me, but let me quickly check that it still works on my computer

PR #376, opened 2025-06-30
feat: implement automatic server restart after app shutdown

samuv:
Automatically restarts servers that were running before app shutdown, improving user experience by preserving server state across app sessions.

- Uses `electron-store` to persist running server list during graceful shutdown
- Implements batch server restart with status polling on app launch
- Adds "restarting" UI status for visual feedback

basic use case:

https://github.com/user-attachments/assets/93d1ce6e-a083-4a8f-bf06-31fe978340d1

edge case:

https://github.com/user-attachments/assets/a7b2ec32-d44c-43fa-81d6-5b5ad481eae8

### DIFF
--- a/main/src/graceful-exit.ts
+++ b/main/src/graceful-exit.ts
@@ -17,7 +17,7 @@ const shutdownStore = new Store({
 })
 
 /** Get the currently running servers from the ToolHive API. */
-async function getRunningServers(port: number): Promise<string[]> {
+async function getRunningServers(port: number): Promise<WorkloadsWorkload[]> {
   const client = createClient({ baseUrl: `http://localhost:${port}` })
   try {
     const response = await getApiV1BetaWorkloads({ client })
@@ -25,12 +25,9 @@ async function getRunningServers(port: number): Promise<string[]> {
       console.warn('No workloads data in API response')
       return []
     }
-    return response.data.workloads
-      .filter(
-        (server: WorkloadsWorkload) =>
-          server.status === 'running' && server.name
-      )
-      .map(({ name }: WorkloadsWorkload) => name as string)
+    return response.data.workloads.filter(
+      (server: WorkloadsWorkload) => server.status === 'running' && server.name
+    )
   } catch (error) {
     console.error('Failed to get running servers:', error)
     return []
@@ -70,7 +67,10 @@ export async function stopAllServers(
 ): Promise<void> {
   const client = createClient({ baseUrl: `http://localhost:${port}` })
   const servers = await getRunningServers(port)
-  console.info(`Found ${servers.length} running servers:`, servers)
+  console.info(
+    `Found ${servers.length} running servers:`,
+    servers.map((item) => item.name)
+  )
 
   if (!servers.length) {
     console.info('No running servers – teardown complete')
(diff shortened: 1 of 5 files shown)


jtenniswood:
Looks great.
I wonder the notification should say 'Restarting servers, this can take some time.'?

samuv:
> Looks great. I wonder the notification should say 'Restarting servers, this can take some time.'?

yeah added your suggestion

main/src/main.ts
```diff
@@ -169,6 +173,11 @@ function createWindow() {
     )
   }
 
+  // Open developer tools at startup in development
+  if (isDevelopment) {
```
kantord, line 177: do we need this?

kantord: I mean do we need to automatically start devtools?

main/src/main.ts
```diff
@@ -347,3 +356,13 @@ ipcMain.handle('restart-toolhive', async () => {
     }
   }
 })
+
+// Shutdown store IPC handlers
+ipcMain.handle('shutdown-store:get-last-servers', () => {
+  return getLastShutdownServers()
+})
+
+ipcMain.handle('shutdown-store:clear-history', () => {
```
kantord, line 365: I have been looking for where we clear the shutdown history, but I cannot find it :thinking:

kantord: never mind, I suppose it's this one 

https://github.com/StacklokLabs/toolhive-studio/pull/376/files#diff-23cbcc91c3896efef71f7c92ab97838e4e8d889effa97f9d393f3406ae46b2e8L48

but I'm not entirely sure why it's necessary to clear the history if we are going to overwrite it anyways :thinking:

samuv: yeah is remapped on clearShutdownHistory ipc handle
This is what we check on the installed page to trigger the restart. So after the mutation, we clear out the state because we don’t need it anymore.

main/src/main.ts
samuv: It’s the only way to check the initial fetch calls. Maybe we can add something to override it with an environment variable

main/src/graceful-exit.ts
```diff
@@ -4,24 +4,30 @@ import {
...
 
 const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
 
+// Create a store instance for tracking shutdown servers
+const shutdownStore = new Store({
+  name: 'server-shutdown',
+  defaults: {
+    lastShutdownServers: [],
+  },
+})
```
lxmcgvrn, line 17: A thought — because electron-store seems to have pretty good support for things like migrations, and maintaining health of your store over time — perhaps just "one store" is a simple mental model, and might be simpler to extend in future?

e.g. if we added new features over time to shutdown/restart, like:
- we want to add granular opt-in to this feature ("only restart server X and server Y at launch")
- we want to make this feature toggleable on/off
- ... etc

or if we start adding other features, like:
- telemetry opt-in/out
- favorites
- ... etc

I guess I kinda view it as a DB / redux store, and expect it to be a singleton, with partitioning internal to the store (i.e. tables/slices)

lxmcgvrn: Not a blocker just a thought

samuv: I didn’t put too much thought into it, but it makes sense to start a conversation about having a global, persistent state. Let’s do that after the MVP

PR #377, opened 2025-06-30
chore: update app & system tray icons

alex-mcgovern:
- Sets up a repeatable workflow to auto-generate `.icns` & `.ico` icons with an npm script: `pnpm run generate-icons`
- Adds the updated branding icons to the project, including the updated system tray icon
- Documents the formats required for updating the icons — these have also been shared with @jtenniswood — you can find the documetation [here](https://github.com/StacklokLabs/toolhive-studio/blob/chore/update-icons/docs/icons.md)
 
Still need to do:
- installer icons
- disk image background
- logo in-app 

---

NOTE: I am using `tsx` as a devDependency because it's a lot more robust & better at handling incorrectly bundled libraries (e.g. `icon-gen`)

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -236,7 +236,7 @@ importers:
         version: 26.1.0
       knip:
         specifier: ^5.58.1
-        version: 5.61.2(@types/node@22.15.34)(typescript@5.8.3)
+        version: 5.61.3(@types/node@22.15.34)(typescript@5.8.3)
       lint-staged:
         specifier: ^16.0.0
         version: 16.1.2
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -74,6 +75,8 @@
     "prettier-plugin-tailwindcss": "^0.6.12",
     "tar": "^7.4.3",
     "ts-node": "^10.9.2",
+    "tsx": "^4.20.3",
```
samuv, line 78: We already had ts-node, so let’s just pick one or the other. I added ts-node for update-api, but I’m totally fine with either. If we want to switch to tsx. Hopefully from Node.js version 24 we can drop those kinds of things altogether because TypeScript will be natively supported

PR #378, opened 2025-06-30
chore: add most of the required code for auto-update

### LINKED ISSUE #19: [Task] In-app updater for ToolHive

kantord:
I wanted something more sophisticated for testing releases features; alas, the electron tooling seems to not help or even actively prevent this... And even after fighting through all the things, I cannot test it on Linux, so really the only way to test update-related feature is to simulate the update without actually using the update system.

There is still value to this, because there are a few components included in wiring the auto update system to the GUI. So I added that wiring + a simple way to simulate an update totally bypassing the actual update system. This way, changes to the update GUI can be debugged without going completely crazy.

fixes #19


peppescg:
we will remove the mock once we have the repo public right?

kantord:
> we will remove the mock once we have the repo public right?

no need to remove the mock. it does not interfere with actual production builds. so it should be useful long term for testing changes to the auto-update experience without having to cut a release

PR #382, opened 2025-07-01
chore: small fixes for .deb

kantord:
this adds 2 fixes:

- add a description to package.json - because this is actually [displayed to people](https://stacklok.slack.com/archives/C072SGY78TS/p1751364699004939) who install ToolHive Studio through the GUI on Ubuntu
- Remove docker dependency for .deb. This is because it turned out, docker packages are a total mess on ubuntu, there are people using unofficial packages and whatnot. So unfortunately it's better to let users use whatever they have on their machines than to do everythign nicely.

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -119,7 +119,7 @@ const config: ForgeConfig = {
         productName: 'ToolHive Studio',
         genericName: 'ToolHive Studio',
         icon: './icons/icon.png',
-        depends: ['docker.io (>= 20.10)'],
+        depends: [],
         maintainer: 'Stacklok Labs',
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "toolhive-studio",
   "productName": "ToolHive Studio",
   "version": "0.0.5",
-  "description": "My Electron application description",
+  "description": "ToolHive Studio is an application that allows you to install, manage and run MCP servers and connect them to AI agents",
```
samuv, line 5: Yeah, this is used on Windows too, I was about to change it myself

kantord: cool

PR #385, opened 2025-07-01
fix: add navigation buttons on error screens

kantord:
fixes https://stacklok.slack.com/archives/C072SGY78TS/p1751364603444459

before:

VirtualBox_ubuntu_01_07_2025_12_07_11

after:

VirtualBox_ubuntu_01_07_2025_17_19_38

### DIFF
--- a/renderer/src/common/components/layout/top-nav/minimal.tsx
+++ b/renderer/src/common/components/layout/top-nav/minimal.tsx
@@ -4,7 +4,7 @@ import { WindowControls } from './window-controls'
 
 export function TopNavMinimal() {
   return (
-    <TopNavContainer className="!flex justify-between">
+    <TopNavContainer className="fixed top-0 !flex justify-between">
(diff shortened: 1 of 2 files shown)


renderer/src/common/components/layout/top-nav/minimal.tsx
```diff
@@ -0,0 +1,12 @@
+import { TopNavContainer } from './container'
+import { TopNavLogo } from './logo'
+import { WindowControls } from './window-controls'
+
+export function TopNavMinimal() {
+  return (
+    <TopNavContainer className="fixed top-0 !flex justify-between">
```
samuv, line 7: `!flex` ?

kantord: the underlying component has a grid layout which I didn't want to change, but it's not the most convenient layout to use here. So this effectively overrides the default layout of that component.

PR #386, opened 2025-07-01
docs(contributing): add dedicated file and DCO

samuv:
as per title, added the section for contributing and DCO


CONTRIBUTING.md
```diff
@@ -0,0 +1,112 @@
+# Contributing to ToolHive <!-- omit from toc -->
```
peppescg: we should have ToolHive studio?

samuv: Yeah, I tried to keep it general and not too specific in some cases, but here it’s better to be clear

PR #387, opened 2025-07-01
feat(telemetry): add page_view and event to traces

peppescg:
In order having a bit of color on user journey, and performance, considering that we don't have an available telemetry service, we can use Sentry traces (dashboard [here ](https://stacklok.sentry.io/traces/?project=4509468911796224&query=analytics.source%3Atracking&statsPeriod=14d)), creating a span `Sentry.startSpan()`.

**trackEvent**
`trackEvent(eventName, data)` is used for generic user events like clicks and actions. It creates a span with op: 'user.event' and automatically adds analytics metadata like `analytics.source: 'tracking', analytics.type: 'event',`.
We can use  `analytics.source is 'tracking'` filter in traces for filtering only internal tracking. 

**trackPageView**
`trackPageView(pageName, data)` is used for page navigation tracking. It creates a span with op: 'page_view' and adds  metadata. 

Screenshot 2025-07-01 at 14 50 26

___
**Set toolhive-studio headers**

Screenshot 2025-07-09 at 12 39 18

### DIFF
--- a/main/src/util.ts
+++ b/main/src/util.ts
@@ -16,7 +16,8 @@ function getVersionFromGit(): string {
         stdio: 'pipe',
       }).trim()
 
-      return describe.replace(/^v/, '')
+      const version = describe.replace(/^v/, '').split('-')[0]
+      return version ?? app.getVersion()
     } catch {
       return app.getVersion()
     }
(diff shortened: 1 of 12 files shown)


peppescg:
Related to:
https://github.com/StacklokLabs/toolhive-studio/issues/359
https://github.com/StacklokLabs/toolhive-studio/issues/360
https://github.com/stacklok/toolhive-studio/issues/462

kantord (approved):

PR #390, opened 2025-07-01
fix: margin main content and logs

peppescg:
**Fix margin main content**
Screenshot 2025-07-01 at 17 31 43

### DIFF
--- a/renderer/src/features/mcp-servers/sub-pages/logs-page/index.tsx
+++ b/renderer/src/features/mcp-servers/sub-pages/logs-page/index.tsx
@@ -34,7 +34,7 @@ export function LogsPage() {
       <div className="mb-2">
         <LinkViewTransition to="/">
           <Button
-            variant="secondary"
+            variant="ghost"
             aria-label="Back"
             className="text-muted-foreground"
           >


renderer/src/features/mcp-servers/sub-pages/logs-page/index.tsx
```diff
@@ -30,15 +30,15 @@ export function LogsPage() {
     : logLines
 
   return (
-    <div className="container mx-auto flex max-h-full flex-1 flex-col p-4">
+    <div className="flex max-h-full w-full flex-1 flex-col">
       <div className="mb-2">
         <LinkViewTransition to="/">
           <Button
-            variant="ghost"
+            variant="secondary"
```
samuv: I don't see the cursor on this button 🤔

peppescg: the button by default don't have the cursor

peppescg: I was thinking it was intentional

peppescg: @jtenniswood all the button don't have by default the cursor pointer...is it intentional?

lxmcgvrn: Either way, this should be:

```tsx
<Button asChild>
  <LinkViewTransition>
  {/* ... */ }
  </LinkViewTransition>
</Button>
```

As a `<button>` is not a valid descendant of an `<anchor>` (You will be able to focus them both independently).

Not related to your change, but I think we should tidy it up.

PR #393, opened 2025-07-02
fix: remove pacakge lock and use strokeWidth instead of stroke-width in logo svg

samuv:
as per title


samuv:
> Oops, embarassing misses, thank you for catching

No worries, I mistakenly ran `npm i` all the time 😄

PR #409, opened 2025-07-03
feat: confirm before closing ToolHive studio along with all MCPs

### LINKED ISSUE #407: [Bug] Add a shutdown confirmation dialog to confirm MCP server shutdown on quit
### Bug Description

Add a dialog panel to confirm that users are aware that shutting down ToolHive will result in the shutdown of their MCP servers. Ideally add a 'do not show this message again' option

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Priority

Medium

### Environment

_No response_

### Additional Context

_No response_

kantord:
https://github.com/user-attachments/assets/0d7e7859-9c90-4e63-b35c-f7be2aec1ffa

fixes #407

also adds a bunch of tests for useConfirm, since the logic is getting a bit more complex now

when quitting through the system tray, the app window is displayed in order to show the confirmation dialog


jtenniswood:
Awesome, nice work

renderer/src/common/components/layout/top-nav/quit-confirmation-listener.tsx
```diff
@@ -0,0 +1,21 @@
...
+      const confirmed = await confirmQuit()
+      if (confirmed) {
+        await window.electronAPI.quitApp()
+      }
+    })
+
+    return cleanup
+  }, [confirmQuit])
+
+  // This component doesn't render anything visible
```
peppescg, line 19: cause it is not render anything you decided to have this component for readability of the TopNav component?

peppescg (approved):
nice 🚀

renderer/src/common/components/layout/top-nav/quit-confirmation-listener.tsx
kantord: not sure I fully understand the question.

I put it int the topnav component, because the topnav component is always there when the app is not in a broken state, which is the only case when the confirmation is necessary. btw this is only necessary when closing the app from the tray icon

PR #415, opened 2025-07-03
fix: ensure toolhive is running before calling api

peppescg:
It is important to run the API after toolhive process is running, so I added a ipc for checking if thv process is running, and call it in `beforeLoad`, this callback will be called before the loader, ensuring the thv process is running and the health is 200

____

the forge.config for MakerDMG needs to include name of the package in order to run the build locally

### DIFF
--- a/renderer/src/common/components/error/index.tsx
+++ b/renderer/src/common/components/error/index.tsx
@@ -17,6 +17,8 @@ export function Error({ error }: ErrorProps = {}) {
     return <KeyringError />
   }
 
+  console.log('error', error)
+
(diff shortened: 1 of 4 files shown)


renderer/src/common/components/error/index.tsx
```diff
@@ -17,6 +17,8 @@ export function Error({ error }: ErrorProps = {}) {
     return <KeyringError />
   }
 
+  console.log('error', error)
```
samuv, line 20: leftover?

peppescg: lol

PR #422, opened 2025-07-04
chore(log): add electron-log in order to troubleshooting issues

peppescg:
Used [electron-log](https://github.com/megahertz/electron-log) replacing `console` with `log` fn.

The log will be saved into the filesystem:

- on Linux: ~/.config/{app name}/logs/main.log
- on macOS: ~/Library/Logs/{app name}/main.log
- on Windows: %USERPROFILE%\AppData\Roaming\{app name}\logs\main.log

___
nit: we can improve or replace later, but we got some issues using directly the dmg download from github repo, that we cannot reproduce locally, so it would helpful to have some additional info here.

`Logs with info|warn|error`

Screenshot 2025-07-04 at 15 09 55

`Logs in both thv serve and logs file`

Screenshot 2025-07-04 at 15 07 05

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -31,6 +31,7 @@ import {
   isToolhiveRunning,
   binPath,
 } from './toolhive-manager'
+import log from './logger'
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
@@ -40,8 +41,8 @@ Sentry.init({
 declare const MAIN_WINDOW_VITE_DEV_SERVER_URL: string | undefined
 declare const MAIN_WINDOW_VITE_NAME: string
 
-console.log(`ToolHive binary path: ${binPath}`)
-console.log(`Binary file exists: ${existsSync(binPath)}`)
+log.info(`ToolHive binary path: ${binPath}`)
+log.info(`Binary file exists: ${existsSync(binPath)}`)
 
 // this implements auto-update
 updateElectronApp()
@@ -56,7 +57,7 @@ app.on('ready', () => {
       return
     }
 
-    console.debug('Simulating a new release for testing purposes')
+    log.info('Simulating a new release for testing purposes')
     mainWindow.webContents.send('update-downloaded')
   }, 2000)
 })
@@ -66,13 +67,12 @@ autoUpdater.on('update-downloaded', () => {
     return
   }
 
-  console.log('Update downloaded — sending to renderer')
+  log.info('Update downloaded — sending to renderer')
   mainWindow.webContents.send('update-downloaded')
 })
 
 autoUpdater.on('error', (message) => {
-  console.error('There was a problem updating the application')
-  console.error(message)
+  log.error('There was a problem updating the application: ', message)
 })
 
 autoUpdater.checkForUpdates()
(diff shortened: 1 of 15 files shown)


utils/constants.ts
```diff
@@ -1 +1 @@
-export const TOOLHIVE_VERSION = process.env.THV_VERSION ?? 'v0.1.1'
+export const TOOLHIVE_VERSION = process.env.THV_VERSION ?? 'v0.1.2'
```
samuv: Let’s do the version bump in a separate PR because it requires regenerating the API

peppescg: I did this, we don't have changes at api layer

peppescg: but anyways let me remove it the same

main/src/main.ts
```diff
@@ -40,8 +41,8 @@ Sentry.init({
 declare const MAIN_WINDOW_VITE_DEV_SERVER_URL: string | undefined
 declare const MAIN_WINDOW_VITE_NAME: string
 
-console.log(`ToolHive binary path: ${binPath}`)
-console.log(`Binary file exists: ${existsSync(binPath)}`)
+log.info(`ToolHive binary path: ${binPath}`)
+log.info(`Binary file exists: ${existsSync(binPath)}`)
 
 // this implements auto-update
 updateElectronApp()
```
kantord: one small thing. this function (auto updater) take logger as an arg. I think we probably have to pass it here? :thinking: 

the opts type for the function for reference

```ts
export interface IUpdateElectronAppOptions<L = ILogger> {
    /**
     * @param {String} repo A GitHub repository in the format `owner/repo`.
     *                      Defaults to your `package.json`'s `"repository"` field
     * @deprecated Use the new `updateSource` option
     */
    readonly repo?: string;
    /**
     * @param {String} host Defaults to `https://update.electronjs.org`
     * @deprecated Use the new `updateSource` option
     */
    readonly host?: string;
    readonly updateSource?: IUpdateSource;
    /**
     * @param {String} updateInterval How frequently to check for updates. Defaults to `10 minutes`.
     *                                Minimum allowed interval is `5 minutes`.
     */
    readonly updateInterval?: string;
    /**
     * @param {Object} logger A custom logger object that defines a `log` function.
     *                        Defaults to `console`. See electron-log, a module
     *                        that aggregates logs from main and renderer processes into a single file.
     */
    readonly logger?: L;
    /**
     * @param {Boolean} notifyUser Defaults to `true`.  When enabled the user will be
     *                             prompted to apply the update immediately after download.
     */
    readonly notifyUser?: boolean;
    /**
     * Optional callback that replaces the default user prompt dialog whenever the 'update-downloaded' event
     * is fired. Only runs if {@link notifyUser} is `true`.
     *
     * @param info - Information pertaining to the available update.
     */
    readonly onNotifyUser?: (info: IUpdateInfo) => void;
}
```

utils/constants.ts
peppescg: done

main/src/main.ts
peppescg: oh cool yes good idea. Let me do it

kantord (dismissed):

main/src/main.ts
peppescg: done https://github.com/StacklokLabs/toolhive-studio/pull/422/commits/64e18806b26a049ed7b3e1a6f340592ecee669cc

samuv (dismissed):
💪🏼 thank you, it should be much easier to debug now

samuv (approved):

Issue #412, opened 2025-07-03
[Task] Add a help menu to ToolHive studio


jtenniswood:
### Description

Add a help icon to the tool bar in the menu
- add link to help and support documentation
- add option to leave feedback (will require some input collection process, maybe GitHub feature request?)

### Priority

Medium

### Task Type

Feature

PR #430, opened 2025-07-04
feat: add help menu

kantord:
Screenshot_select-area_20250704164714

fixes: #412

### DIFF
--- a/renderer/src/common/components/help/help-dropdown.tsx
+++ b/renderer/src/common/components/help/help-dropdown.tsx
@@ -1,0 +1,71 @@
+import { HelpCircle } from 'lucide-react'
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from '@/common/components/ui/dropdown-menu'
+import { Button } from '@/common/components/ui/button'
+import { twMerge } from 'tailwind-merge'
+
+export function HelpDropdown({ className }: { className?: string }) {
+  const handleOpenDocumentation = () => {
+    window.electronAPI.openExternal(
+      'https://github.com/StacklokLabs/toolhive-studio?tab=readme-ov-file#getting-started'
+    )
+  }
+
+  const handleOpenDiscord = () => {
+    window.electronAPI.openExternal('https://discord.gg/stacklok')
+  }
+
+  const handleOpenFeedback = () => {
+    window.electronAPI.openExternal(
+      'https://github.com/StacklokLabs/toolhive-studio/issues'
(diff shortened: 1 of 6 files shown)


renderer/src/common/components/help/help-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+          className={twMerge('cursor-pointer', className)}
+        >
+          <HelpCircle className="text-muted-foreground size-4" />
+          <span className="sr-only">Help</span>
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent align="end" className="w-48">
+        <DropdownMenuItem asChild>
+          <a
+            href="https://github.com/StacklokLabs/toolhive-studio?tab=readme-ov-file#getting-started"
```
samuv, line 27: This is temporary until we have an official page in the Toolhive docs, right?

renderer/src/common/components/help/help-dropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+            target="_blank"
+            rel="noopener noreferrer"
+            className="cursor-pointer"
+          >
+            Discord Community
+          </a>
+        </DropdownMenuItem>
+        <DropdownMenuItem asChild>
+          <a
+            href="https://github.com/StacklokLabs/toolhive-studio/issues"
```
samuv, line 47: Should we have a template for feature requests? 🤔 What do you think?

renderer/src/common/components/help/help-dropdown.tsx
kantord: yeah. we don't have that right now, do we?

renderer/src/common/components/help/help-dropdown.tsx
kantord: I was also considering that, although in that case I think it would need to be 2 links: one for feature requests and one for bug reports.

I found that the issue type selector screen is not linkable. at least it was not trivial to link it. but let me double check that

kantord: although in both cases, leading straight to the form would also encourage the users to file a lot of duplicate issues :thinking:  basically we would be hiding the existing issues form them, encouraging them to file a new one without checking. so on second thought, I think that is not such a good idea

kantord: [so it is possible to link to this page](https://github.com/StacklokLabs/toolhive-studio/issues/new/choose), but it does not solve the problem that this almost forces users to double-report issues

PR #440, opened 2025-07-07
fix: show notification on claude code client after new mcp installed

peppescg:
Claude Code needs to be restarted for connecting to the new MCP servers, show an info icon with a tooltip as a use case description.

Screenshot 2025-07-11 at 20 16 36

### DIFF
--- a/renderer/src/features/mcp-servers/lib/__tests__/orchestrate-run-custom-server.test.tsx
+++ b/renderer/src/features/mcp-servers/lib/__tests__/orchestrate-run-custom-server.test.tsx
@@ -13,6 +13,10 @@ vi.mock('@sentry/electron/renderer', () => ({
   startSpan: vi.fn(),
 }))
 
+vi.mock('../restart-client-notification', () => ({
+  restartClientNotification: vi.fn(),
+}))
+
 const mockStartSpan = vi.mocked(Sentry.startSpan)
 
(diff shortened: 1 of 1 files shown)


jtenniswood:
I think this message should be shown then the user add a new server, and it should be on the installed server page. 
No one goes to the client page (unless somethings broken)

kantord:
> I think this message should be shown then the user add a new server, and it should be on the installed server page. No one goes to the client page (unless somethings broken)

only problem is, what logic will we use to remove the message when it's not needed? I don't think we can keep track of the clients being restarted

peppescg:
what about showing it in the notification?

peppescg:
My ideas was that as a user if there is an issue with an mcp that it is not reachable to the client, I am going to take a look to Studio clients looking if it is still connected

kantord:
> My ideas was that as a user if there is an issue with an mcp that it is not reachable to the client, I am going to take a look to Studio clients looking if it is still connected

this logic makes sense too.

I guess that showing it in the clients section (like it is now) PLUS showing a notification should be a pretty failsafe solution :thinking:  @jtenniswood  what do you think?

peppescg:
@jtenniswood I updated the logic here, showing only a simple info icon with a tooltip for claude code use case

PR #458, opened 2025-07-08
fix: fix restart after update

kantord:


main/src/main.ts
```diff
@@ -394,6 +401,36 @@ ipcMain.handle('restart-toolhive', async () => {
...
+  stopToolhive()
+  
+  // Destroy tray
+  tray?.destroy()
+  
+  // Install update and restart
+  autoUpdater.quitAndInstall()
+  return { success: true }
+})
+
```
peppescg: it kinda makes sense to me, it should work

PR #465, opened 2025-07-09
feat: show current version in tray

peppescg:
Screenshot 2025-07-09 at 10 27 33


peppescg:
@jtenniswood we was thinking to show the current version of the app in the tray system. wdyt?

Issue #466, opened 2025-07-09
[Bug] Gtk-ERROR **: 08:59:46.452: GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported


RoddieKieley:
### Bug Description

Environment: Fedora 42
Version Tag: v0.0.13

Following the included README.md markdown file after successfully executing pnpm install and pnpm thv, pnpm run start was executed. 

This resulted in a successful build, but an unsuccessful execution, with the execution NOT showing the GUI but instead showing this as the last few lines in the terminal:

> ...
> .vite/build/main.js  1,690.04 kB │ gzip: 338.56 kB │ map: 3,738.10 kB
> built in 3253ms.
> 
> 08:59:26.419 > ToolHive binary path: /wip/src/github.com/stacklok/toolhive-studio/bin/linux-x64/thv
> 08:59:26.421 > Binary file exists: true
> 08:59:26.422 > update-electron-app config looks good; aborting updates since app is in development mode
> 
> (process:130782): Gtk-ERROR **: 08:59:46.452: GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported
> 

Asking Cursor backed by o3 the following workaround is presented that allowed a successful launch of the GUI:

> ...
> or launch the dev server with
```
$ bash
   pnpm run start -- --gtk-version=3
```
> (Electron-Forge forwards extra CLI switches).
> Result: the process only ever loads GTK 3, which is compatible with
> libappindicator, so the crash disappears.

As well there are recommendations on how to more permanently resolve the issue, however as I just launched the toolhive-studio for the first time I likely don't have the context to determine which if any are appropriate fixes. Will attach more context subsequent to issue creation.

### Priority

Medium

RoddieKieley:
[toolhive-studio-issue-466-additional-context.txt](https://github.com/user-attachments/files/21142189/toolhive-studio-issue-466-additional-context.txt)

peppescg:
Hey @RoddieKieley thanks for reaching out 🙌 , we will take care of this

kantord:
This appears to be [a bug in Electron itself](https://github.com/electron/electron/issues/46538), however some people report that it is fixed in 37.0.0.

We are still on 36.5.0, let me see if we can use the newer version now!

kantord:
[...]

PR #472, opened 2025-07-09
build: use electron 37.x

kantord:
https://github.com/stacklok/toolhive-studio/issues/466


package.json
```diff
@@ -61,7 +61,7 @@
     "@vitest/coverage-istanbul": "^3.1.4",
     "autoprefixer": "^10.4.21",
     "dotenv": "^16.5.0",
-    "electron": "36.5.0",
+    "electron": "37.0.0",
```
JAORMX: There's 37.2.0

kantord: good point

peppescg (approved):
Tested in mac and I don't see issues 🚢

PR #474, opened 2025-07-09
fix: quit will always trigger graceful-exit

peppescg:
https://github.com/user-attachments/assets/33f2bfdb-a2bf-4da5-81d7-b752956e0791

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -32,7 +32,7 @@ import {
   binPath,
 } from './toolhive-manager'
 import log from './logger'
-import { getAppVersion, pollWindowReady } from './util'
+import { delay, getAppVersion, pollWindowReady } from './util'
 
 import Store from 'electron-store'
 
@@ -123,7 +123,7 @@ export async function blockQuit(source: string, event?: Electron.Event) {
       mainWindow.webContents.send('graceful-exit')
 
       // Give renderer time to navigate to shutdown page
-      await new Promise((resolve) => setTimeout(resolve, 500))
+      await delay(500)
     }
   } catch (err) {
     log.error('Failed to send graceful-exit message: ', err)
@@ -132,7 +132,7 @@ export async function blockQuit(source: string, event?: Electron.Event) {
   try {
     const port = getToolhivePort()
     if (port) {
-      await stopAllServers(binPath, port) // Questo può richiedere molto tempo
+      await stopAllServers(binPath, port)
     }
   } catch (err) {
     log.error('Teardown failed: ', err)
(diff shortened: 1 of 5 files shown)


main/src/main.ts
```diff
@@ -103,8 +103,27 @@ export async function blockQuit(source: string, event?: Electron.Event) {
 
   // Only send graceful-exit message if mainWindow is still valid
```
kantord: nitpick: I guess this comment is a bit outdated here now

main/src/main.ts
```diff
@@ -103,8 +103,27 @@ export async function blockQuit(source: string, event?: Electron.Event) {
...
+
+      mainWindow.show()
+      mainWindow.focus()
+
+      await pollWindowReady(mainWindow)
+
       mainWindow.webContents.send('graceful-exit')
+
+      // Give renderer time to navigate to shutdown page
+      await new Promise((resolve) => setTimeout(resolve, 500))
```
kantord: I think for this type of thing we should introduce an "async sleep function" that can be called like this 


```ts
await sleep(500)
```


basically the code is the same, I would just it in some utils file and make it generic


```ts
export async function sleep(delayMiliseconds: number) {
  return new Promise(resolve => setTimeout(resolve, delayMiliseconds)
}
```

main/src/main.ts
```diff
@@ -113,7 +132,7 @@ export async function blockQuit(source: string, event?: Electron.Event) {
   try {
     const port = getToolhivePort()
     if (port) {
-      await stopAllServers(binPath, port)
+      await stopAllServers(binPath, port) // Questo può richiedere molto tempo
```
kantord:
```suggestion
      await stopAllServers(binPath, port) // can take a long time
```

main/src/util.ts
```diff
@@ -31,3 +32,18 @@ export function getAppVersion(): string {
...
+
+export async function pollWindowReady(window: BrowserWindow): Promise<void> {
+  return new Promise((resolve) => {
+    const checkReady = () => {
+      if (window?.isVisible() && window.webContents.isLoading() === false) {
+        log.info('Window is ready and visible')
+        resolve(void 0)
+      } else {
+        log.info('Window not ready yet, waiting...')
+        setTimeout(checkReady, 100)
```
kantord: I guess the async sleep function I suggested could already be reused in this function :thinking:  probably we could entirely remove the `new Promise`thing from here

main/src/main.ts
peppescg: makes sense

peppescg: I guess we remove also this from graceful-exit and reuse the above everywhere
`main/src/graceful-exit.ts:10`

kantord (approved):

PR #476, opened 2025-07-09
fix: correct VS Code Insiders label

danbarr:
Minor nitpick, the proper name is "VS Code Insiders" (internally thv uses just "insider" but MS pluralizes it)


JAORMX (approved):

peppescg:
Thanks mate

PR #484, opened 2025-07-10
fix: app forgets light/dark mode selection

kantord:


renderer/src/common/components/theme/theme-provider.tsx
```diff
@@ -24,19 +24,16 @@ export function ThemeProvider({
...
-          // Only sync if the stored theme doesn't match the native theme
           const storedTheme = localStorage.getItem(storageKey)
-          if (!isValidTheme(storedTheme) || storedTheme !== nativeThemeSource) {
-            setTheme(nativeThemeSource)
-            localStorage.setItem(storageKey, nativeThemeSource)
+
+          if (!isValidTheme(storedTheme)) {
+            const nativeThemeState = await window.electronAPI.darkMode.get()
+            setTheme(nativeThemeState.themeSource)
+            localStorage.setItem(storageKey, nativeThemeState.themeSource)
```
peppescg, line 36: Shouldn't be saved into file system directly? 🤔

peppescg (approved):

renderer/src/common/components/theme/theme-provider.tsx
kantord: I think that is a great idea, we should have some local YAML config file system or sth like for this, and also for other features.

peppescg: you can use electron-store, that we are currently using for telemetry and restart containers after restart pc

kantord: created an issue for that here https://github.com/stacklok/toolhive-studio/issues/490

Issue #470, opened 2025-07-09
[Bug] Hide table sorting for secrets


jtenniswood:
### Bug Description

<img width="187" height="94" alt="Image" src="https://github.com/user-attachments/assets/ca25e096-08d1-428c-b553-848bedd49276" />

As there's nothing else to sort by

### Priority

Medium

PR #497, opened 2025-07-11
fix: replace title page font, remove sorting on secrets table

peppescg:
Remove sorting on secrets table. Replace all title pages font, created a shared title page component


peppescg:
this https://github.com/stacklok/toolhive-studio/issues/470 is fixed in this pr

PR #500, opened 2025-07-11
feat: implement network isolation for registry MCPs

kantord:
For now, this is behind a feature flag. The feature flag can be enabled like so:

Screenshot_select-area_20250714132322

### DIFF
--- a/renderer/src/features/registry-servers/lib/get-form-schema-run-from-registry.ts
+++ b/renderer/src/features/registry-servers/lib/get-form-schema-run-from-registry.ts
@@ -38,7 +38,7 @@ function refineEnvVar(
   return true
 }
 
-export const validatePort = (val: string) => {
+const validatePort = (val: string) => {
   if (!val.trim()) return 'Port is required'
   if (!/^[0-9]+$/.test(val)) return 'Port must be a number'
   const num = Number(val)
@@ -46,7 +46,7 @@ export const validatePort = (val: string) => {
   return null
 }
 
-export const validateHost = (val: string) => {
+const validateHost = (val: string) => {
   if (!val.trim()) return 'Host is required'
   if (!/^\.?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/.test(val)) return 'Invalid host'
   return null


package.json
```diff
@@ -21,6 +21,7 @@
     "prepare": "husky install",
     "thv": "ts-node ./utils/fetch-thv.ts --run",
     "test": "vitest",
+    "test:nonInteractive": "vitest run",
```
kantord, line 24: I just added this so that a cursorrule/similar can easily be created to always run tests before and after making any changes.

peppescg:
localStorage is either good for enable/disable features 👍

peppescg:
Screenshot 2025-07-14 at 14 03 53
the checkbox seem to have a weird bg color 🤔

peppescg:
Screenshot 2025-07-14 at 14 04 19

I would add some placeholder here

peppescg:
Screenshot 2025-07-14 at 14 05 12

nit: I would trigger validation onLeave for the mcp name, in order to avoid filling the other fields and come back renaming it

peppescg:
After enable the switch I would expact to already have one input field, without clicking on add button, the same for post.

Question, why we don't pre-filling with default values, after enable network isolation?
Screenshot 2025-07-14 at 14 10 07

renderer/src/features/registry-servers/components/form-run-from-registry/allowed-protocols-field.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Label htmlFor="allowed-protocols-group">Allowed Protocols</Label>
+      <div
+        id="allowed-protocols-group"
+        role="group"
+        aria-label="Allowed Protocols"
+      >
+        <div className="mt-2 flex items-center gap-4">
+          <label>
+            <input
+              type="checkbox"
```
peppescg: you can import and use shadcn checkbox https://ui.shadcn.com/docs/components/checkbox

renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
```diff
@@ -0,0 +1,283 @@
...
+function TooltipValueRequired() {
+  return (
+    <Tooltip>
+      <TooltipTrigger asChild autoFocus={false}>
+        <AsteriskIcon className="text-muted-foreground size-4" />
+      </TooltipTrigger>
+      <TooltipContent>Required</TooltipContent>
+    </Tooltip>
+  )
+}
```
peppescg, line 46: I would handle this in a separate PR

renderer/src/features/registry-servers/components/dynamic-array-field.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const DynamicArrayField = ({
+  label,
+  value,
+  onChange,
+  inputLabelPrefix = 'Item',
+  inputProps = {},
+  addButtonText = 'Add',
+  validate,
+}: DynamicArrayFieldProps) => {
+  const [touched, setTouched] = useState<boolean[]>(value.map(() => false))
```
peppescg: I guess you can simplify a lot using `useFieldArray` take a look `renderer/src/features/mcp-servers/components/form-fields-array-custom-secrets.tsx:24`

kantord:
> Screenshot 2025-07-14 at 14 05 12
> nit: I would trigger validation onLeave for the mcp name, in order to avoid filling the other fields and come back renaming it

hmmm, so basically to get rid of the logic for finding the correct tab to show?

renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
kantord: :thinking: not sure what you mean. I didn't write this code, I just moved it to this file as part of refactoring

peppescg:
> > Screenshot 2025-07-14 at 14 05 12
> > nit: I would trigger validation onLeave for the mcp name, in order to avoid filling the other fields and come back renaming it
> 
> hmmm, so basically to get rid of the logic for finding the correct tab to show?

I was thinking to only start the validation on leave, cause basically the name is the first input filled by the user, and we are already calling the list of workloads

renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
peppescg: ah ok I was thinking that there was a refactor on required secrets in the configuration form

renderer/src/features/registry-servers/components/form-run-from-registry/allowed-protocols-field.tsx
kantord: fixed

renderer/src/features/registry-servers/components/dynamic-array-field.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <Label>{label}</Label>
+      <div role="group" aria-label={label} className="w-full">
+        <div className="mt-2 flex w-full flex-col gap-2">
+          {value.map((item, idx) => {
+            const error = validate ? validate(item) : null
+            return (
+              <div key={idx} className="flex w-full flex-col gap-1">
+                <div className="flex w-full items-center gap-2">
+                  <Input
+                    type="text"
```
peppescg: I guess it is better duplicating the fields rather  than use `DynamicArrayField` with config, cause using react-hook-form is really easy to do it, and you can use soz for validation.
Anyways in case you don't want to use react-hook-form I would dynamic set the type 
`type="number"` based on fields, in this way you don't have
Screenshot 2025-07-14 at 15 55 22

PR #502, opened 2025-07-11
feat: replace installed menu item

peppescg:
Screenshot 2025-07-11 at 18 28 41


kantord:
makes a lot of sense. now we just have to replace "Registry" with "Store" :speak_no_evil:

Issue #513, opened 2025-07-14
First MCP server card is too close to left border


blkt:
Clicking on the drop down menu opens a panel that overlaps with the left border.

<img width="334" height="356" alt="Image" src="https://github.com/user-attachments/assets/e018db49-21a3-40a6-a019-15855ffa12a4" />

peppescg:
fixed https://github.com/stacklok/toolhive-studio/pull/517

Issue #511, opened 2025-07-14
[Bug] Wrap long tooltip text


jtenniswood:
### Bug Description

<img width="673" height="70" alt="Image" src="https://github.com/user-attachments/assets/6f198861-d339-43a5-93a7-0131031643ce" />

Set a max width of something reasonable, like 300px. 
Pick something the makes sense

### Priority

Medium

peppescg:
fixed https://github.com/stacklok/toolhive-studio/pull/517

PR #517, opened 2025-07-14
fix: padding registry detail actions section

peppescg:
Screenshot 2025-07-14 at 16 09 56


peppescg:
fixed in this PR:
-  https://github.com/stacklok/toolhive-studio/issues/511
- https://github.com/stacklok/toolhive-studio/issues/513

PR #533, opened 2025-07-14
fix: show notification only if code is registered and installed

peppescg:
https://github.com/user-attachments/assets/bd0d56d4-0a49-4f34-bf7f-11a7d1aa85e3


kantord (approved):

kantord:
LGTM but I would change the commit name to make clear that this is the "client needs to be restarted" notification

peppescg:
> LGTM but I would change the commit name to make clear that this is the "client needs to be restarted" notification

yep it was the old one

PR #551, opened 2025-07-15
feat: add header for release build

peppescg:
Add header in order to track release build


forge.config.ts
```diff
@@ -79,8 +79,8 @@ const config: ForgeConfig = {
           owner: 'stacklok',
           name: 'toolhive-studio',
         },
-        draft: false,
-        prerelease: false,
+        draft: true,
```
kantord: :thinking:

peppescg: testing the release

peppescg: I am going to commit after the test and remove it

kantord (dismissed):

kantord (approved):

PR #554, opened 2025-07-15
fix: fix UI validation for network isolation

kantord:
now validation works as expected, and not tests are marked skipped

https://github.com/user-attachments/assets/82c8539e-71b7-4333-861a-58723738066d

### DIFF
--- a/renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx
+++ b/renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx
@@ -1,7 +1,6 @@
 import type {
   RegistryEnvVar,
   RegistryImageMetadata,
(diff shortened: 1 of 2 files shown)


renderer/src/feature-flags/toggles.json
```diff
@@ -1,3 +1,3 @@
 {
-  "network-isolation": false
+  "network-isolation": true
```
peppescg, line 2: can we remove directly it, cause it is basically GA?
I mean I would remove  `isFeatureEnabled('network-isolation') ` directly

kantord: yes, but I would do it in a different pr

just in case we want to remove it quickly

peppescg (approved):
🚢

PR #580, opened 2025-07-16
fix: update network isolation labels

danbarr:
Updating the label on the network isolation toggle switch to "Enable outbound network filtering" for clarity.

Aligns with the [design](https://www.figma.com/design/KuQGsgrNzRCGmSTl520Hwh/Toolhive-Desktop?node-id=2379-14621&t=3JO9gnfu9XQyDgHW-4).

Also updated the "Allowed hosts"/"Allowed ports" labels to sentence case to match other input box labels around the app.


peppescg (approved):
🚢

PR #617, opened 2025-07-18
fix(registry): fill default args into form

peppescg:
Handles args as tags, the default args are not deletable, the user can add arg pressing enter or space or on input blur 

https://github.com/user-attachments/assets/18906a61-90b8-4f2a-89e3-430a21da72e4

### DIFF
--- a/renderer/src/features/registry-servers/components/form-run-from-registry/index.tsx
+++ b/renderer/src/features/registry-servers/components/form-run-from-registry/index.tsx
@@ -168,6 +168,7 @@ export function FormRunFromRegistry({
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
       <DialogContent
         className="p-0 sm:max-w-2xl"
+        onCloseAutoFocus={() => form.reset()}
         onInteractOutside={(e) => {
           // Prevent closing the dialog when clicking outside
           e.preventDefault()


danbarr:
Quick note, when you run out of the CLI you don't have the opportunity to change these. In the current cases, they're required, to match the transport_type set in the registry. If users change it, it'll break the MCP.

So maybe the UI should do what the CLI does, and just pass these args through and add any other args the user inputs? Vs. populating them into the form?

peppescg:
> Quick note, when you run out of the CLI you don't have the opportunity to change these. In the current cases, they're required, to match the transport_type set in the registry. If users change it, it'll break the MCP.
> 
> So maybe the UI should do what the CLI does, and just pass these args through and add any other args the user inputs? Vs. populating them into the form?

good point mate. So correct me if I am wrong, is it possible to add other args without removing the default one, avoiding breaks it, right?

danbarr:
> good point mate. So correct me if I am wrong, is it possible to add other args without removing the default one, avoiding breaks it, right?

Yeah that's right. So in the buildkite example, on the CLI if you ran like this:
```
thv run buildkite -- --some-other-arg
```

ToolHive will run it with the default from the registry _plus_ your extra args, like `buildkite stdio --some-other-arg`

danbarr:
I found an issue with treating the input like tags...it works fine if you're typing and hit space or mouse out of the control, but if you paste in a set of args (like from the MCP's documentation), it gets added as a single "tag", and it's being added to the container's args as a single string.

Example:
For the GitHub MCP, I pasted in `--toolsets repos,issues,pull_requests --read-only`

When I clicked out of the input, it appeared like this, as a single item instead of 3:
image

Inspecting the container:
```json
        "Args": [
            "--toolsets repos,issues,pull_requests --read-only"
        ],
```

It should look like:

```json
        "Args": [
            "--toolsets",
            "repos,issues,pull_requests",
            "--read-only"
        ],
```

peppescg:
good call @danbarr this is something that I can try to improve

peppescg:
https://github.com/user-attachments/assets/a77f7bef-d70a-40e2-baed-52f27c25b232

@danbarr wdyt?

danbarr:
LGTM!

samuv:
Nice work! I tested it a bit, and it works well.
@peppescg  does it make sense to rebase and apply this to the custom server too?

peppescg:
> Nice work! I tested it a bit, and it works well. @peppescg does it make sense to rebase and apply this to the custom server too?

yep, make sense let me rebase

peppescg:
**Fix https://github.com/stacklok/toolhive-studio/pull/617/commits/80911079f1bcfe6d428037d5491ac616f3d36561 reset form on dialog close**

https://github.com/user-attachments/assets/9f29a67a-0c4a-4e4e-9fbf-241d1d3316cf

peppescg:
**Add https://github.com/stacklok/toolhive-studio/pull/617/commits/6cf14af2bce8cc0432984c5c200bd43f06e16759 cmd_arg badge on custom run form**

https://github.com/user-attachments/assets/af57dca2-2253-4620-8e11-e0772aaee6f6

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
```diff
@@ -102,6 +102,7 @@ export function DialogFormRunMcpServerWithCommand({
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
       <DialogContent
         className="p-0 sm:max-w-2xl"
+        onCloseAutoFocus={() => form.reset()}
```
samuv, line 105: I thought that not resetting the form on closing when no submission was made was something we wanted. We have the same behavior on the registry too

peppescg: 🤔 why? I was wondering that if I cancel is because I don't need that mcp...for starting from scratch the user needs to reload or delete manually the fields

samuv: yeah I don't know the reason, but in any case let's be consistent if we want to change it

samuv (approved):
🚀

PR #620, opened 2025-07-18
feat(registry): show tooltip for secrets and env var

peppescg:
Screenshot 2025-07-18 at 11 30 04
Screenshot 2025-07-18 at 11 28 56

### DIFF
--- a/renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
+++ b/renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
@@ -15,6 +15,7 @@ import type { GroupedEnvVars } from '../../lib/group-env-vars'
 import type { RegistryEnvVar } from '@/common/api/generated/types.gen'
 import { cn } from '@/common/lib/utils'
 import { FormComboboxSecretStore } from '@/common/components/secrets/form-combobox-secrets-store'
+import { TooltipInfoIcon } from '@/common/components/ui/tooltip-info-icon'
 
 interface ConfigurationTabContentProps {
   error: string | null
@@ -51,6 +52,9 @@ function SecretRow({
                   )}
                 >
                   {secret.name}
+                  <TooltipInfoIcon className="m-w-90">
+                    {secret.description}
+                  </TooltipInfoIcon>
                 </FormLabel>
               </FormControl>
             </div>
@@ -127,10 +131,13 @@ function EnvVarRow({
                   htmlFor={`envVar.${index}.value`}
                   className={cn(
                     `text-muted-foreground !border-input flex h-full
-                    items-center gap-1 font-mono !ring-0`
+                    items-center gap-2 font-mono !ring-0`
                   )}
                 >
                   {envVar.name}
+                  <TooltipInfoIcon className="w-90">
+                    {envVar.description}
+                  </TooltipInfoIcon>
                 </FormLabel>
               </FormControl>
             </div>
--- a/renderer/src/common/components/ui/tooltip-info-icon.tsx
+++ b/renderer/src/common/components/ui/tooltip-info-icon.tsx
@@ -13,7 +13,10 @@ export function TooltipInfoIcon({
   return (
     <Tooltip>
       <TooltipTrigger asChild autoFocus={false}>
-        <InfoIcon className="text-muted-foreground size-4 rounded-full" />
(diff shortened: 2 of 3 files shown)


kantord (changes requested):
looks great to me, the only thing I am missing is a few test cases

renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
```diff
@@ -127,10 +131,13 @@ function EnvVarRow({
                   htmlFor={`envVar.${index}.value`}
                   className={cn(
                     `text-muted-foreground !border-input flex h-full
-                    items-center gap-1 font-mono !ring-0`
+                    items-center gap-2 font-mono !ring-0`
```
kantord, line 134: why `!ring-0`? :thinking:

peppescg: it was there before my change I don't know

peppescg: I guess the requirement was to remove the shadow completely from the the input

peppescg:
> looks great to me, the only thing I am missing is a few test cases

the tests case for the form are already handled here `renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx`

kantord:
> > looks great to me, the only thing I am missing is a few test cases
> 
> the tests case for the form are already handled here `renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx`

yeah, but the tests should cover the tooltip

peppescg:
> > > looks great to me, the only thing I am missing is a few test cases
> > 
> > 
> > the tests case for the form are already handled here `renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx`
> 
> yeah, but the tests should cover the tooltip

🤔  i am not totally agree, cause the tooltip is a 3d party library, what is the value to test if the 3d party library works?
I am I don't think we are testing the 3d party library in our app, and it makes sense to me, we should test only our stuff

kantord:
> > > > looks great to me, the only thing I am missing is a few test cases
> > > 
> > > 
> > > the tests case for the form are already handled here `renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx`
> > 
> > 
> > yeah, but the tests should cover the tooltip
> 
> 🤔 i am not totally agree, cause the tooltip is a 3d party library, what is the value to test if the 3d party library works? I am I don't think we are testing the 3d party library in our app, and it makes sense to me, we should test only our stuff

the point of the test is not to test that the 3rd party library works. The point is to test that the user facing behavior is as expected (interaction is wired to hover event, for instance) and that the data is correctly wired (the actual data coming from the API is displayed, and the data displayed corresponds the actual item (for instance, there is no index-mismatch)

So yeah, we definitely *should* avoid testing the 3rd party library's functionality, but the test should still described the expected behavior (`As a user, if I am on the registry form and hover over the icon next to X environment variable, I can read the corresponding documentation`)

peppescg:
> > > > > looks great to me, the only thing I am missing is a few test cases
> > > > 
> > > > 
> > > > the tests case for the form are already handled here `renderer/src/features/registry-servers/components/__tests__/form-run-from-registry.test.tsx`
> > > 
> > > 
> > > yeah, but the tests should cover the tooltip
> > 
> > 
> > 🤔 i am not totally agree, cause the tooltip is a 3d party library, what is the value to test if the 3d party library works? I am I don't think we are testing the 3d party library in our app, and it makes sense to me, we should test only our stuff
> 
> the point of the test is not to test that the 3rd party library works. The point is to test that the user facing behavior is as expected (interaction is wired to hover event, for instance) and that the data is correctly wired (the actual data coming from the API is displayed, and the data displayed corresponds the actual item (for instance, there is no index-mismatch)
> 
> So yeah, we definitely _should_ avoid testing the 3rd party library's functionality, but the test should still described the expected behavior (`As a user, if I am on the registry form and hover over the icon next to X environment variable, I can read the corresponding documentation`)

mm I am not totally agree for this specific use case, cause it is a raw description returned by BE without any UI wrapper. But anyway let me add it 👍

kantord (approved):

kantord (approved):

PR #621, opened 2025-07-18
test: add end-to-end testing

kantord:
This PR:

- Sets up playwright for e2e tests
- Adds video recording capability
- Makes all fixtures downloadable after GHA run
- Adds a GHA workflow that sets up a test environment with a virtual screen (valid for screen recording) and a secrets provider (required for tests to pass)
- Contains a relatively minimal test case that actually installs and removes an MCP
- A tiny second test case is included - even though this test case is not very useful, I want to keep it until we have a more legit second test case. reason: I want to know if a configuration/fixture change breaks test the system's ability to run multiple test cases

What's not included:
- some features cannot be tested even with playwright, because playwright is not able to interact with GUI features outside of the browser context. We could still test these with linux specific testing tools that we merely coordinate through playwright, but it would be a lot of extra logic and the tests would be more difficult to run on Mac OS
- You cannot run the e2e test and local dev server locally _at the same time_: this is actually because the app only supports one instance. we could solve this by giving an "alter ego" to toolhive when running the dev server - out of scope for this PR
- Test isolation: the current solution runs an OS-level `thv` instead of running it on test-level, so test isolation is not properly implemented. I have a solution for this, but I reverted the code regarding this in this PR in the interest of keeping the PR small
- I didn't look into collecting coverage from playwright yet

Gotchas:
- running `npx playwright install ffmpeg` _might_ be required on some systems for the e2e tests to run locally. Most systems might have this even if you don't install it directly, because a lot of video related software such as video players, screen recorders, video editors etc. depend on it

### Reports

The debugging experience is surprisingly good. Playwright seems to work more or less as well as it works for web apps, and **a proper report is generated with screenshots included**. This report is even included as an artifact for GHA test runs:

Screenshot_select-area_20250721123201

### Video recordings

Video recordings are included for debugging, and this works on GHA too - recordings are retained for 30 days as GHA artifacts.

Here's an example of a video recording I downloaded from GitHub actions. **It also shows that the test that we are running actually is already a minimally useful end-to-end test**:

[0507350c8a254d17d2349d74fc1f2863.webm](https://github.com/user-attachments/assets/763be8c6-b0ae-446b-b0c2-47507db60869)

fixes [#627](https://github.com/stacklok/toolhive-studio/issues/627)


.github/workflows/_e2e.yml
```diff
@@ -0,0 +1,83 @@
...
+            echo "GNOME_KEYRING_CONTROL=$GNOME_KEYRING_CONTROL"
+            echo "GNOME_KEYRING_PID=$GNOME_KEYRING_PID"
+          } >> "$GITHUB_ENV"
+          sleep 2
+
+      - name: Run tests
+        env:
+          ELECTRON_DISABLE_SANDBOX: 'true'
+        run: |
+          xvfb-run --auto-servernum --server-args="-screen 0 1920x1080x24" pnpm run playwright test
```
kantord, line 62: not sure if this needs any extra commenting, but `xvfb-run` lends a virtual monitor to playwright, so all GUI features work properly, and even video recording is possible

.github/workflows/_e2e.yml
```diff
@@ -0,0 +1,83 @@
+name: End-to-end tests
+on:
+  push:
+    branches: [ main, master ]
```
peppescg: we don't have master we can remove it, the same below for pr

peppescg:
wow it seems extremely easy and smooth the configuration. So we are using `sudo systemctl start docker` for docker. It seems it is using the build from forge right? basically pnpm make

kantord:
> wow it seems extremely easy and smooth the configuration. So we are using `sudo systemctl start docker` for docker. It seems it is using the build from forge right? basically pnpm make

yup exactly, it's using the electron-forge build system, surprisingly no hacks there at all!

.github/workflows/_e2e.yml
kantord: true

kantord: thanks for pointing it out

kantord:
> wow it seems extremely easy and smooth the configuration. So we are using `sudo systemctl start docker` for docker. It seems it is using the build from forge right? basically pnpm make

and yup exactly, that is a totally standard way of running docker in ubuntu, also no hacks.

There were a few things that took a lot of time to properly configure:
- Artifact uploading, especially video recording
- Making electron not crash in the GHA environment (sandboxing, lack of screen)
- Proper configuration of system keyring & unlocking it - as this is not configured in the ubuntu image by default and there seems to be no widely used Action for it in the marketplace
- Test isolation (but in the end I didn't end up including it)

But the great thing is that all the solutions just needed a lot of trial-and-error, but no hacks at all

PR #622, opened 2025-07-18
feat: add settings page

peppescg:
# Settings Tabs Components

## Overview
Four React components that provide comprehensive settings management for the ToolHive application.

### GeneralTab
Core application settings including auto-launch, error reporting, quit confirmation, and application exit functionality.

### VersionTab
Displays version information for the desktop UI, ToolHive binary, and build type (Release/Development).

### LogsTab
Provides access to application logs with platform-specific paths, clipboard copy functionality, and log file download.

### SettingsTabs
Main container with vertical tab navigation, keyboard accessibility, and responsive layout managing all settings sections.

## Key Features
- TypeScript support with React Query integration
- Electron API integration for platform-specific functionality
- Comprehensive error handling and accessibility compliance
- Modern UI with consistent styling patterns
- moved theme to general settings
- Removed settings dropdown ( redirect on click to settings page directly)
- Removed quit button from settings

https://github.com/user-attachments/assets/4c851c6d-f643-4fab-b291-8c65fcaa0f0c

___
Moved Theme to general settings
Screenshot 2025-07-22 at 10 09 27

### DIFF
--- a/renderer/src/common/components/settings/tabs/general-tab.tsx
+++ b/renderer/src/common/components/settings/tabs/general-tab.tsx
@@ -19,12 +19,75 @@ import log from 'electron-log/renderer'
 
 const CONFIRM_QUIT_STORAGE_KEY = 'doNotShowAgain_confirm_quit'
 
-export function GeneralTab() {
-  const { data: autoLaunchStatus, isLoading: isAutoLaunchLoading } =
-    useAutoLaunchStatus()
-  const { mutateAsync: setAutoLaunch, isPending: isSetPending } =
-    useSetAutoLaunch()
+function WrapperField({
+  children,
+  label,
+  description,
+  htmlFor,
+}: {
+  children: React.ReactNode
+  label: string
+  description: string
+  htmlFor: string
+}) {
+  return (
+    <div className="flex items-center justify-between">
+      <div className="space-y-0.5">
+        <Label htmlFor={htmlFor}>{label}</Label>
+        <p className="text-muted-foreground text-sm">{description}</p>
+      </div>
+      {children}
+    </div>
+  )
+}
+
+function ThemeField() {
   const { theme, setTheme } = useTheme()
+
+  const handleThemeChange = async (newTheme: 'light' | 'dark' | 'system') => {
+    try {
+      await setTheme(newTheme)
+    } catch (error) {
+      log.error('Failed to change theme:', error)
+    }
+  }
+
+  return (
+    <WrapperField
+      label="Theme"
+      description="Choose how ToolHive looks to you"
+      htmlFor="theme-select"
+    >
+      <Select value={theme} onValueChange={handleThemeChange}>
+        <SelectTrigger id="theme-select" className="w-32">
+          <SelectValue placeholder="Select theme" />
+        </SelectTrigger>
+        <SelectContent>
+          <SelectItem value="light">
+            <div className="flex items-center gap-2">
+              <Sun className="size-4" />
+              Light
+            </div>
+          </SelectItem>
+          <SelectItem value="dark">
+            <div className="flex items-center gap-2">
+              <Moon className="size-4" />
+              Dark
+            </div>
+          </SelectItem>
+          <SelectItem value="system">
+            <div className="flex items-center gap-2">
+              <Monitor className="size-4" />
+              System
+            </div>
+          </SelectItem>
+        </SelectContent>
+      </Select>
+    </WrapperField>
+  )
+}
+
+function TelemetryField() {
   const { mutateAsync: sentryOptIn, isPending: isOptInPending } = useMutation({
     mutationFn: window.electronAPI.sentry.optIn,
   })
@@ -37,19 +100,6 @@ export function GeneralTab() {
     queryFn: window.electronAPI.sentry.isEnabled,
     queryKey: ['sentry.is-enabled', { isOptInPending, isOptOutPending }],
   })
-  const [skipQuitConfirmation, setSkipQuitConfirmation] =
-    useState<boolean>(false)
-
-  useEffect(() => {
-    const quitConfirmationDisabled =
-      localStorage.getItem(CONFIRM_QUIT_STORAGE_KEY) === 'true'
-    setSkipQuitConfirmation(quitConfirmationDisabled)
-  }, [])
-
-  const handleAutoLaunchToggle = async () => {
-    if (isAutoLaunchLoading || isSetPending) return
-    await setAutoLaunch(!autoLaunchStatus)
-  }
 
   const handleTelemetryToggle = async () => {
     if (isSentryLoading || isOptInPending || isOptOutPending) return
@@ -59,6 +109,60 @@ export function GeneralTab() {
       await sentryOptIn()
     }
   }
+
+  return (
+    <WrapperField
+      label="Error reporting"
+      description="Help improve ToolHive by sending error reports to Sentry"
+      htmlFor="telemetry"
+    >
+      <Switch
+        id="telemetry"
+        checked={isTelemetryEnabled || false}
+        onCheckedChange={handleTelemetryToggle}
+        disabled={isSentryLoading || isOptInPending || isOptOutPending}
+      />
+    </WrapperField>
+  )
+}
+
+function AutoLaunchField() {
+  const { mutateAsync: setAutoLaunch, isPending: isSetPending } =
+    useSetAutoLaunch()
+
+  const { data: autoLaunchStatus, isLoading: isAutoLaunchLoading } =
+    useAutoLaunchStatus()
+
+  const handleAutoLaunchToggle = async () => {
+    if (isAutoLaunchLoading || isSetPending) return
+    await setAutoLaunch(!autoLaunchStatus)
+  }
+
+  return (
+    <WrapperField
+      label="Start on login"
+      description="Automatically start ToolHive when you log in to your computer"
+      htmlFor="auto-launch"
+    >
+      <Switch
+        id="auto-launch"
+        checked={autoLaunchStatus || false}
+        onCheckedChange={handleAutoLaunchToggle}
+        disabled={isAutoLaunchLoading || isSetPending}
+      />
+    </WrapperField>
+  )
+}
+
+function QuitConfirmationField() {
+  const [skipQuitConfirmation, setSkipQuitConfirmation] =
+    useState<boolean>(false)
+
+  useEffect(() => {
+    const quitConfirmationDisabled =
+      localStorage.getItem(CONFIRM_QUIT_STORAGE_KEY) === 'true'
+    setSkipQuitConfirmation(quitConfirmationDisabled)
+  }, [])
 
   const handleQuitConfirmationToggle = () => {
     const newValue = !skipQuitConfirmation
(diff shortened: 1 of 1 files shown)


renderer/src/common/components/settings/tabs/logs-tab.tsx
```diff
@@ -0,0 +1,114 @@
...
+          {logPath && (
+            <Button
+              variant="outline"
+              size="sm"
+              onClick={handleDownloadLog}
+              disabled={isDownloading}
+              className="mt-3 w-fit"
+            >
+              <Download className="mr-2 size-4" />
+              {isDownloading ? 'Downloading...' : 'Download log file'}
```
kantord:
```suggestion
              {isDownloading ? 'Loading...' : 'Save log file'}
```

Nit-pick, but I think this wording is much better for a desktop app

peppescg: makes totally sense

renderer/src/common/components/settings/tabs/logs-tab.tsx
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      await navigator.clipboard.writeText(logPath)
+      setCopied(true)
+      delay(2000).then(() => setCopied(false))
+    } catch (error) {
+      log.error('Failed to copy to clipboard:', error)
+    }
+  }
+
+  const handleDownloadLog = async () => {
```
kantord, line 43: I think this would merit it's own function for readability/separation of concerns

I guess here I would want to see sth like 

```tsx
setIsDownloading(true)
try {
   downloadFile("toolhive-log.txt", getLogContent())
} catch (error) {
   // same logs
} finally {
   setIsDownloading(false)
}
```



but I think even the transition features from react could be leveraged to reduce the manual logic here?


```tsx
  const [isDownloading, startDownload] = useTransition();

// ... etc

   const handleDownloadLog = () => {
      startDownload(() => {
         downloadFile("toolhive-log.txt", getLogContent())
      })
   }
```

peppescg: I guess a hook would makes sense

kantord: yup, also now that I'm thinking about it, I think that the interactive parts of the page could be extracted to a `<LogDownloader />` component as well

peppescg: what do you mean with interactive part 🤔 ?

renderer/src/common/components/settings/tabs/logs-tab.tsx
```diff
@@ -0,0 +1,114 @@
...
+import log from 'electron-log/renderer'
+import { toast } from 'sonner'
+
+const LOG_PATHS = {
+  darwin: '~/Library/Logs/ToolHive/main.log',
+  win32: '%USERPROFILE%\\AppData\\Roaming\\ToolHive\\logs\\main.log',
+  linux: '~/.config/ToolHive/logs/main.log',
+} as const
+
+export function LogsTab() {
```
kantord, line 14: This is not a code issue but a random ux thing I thought of :thinking: 

it is slightly confusing that we have a logs page here, but there is no mention of MCP-related logs. It might create the impression that those logs don't even exist/are not available.

I suppose adding an extra note about this would be enough, or we could just simply list links to all MCP-server log pages here in a simple table. Basically 2 columns: MCP server name, and a logs link/button

kantord: but I am happy to add this in a follow up PR for you after this one is merged, if you prefer

renderer/src/common/components/settings/tabs/logs-tab.tsx
peppescg: it is already inside logs tab this specific page, I would say to having a specific LogDownloader will be not used in other page 🤔 in case we will have other use case like this, I would image to have this kind of component, but right now, it seems only related to settings page

renderer/src/common/components/settings/tabs/settings-tabs.tsx
```diff
@@ -0,0 +1,60 @@
...
+    label: 'Version',
+    value: 'version',
+  },
+  {
+    label: 'Logs',
+    value: 'logs',
+  },
+]
+
+export function SettingsTabs() {
```
kantord, line 21: I love this component, it's very well structured and the code is easy to read! :100:

renderer/src/common/components/settings/tabs/version-tab.tsx
```diff
@@ -0,0 +1,62 @@
...
+    isReleaseBuild: boolean
+    toolhiveVersion: string
+  }>({
+    appVersion: 'N/A',
+    isReleaseBuild: false,
+    toolhiveVersion: 'N/A',
+  })
+
+  useEffect(() => {
+    const fetchVersionInfo = async () => {
```
kantord: I guess my only complaint is that I would move this to a separate function/hook to move it out of the component body (or even to a common file)

renderer/src/common/components/settings/tabs/logs-tab.tsx
kantord: you are right that it is already a specific page. There are no huge red-flags here. 

I guess I am influenced by Clean Code here, but for me, the value in creating function is not just that they can be reused elsewhere, but also that they divide complexity and allow us to use naming to hide irrelevant details from the reader.

When there is some business logic, I like to always separate that into it's own function (or even multiple functions) instead of mixing it in with other stuff

renderer/src/common/components/settings/tabs/__tests__/logs-tab.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+      loading: vi.fn(),
+      success: vi.fn(),
+      warning: vi.fn(),
+      error: vi.fn(),
+      dismiss: vi.fn(),
+    },
+  }
+})
+
+describe('LogsTab', () => {
```
kantord, line 47: love the test code as well :rocket: :100: :heart:

renderer/src/common/components/settings/tabs/__tests__/settings-tabs.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+  it('shows Version tab as default active tab', async () => {
+    renderWithProviders(<SettingsTabs />)
+    await userEvent.click(screen.getByRole('tab', { name: 'Version' }))
+
+    await waitFor(() => {
+      expect(
+        screen.getByRole('heading', { name: 'Version Information' })
+      ).toBeVisible()
+    })
+  })
```
kantord, line 93:
```suggestion
  })
  
```

renderer/src/common/components/settings/tabs/__tests__/settings-tabs.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    mockElectronAPI.getAppVersion.mockResolvedValue('1.0.0')
+    mockElectronAPI.isReleaseBuild.mockResolvedValue(true)
+    mockElectronAPI.getToolhiveVersion.mockResolvedValue('0.9.0')
+    mockElectronAPI.sentry.isEnabled.mockResolvedValue(true)
+  })
+
+  it('renders all tab triggers', async () => {
+    renderWithProviders(<SettingsTabs />)
+
+    expect(screen.getByRole('tab', { name: 'General' })).toBeInTheDocument()
```
kantord: why not?

```suggestion
    expect(screen.getByRole('tab', { name: 'General' })).toBeVisible()
```

renderer/src/common/components/settings/tabs/logs-tab.tsx
peppescg: I created an hook con root folder, in this way let's see if there will be other use case for a specific LogDownloader component, let's start with the hook that is totally makes sense in terms of readability

renderer/src/common/components/settings/tabs/general-tab.tsx
```diff
@@ -0,0 +1,167 @@
...
+          </div>
+          <Switch
+            id="auto-launch"
+            checked={autoLaunchStatus || false}
+            onCheckedChange={handleAutoLaunchToggle}
+            disabled={isAutoLaunchLoading || isSetPending}
+          />
+        </div>
+
+        <div className="flex items-center justify-between">
```
kantord: My nit-pick with these (along the same Clean Code-based logic that I argued in the other thread) is that all the fields could be separate simple functions, and then the component would read sth like this:

```tsx
<PageWrapper>
  <h2 className="text-lg font-semibold">General Settings</h2>

  <ThemeField />
  <StartOnLoginField />
  <TelemetryField />
  <QuitConfirmationField />
</PageWrapper>
```

again, there are no red flags in the current code at all. It's just my code type preference because this way, for example, if someone wants to fix something in the telemetry field, they can just use go-to-definition real quick and only have to look at the relevant code, not a longer file

renderer/src/common/components/settings/tabs/logs-tab.tsx
peppescg: mm I am not agree, this is a settings page, settings page is always related to application level, so the logs are related to app, not mcp. I think adding a table with mcp server logs doesn't math the scope of the page, and it also adds extra logic and noise on this page

renderer/src/common/components/settings/tabs/version-tab.tsx
peppescg: it is really tiny and only fetching from ipc, but yep let me create a custom hook for this

renderer/src/common/components/settings/tabs/__tests__/settings-tabs.test.tsx
peppescg: sorry I am missing

renderer/src/common/components/settings/tabs/logs-tab.tsx
kantord: while this is largely true, as a user I would totally see myself wandering around the app and not analyzing the structure of the app, just seeing the word "Logs" and thinking that this is where I would find MCP logs I have been looking for 

about extra logic and noise I could agree, I guess that we should just point the user to the appropriate place. However, we don't have an overall logs page :thinking: so pointing the user to the appropriate place could be challenging :thinking:

peppescg: I see more an issue not having a detail/edit page per mcp, where we can have a log tab...it seems the proper way in terms of user expection

kantord: that is true as well. But for example, if we have a "network isolation" tab in settings that could also be confusing, because we also have it per-MCP. I see the same problem with the logs

renderer/src/common/components/settings/tabs/__tests__/settings-tabs.test.tsx
kantord: just an empty line between 2 functions

renderer/src/common/components/settings/tabs/general-tab.tsx
peppescg: makes sense 👍

peppescg: done https://github.com/stacklok/toolhive-studio/pull/622/commits/ddf98169279de19f7d0bc4fb0d593555d561d3ae

renderer/src/common/components/settings/tabs/logs-tab.tsx
peppescg: I would ask to James if he see values showing mcp logs in these pages 🤔 cause honestly I am missing the relation, anyway I would consider it in a separate PR

PR #634, opened 2025-07-22
feat(main): fix and extend container engine check with rancher desktop dockerd

samuv:
Enhanced Container Engine Detection with Rancher Desktop Support

We had a problem where we had a packaged app failures: ENOENT errors in built Electron apps due to limited PATH, so I refactored a little bit the container engine sections and extend with Rancher Desktop that we should support now

Screenshot 2025-07-22 at 15 25 06

https://github.com/user-attachments/assets/c049c517-1238-4d1d-a5bd-01ccf3eccdd1

### DIFF
--- a/main/src/container-engine.ts
+++ b/main/src/container-engine.ts
@@ -1,12 +1,47 @@
 import { exec } from 'node:child_process'
 import { promisify } from 'node:util'
+import { platform } from 'node:os'
 import log from './logger'
 
 const execAsync = promisify(exec)
 
+// Container engine installation paths
+const WINDOWS_CONTAINER_PATHS = [
+  // Docker Desktop
+  'C:\\Program Files\\Docker\\Docker\\resources\\bin',
+  // Podman
+  'C:\\Program Files\\RedHat\\Podman',
+  // Rancher Desktop
+  '%APPDATA%\\rancher-desktop\\bin',
+  '%USERPROFILE%\\AppData\\Roaming\\rancher-desktop\\bin',
+] as const
+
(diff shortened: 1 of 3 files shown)


main/src/container-engine.ts
```diff
@@ -23,56 +58,137 @@ const createEngineResult = (
...
+    const successfulResults = getSuccessfulResults(results)
+    const status = combineEngineResults(successfulResults)
+
+    const foundEngines = successfulResults
+      .filter((r) => r.available)
+      .map((r) => r.name)
+    if (foundEngines.length > 0) {
+      log.info(`Container engines detected: ${foundEngines.join(', ')}`)
+    }
+
```
peppescg: maybe it would be useful to add a log in case all the foundEngines.length are not detected?

main/src/container-engine.ts
```diff
@@ -1,12 +1,47 @@
...
+  '/snap/bin',
+  // Flatpak
+  '/var/lib/flatpak/exports/bin',
+  // Rancher Desktop user install
+  '~/.rd/bin',
+  // Rancher Desktop system install (Linux)
+  '/opt/rancher-desktop/bin',
+  // Rancher Desktop (macOS)
+  '/Applications/Rancher Desktop.app/Contents/Resources/resources/darwin/bin',
+] as const
```
peppescg: nice

main/src/container-engine.ts
samuv: I added just one message for when no container engine is available, after all that’s the only feedback we need in the log

peppescg (approved):
🚢

PR #636, opened 2025-07-22
fix: dismiss toast in case of shutting down

peppescg:
Enhanced the useMutationRestartServerAtStartup hook to automatically dismiss the restart servers toast when a server shutdown event occurs.

**Key Improvements**
- Toast ID Management: Added dedicated TOAST_ID constant ('restart-servers-startup') for consistent toast identification
- Automatic Cleanup: Implemented onServerShutdown event listener to dismiss active restart toast when servers are shut down
- Better UX: Prevents lingering "Restarting servers..." notifications when the application or servers are being shut down

https://github.com/user-attachments/assets/f2913763-26d6-4670-ae8d-78e361e3f06f

### DIFF
--- a/renderer/src/common/hooks/use-toast-mutation.ts
+++ b/renderer/src/common/hooks/use-toast-mutation.ts
@@ -17,11 +17,13 @@ export function useToastMutation<
   successMsg,
   errorMsg,
   loadingMsg,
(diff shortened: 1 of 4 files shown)


samuv (approved):
💪🏼

PR #396, opened 2025-07-02
feat(registry): improve registry install form

samuv:
This is moving from toast only messages to inline form, that handle the possible errors directly on the submit of the form

new flow:

https://github.com/user-attachments/assets/172795b1-ba5a-435f-ae2e-996a833ac713

fine tunned error:

Screenshot 2025-07-04 at 16 51 05

PR #651, opened 2025-07-24
feat(mcp): align feedback on custom server installation

samuv:
As per title, in this PR I'm aligning the UX that we have in registry installation with the work done [here](https://github.com/stacklok/toolhive-studio/pull/396)

- moved some component that are reused in the correct folder
- adjusted the validations for some fields
- refreshed test suites to match new component structure
- cove the form with some tests

https://github.com/user-attachments/assets/cee31cec-ea4d-4754-8635-628cd9304970


renderer/src/features/mcp-servers/hooks/use-run-custom-server.tsx
```diff
@@ -0,0 +1,193 @@
...
+        newlyCreatedSecrets,
+        existingSecrets,
+      }
+    },
+  })
+
+  const { mutate: installServerMutation } = useMutation({
+    mutationFn: async ({ data }: { data: FormSchemaRunMcpCommand }) => {
+      const { newlyCreatedSecrets, existingSecrets } = await handleSecrets(data)
+      // Step 4: Create the MCP server workload
```
kantord, line 157: are these comments (step 1, step 2, etc) intentional?

samuv: this is what we had before,I didn’t change it. I think is still valuable to understand the flow in any case, I just moved some logic to be executed in different phases instead of within a single async function

Let me know if you need to adjust anything for network isolation, maybe we can revisit the comments then

kantord (approved):

PR #653, opened 2025-07-24
feat: network filtering for custom servers

kantord:
Notes:

- removed feature flagging for this feature altogether
- did some minimal refactoring to reduce code duplication

https://github.com/user-attachments/assets/0de8e9a4-b564-4335-a569-fc2b47796072


renderer/src/common/hooks/use-form-tab-state.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+interface UseFormTabStateReturn {
+  activeTab: string
+  setActiveTab: (tab: string) => void
+  showFieldError: (fieldName: string) => void
+  activateTabWithError: (errors: Record<string, unknown>) => void
+}
+
+export function useFormTabState({
```
kantord: note, this just extracts existing common functionality, but it's already existing code. 

in a later design the tabs will likely be removed and replaced with a multi-step form where likely there will be a validation at each step.

however, this does **not** mean that this hook will no longer be needed. this is because some errors, such as duplicated MCP name will only be discovered at the final submission. So this same logic will likely be kept long term, perhaps with small adjustments

kantord: also perhaps there is a more sophisticated/idiomatic way to do it with react-hook-forms :thinking: I didn't want to do a mega refactor here, but I am open to suggestions on that and I can implement it in a follow-up PR :thinking:

renderer/src/common/hooks/use-form-tab-state.ts
```diff
@@ -0,0 +1,54 @@
...
+  fieldTabMap,
+  defaultTab = 'configuration',
```
peppescg: I cause it is in common, I wouldn't expect to have a default tab that is related to specific feature logic.
I would probably set it as mandatory prop at this point

renderer/src/common/hooks/use-form-tab-state.ts
```diff
@@ -0,0 +1,54 @@
...
+  const [activeTab, setActiveTab] = useState(defaultTab)
+
+  const showFieldError = (fieldName: string) => {
+    const tabForField = fieldTabMap.find(
+      ({ field }) => field === fieldName
+    )?.tab
+    if (tabForField) {
+      setActiveTab(tabForField)
+    }
+  }
```
peppescg: is this needed 🤔 ?
I guess that the `activateTabWithError` is enough maybe? and it is also not used in the code

renderer/src/common/hooks/use-form-tab-state.ts
peppescg: I think the logic makes sense, I would not expect that react-hook-forms can be aware that there are form in different tabs, it is up to us to handle it, and this hook makes sense. 
Maybe, I would add a simple unit test

renderer/src/common/hooks/use-form-tab-state.ts
kantord: that is a good point, actually, this hook should not even know any "data"

renderer/src/common/hooks/use-form-tab-state.ts
kantord: what I noticed is that due to how the dialog system works, after closing the modal it does not jump to the first tab, but to the last tab that was active. this was a simplest way I could fix that, but I also don't like this solution a lot, so I'm curious if you have a better idea.

One option I was thinking is moving this into the tab hook, but it would still need some argument that changes between re-render I guess

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
```diff
@@ -24,6 +24,27 @@ import { FormFieldsArrayCustomSecrets } from './form-fields-array-custom-secrets
...
+  { field: 'protocol', tab: 'configuration' },
+  { field: 'package_name', tab: 'configuration' },
+  { field: 'target_port', tab: 'configuration' },
+  { field: 'cmd_arguments', tab: 'configuration' },
+  { field: 'envVars', tab: 'configuration' },
+  { field: 'secrets', tab: 'configuration' },
+  { field: 'allowedHosts', tab: 'network-isolation' },
+  { field: 'allowedPorts', tab: 'network-isolation' },
+  { field: 'networkIsolation', tab: 'network-isolation' },
+]
```
peppescg:
```suggestion
const FIELD_TAB_MAP = [
  { field: 'name', tab: 'configuration' },
  { field: 'transport', tab: 'configuration' },
  { field: 'type', tab: 'configuration' },
  { field: 'image', tab: 'configuration' },
  { field: 'protocol', tab: 'configuration' },
  { field: 'package_name', tab: 'configuration' },
  { field: 'target_port', tab: 'configuration' },
  { field: 'cmd_arguments', tab: 'configuration' },
  { field: 'envVars', tab: 'configuration' },
  { field: 'secrets', tab: 'configuration' },
  { field: 'allowedHosts', tab: 'network-isolation' },
  { field: 'allowedPorts', tab: 'network-isolation' },
  { field: 'networkIsolation', tab: 'network-isolation' },
] as const satisfies ReadonlyArray<{
  field: string;
  tab: 'configuration' | 'network-isolation';
}>;
```

adding `as const` and `satisfies` you are enforcing the check value, avoiding possible typo

peppescg: probably I would also create a union type
`type Tab = 'configuration' | 'network-isolation';` using it everywhere

renderer/src/common/hooks/use-form-tab-state.ts
peppescg: Is this fn `showFieldError` used 🤔 ? cause I don't see it used in the code

peppescg: after closing the modal i would expect a reset of the form and tabs, but this should be handled here
` onCloseAutoFocus={() => form.reset()}` yesterday I fixed it for the forms, maybe you need to add the setActiveTab to default one here 🤔 ?

kantord: good idea. although my "dream" solution would be something that is totally implicit/at least declarative. but this suggestion is certainly a lot better than my implementation

kantord: fixed, I even added a reset function to the tabs hook so that the close is clean

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
kantord: cool, I will implement this idea. I also added type generics for the hook, so the hook can internally avoid typos as well.

kantord: might be almost overengineering, but I changed it in such a way that a record is used to model the mapping. The keys can be (at least partially) derived from the zod schema, this way, we get a warning not only for typos, but even when we add a new field and forget to map it. (Only exception is the conditional part of the forms, I didn't find a trivial way to derive that part, so in the interest of time, I did still hardcode that part)

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
```diff
@@ -24,6 +24,34 @@ import { FormFieldsArrayCustomSecrets } from './form-fields-array-custom-secrets
...
+  protocol: 'configuration',
+  package_name: 'configuration',
+  target_port: 'configuration',
+  cmd_arguments: 'configuration',
+  envVars: 'configuration',
+  secrets: 'configuration',
+  allowedHosts: 'network-isolation',
+  allowedPorts: 'network-isolation',
+  networkIsolation: 'network-isolation',
+} satisfies FieldTabMapping<Tab, Field>
```
peppescg, line 54: I would add `as const satisfies FieldTabMapping<Tab, Field>` in this way the object is readonly

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
peppescg: it makes sense to me 🚀 
I would only add as const, for leveraging the readonly capability, but it is a nit

peppescg:
from the video in the description it seems you submitted the form having a validation error for form field 🤔 
how is it possible?

renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx
kantord: you are right, I forgot it in one of the places

kantord:
> from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?

hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out

kantord:
> > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> 
> hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out

yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it

peppescg:
> > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > 
> > 
> > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> 
> yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it

I don't understand I added the validation onChange....it should work 🤔

renderer/src/features/mcp-servers/components/network-isolation-tab-content.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    </AlertDescription>
+                  </Alert>
+                )}
+                <Controller
+                  control={form.control}
+                  name="allowedHosts"
+                  render={() => (
+                    <DynamicArrayField<FormSchemaRunMcpCommand>
+                      /*
+                         // @ts-expect-error no time to fix this */
```
samuv, line 58: ?

renderer/src/features/mcp-servers/components/form-fields-run-mcp-command.tsx
```diff
@@ -34,16 +34,30 @@ export function FormFieldsRunMcpCommand({
...
-            <TabsList id={field.name} className="grid w-full grid-cols-2">
-              <TabsTrigger value="docker_image">Docker image</TabsTrigger>
-              <TabsTrigger value="package_manager">Package manager</TabsTrigger>
-            </TabsList>
-          </Tabs>
+          <FormItem className="mb-8">
+            <FormLabel className="mb-2">Server type</FormLabel>
+            <FormControl>
+              <RadioGroup
+                onValueChange={field.onChange}
```
samuv, line 41: should we reset the validation after we select the server type?

edit: ops I just realize that you were talking about this in other comments

renderer/src/features/mcp-servers/components/network-isolation-tab-content.tsx
kantord: not actually added here, this is old code. but fair enough, it should be fixed if possible

peppescg:
> > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > 
> > > 
> > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > 
> > 
> > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> 
> I don't understand I added the validation onChange....it should work 🤔

@kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
```
 reValidateMode: 'onChange',
 mode: 'onChange',

peppescg (approved):
nice great works 🚢

kantord:
> > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > 
> > > > 
> > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > 
> > > 
> > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > 
> > 
> > I don't understand I added the validation onChange....it should work 🤔
> 
> @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> 
> ```
>  reValidateMode: 'onChange',
>  mode: 'onChange',
> ```

this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:


https://github.com/user-attachments/assets/e487d38b-2f09-4683-9e2d-76857152df7d

but this bug seems to exist also on the `main` branch. This is the behavior on the main branch: 


https://github.com/user-attachments/assets/b5598993-1f15-41f8-aeb0-5053dbf2cda9

peppescg:
> > > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > > 
> > > > > 
> > > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > > 
> > > > 
> > > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > > 
> > > 
> > > I don't understand I added the validation onChange....it should work 🤔
> > 
> > 
> > @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> > ```
> >  reValidateMode: 'onChange',
> >  mode: 'onChange',
> > ```
> 
> this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:
> 
>  Record_select-area_20250725163632.mp4 
> but this bug seems to exist also on the `main` branch. This is the behavior on the main branch:
> 
>  Record_select-area_20250725163752.mp4

did you checking the difference with registry? where there is this implementation of validation on change and it works?

kantord:
> > > > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > > > 
> > > > > > 
> > > > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > > > 
> > > > > 
> > > > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > > > 
> > > > 
> > > > I don't understand I added the validation onChange....it should work 🤔
> > > 
> > > 
> > > @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> > > ```
> > >  reValidateMode: 'onChange',
> > >  mode: 'onChange',
> > > ```
> > 
> > 
> > this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:
> > Record_select-area_20250725163632.mp4
> > but this bug seems to exist also on the `main` branch. This is the behavior on the main branch:
> > Record_select-area_20250725163752.mp4
> 
> did you checking the difference with registry? where there is this implementation of validation on change and it works?

both videos were from registry. the second video is from the current main branch. So there is no difference, this is the behavior for registry, and was not broken on this branch :thinking:

peppescg:
> > > > > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > > > > 
> > > > > > > 
> > > > > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > > > > 
> > > > > > 
> > > > > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > > > > 
> > > > > 
> > > > > I don't understand I added the validation onChange....it should work 🤔
> > > > 
> > > > 
> > > > @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> > > > ```
> > > >  reValidateMode: 'onChange',
> > > >  mode: 'onChange',
> > > > ```
> > > 
> > > 
> > > this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:
> > > Record_select-area_20250725163632.mp4
> > > but this bug seems to exist also on the `main` branch. This is the behavior on the main branch:
> > > Record_select-area_20250725163752.mp4
> > 
> > 
> > did you checking the difference with registry? where there is this implementation of validation on change and it works?
> 
> both videos were from registry. the second video is from the current main branch. So there is no difference, this is the behavior for registry, and was not broken on this branch 🤔

this is from main

https://github.com/user-attachments/assets/cd8696cb-eba4-4cff-9d89-4fd984d6e7af

kantord:
> > > > > > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > > > > > 
> > > > > > > > 
> > > > > > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > > > > > 
> > > > > > > 
> > > > > > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > > > > > 
> > > > > > 
> > > > > > I don't understand I added the validation onChange....it should work 🤔
> > > > > 
> > > > > 
> > > > > @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> > > > > ```
> > > > >  reValidateMode: 'onChange',
> > > > >  mode: 'onChange',
> > > > > ```
> > > > 
> > > > 
> > > > this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:
> > > > Record_select-area_20250725163632.mp4
> > > > but this bug seems to exist also on the `main` branch. This is the behavior on the main branch:
> > > > Record_select-area_20250725163752.mp4
> > > 
> > > 
> > > did you checking the difference with registry? where there is this implementation of validation on change and it works?
> > 
> > 
> > both videos were from registry. the second video is from the current main branch. So there is no difference, this is the behavior for registry, and was not broken on this branch 🤔
> 
> this is from main
> 
>  Kapture.2025-07-25.at.16.45.20.mp4

you need to submit the form first. Even on this branch, this bug only happens when you submit the form. And only the error coming from the form submission behaves this way. So it's  special edge case

peppescg:
> > > > > > > > > > from the video in the description it seems you submitted the form having a validation error for form field 🤔 how is it possible?
> > > > > > > > > 
> > > > > > > > > 
> > > > > > > > > hmmm, I don't think it's technically a validation error, maybe the problem is that it's not being validated until you press the button. let me check it out
> > > > > > > > 
> > > > > > > > 
> > > > > > > > yeah, the problem seems to be that it does not get revalidated without submitting the form. but this only seems to apply to errors that appeared when submitting the form, not ones that appeared as you were typing. let me look into it
> > > > > > > 
> > > > > > > 
> > > > > > > I don't understand I added the validation onChange....it should work 🤔
> > > > > > 
> > > > > > 
> > > > > > @kantord I guess you only need to add the below props on `renderer/src/features/mcp-servers/components/dialog-form-run-mcp-command.tsx` in order to fix the validation
> > > > > > ```
> > > > > >  reValidateMode: 'onChange',
> > > > > >  mode: 'onChange',
> > > > > > ```
> > > > > 
> > > > > 
> > > > > this did not change the behavior. but actually, the behavior seems to be consistent with what is happening on the registry MCPs:
> > > > > Record_select-area_20250725163632.mp4
> > > > > but this bug seems to exist also on the `main` branch. This is the behavior on the main branch:
> > > > > Record_select-area_20250725163752.mp4
> > > > 
> > > > 
> > > > did you checking the difference with registry? where there is this implementation of validation on change and it works?
> > > 
> > > 
> > > both videos were from registry. the second video is from the current main branch. So there is no difference, this is the behavior for registry, and was not broken on this branch 🤔
> > 
> > 
> > this is from main
> > Kapture.2025-07-25.at.16.45.20.mp4
> 
> you need to submit the form first. Even on this branch, this bug only happens when you submit the form. And only the error coming from the form submission behaves this way. So it's special edge case

let me take a look to the registry one, the fix would be the same for the custom one

samuv:
FYI: IDK if you want to fix in this PR, with default height and width of the app, the modal height it seems not fitting

Screenshot 2025-07-25 at 16 03 12

kantord:
> FYI: IDK if you want to fix in this PR, with default height and width of the app, the modal height it seems not fitting
> 
> Screenshot 2025-07-25 at 16 03 12

oh, that is not great. I never notice this, because i3wm does not have a concept of default window so I never see that :smile_cat: 

But I guess this should be solved by some hard limit on modal heights across the app, expressed in pixels/`calc()` with `!important` so it cannot be accidentally overridden?

peppescg:
@kantord free feel to merge we can address in next PRs, cause it is not related to your current changes

samuv:
> > FYI: IDK if you want to fix in this PR, with default height and width of the app, the modal height it seems not fitting
> > Screenshot 2025-07-25 at 16 03 12
> 
> oh, that is not great. I never notice this, because i3wm does not have a concept of default window so I never see that 😸
> 
> But I guess this should be solved by some hard limit on modal heights across the app, expressed in pixels/`calc()` with `!important` so it cannot be accidentally overridden?

mmm I think you just need to take in consideration the height of the tabs in the wrapper of the content
`
className="relative max-h-[65dvh] space-y-4 overflow-y-auto px-6"
`
that take care of the scroll

PR #675, opened 2025-07-29
fix: removed accelerator shortcut for quit app in tray

peppescg:
We are using `CMD/CTRL + Q` for hiding the window, the accelerator shortcut is wrong, we can remove it


main/src/system-tray.ts
```diff
@@ -159,7 +159,6 @@ const createShowMenuItem = () => ({
 
 const createHideMenuItem = () => ({
   label: 'Hide Window',
-  accelerator: 'CmdOrCtrl+H',
```
samuv: wrong item

peppescg: right

PR #690, opened 2025-07-31
fix: remove thv check on e2e gh action

peppescg:
Every time we used a `thv` command, using the CLI, we are sending an heart beat to telemetry.
In the `e2e` action we are checking if thv is available running `thv list`, this generates at every run a new ID, that it will be considered as a new user at every run.
In the `e2e` we don't need to check if thv is available, there are several check in the application.

### DIFF
--- a/.github/workflows/_e2e.yml
+++ b/.github/workflows/_e2e.yml
@@ -43,11 +43,7 @@ jobs:
         run: |
           sudo systemctl start docker
           sudo systemctl status docker
-
-      - name: Verify that `thv` binary works
-        run: |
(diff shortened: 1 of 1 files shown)


kantord (approved):
Looks good to me. Should we do something else to disable telemetry though?

peppescg:
> Looks good to me. Should we do something else to disable telemetry though?


The issue is not related to UI, but to CLI telemetry. Every time a CLI command is used is sending heart beat, and because it is a gh action is sending with new ID at every run

kantord (approved):

PR #701, opened 2025-08-06
feat: support mounting storage volumes

kantord:
Support storage volumes configuring by default read-write access and `ro` for read-only mode:
```
 '/path/to/host:/path/to/container',
 '/path/to/host:/path/to/container:ro',
 ```
The current logic between custom run and registry form is handled in two different components, it should be shared in some way. I am going to tackle that in a dedicated PR, cause the amount of change will be huge.

**Configure volumes from registry form**

https://github.com/user-attachments/assets/cc5bac39-8355-4c05-8122-fb3668938796

**Configure volumes from custom run**

https://github.com/user-attachments/assets/1922c8a4-423c-4d72-b6a6-d7fd392e14ad

**Docker inspect, container configuration**
Screenshot 2025-08-08 at 21 37 05

### CODE REVIEW

File: renderer/src/features/registry-servers/components/storage-volumes-field.tsx [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    <FormItem className="mb-10">
+      <FormLabel>Storage volumes</FormLabel>
+      <FormDescription>
+        Provide the MCP server access to a local folder. Optionally specific
+        individual files.
+      </FormDescription>
+    </FormItem>
+  )
+}
```
peppescg: Is it a leftover🤔 ? it seems duplicated
`renderer/src/common/components/workloads/storage-volumes-field.tsx`

kantord: yeah, it should not be duplicated. indeed it was caught by snip

kantord: knip. or whatever

renderer/src/features/network-isolation/components/network-isolation-tab-content.tsx
```diff
@@ -58,9 +64,44 @@ export function NetworkIsolationTabContent({
                       label="Allowed hosts"
                       inputLabelPrefix="Host"
                       addButtonText="Add a host"
-                      tooltipContent="Specify domain names or IP addresses. To include subdomains, use a leading period (“.”)"
+                      tooltipContent={`Specify domain names or IP addresses. To include subdomains, use a leading period (".")`}
```
peppescg, line 67: why the backtick here 🤔, this is the LLM for sure 😆  ?

peppescg: or right because the quote, it needs the escape the quote or use the backtick, so it is ok

kantord: I didn't actually even change that part, just moved around/copy pasted it

renderer/src/features/registry-servers/components/volumes-field.tsx
```diff
@@ -0,0 +1,148 @@
+import { FormControl, FormField, FormItem } from '@/common/components/ui/form'
```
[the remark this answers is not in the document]
peppescg: refactor to FormFieldVolumes

renderer/src/common/lib/utils.ts
```diff
@@ -12,3 +12,18 @@ export function cn(...inputs: ClassValue[]) {
...
+    host: string
+    container: string
+    accessMode?: 'ro' | 'rw'
+  }>
+): Array<string> {
+  return volumes
+    .filter((volume) => volume.host && volume.container)
+    .map(
+      (volume) =>
+        `${volume.host}:${volume.container}${volume.accessMode === 'ro' ? ':ro' : ''}`
```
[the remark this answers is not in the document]
peppescg: i think it is worse in terms of readability

eleftherias (approved):

PR #719, opened 2025-08-08
feat: configure custom registry

peppescg:
### Overview
This PR introduces a complete custom registry configuration interface in the Settings panel. Users can now choose between three registry types:
- **Default ToolHive registry** 
- **Remote registry** via HTTP/HTTPS URL
- **Local registry** via JSON file path

https://github.com/user-attachments/assets/aeb5eeb6-325f-4857-9e05-443d1eb04dba


renderer/src/common/components/settings/registry/registry-tab.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const registryData = registry?.data
+
+  const { mutateAsync: updateRegistry, isPending: isPendingUpdate } =
+    useToastMutation({
+      mutationFn: async (data: RegistryFormData) => {
+        const body =
+          data.type === 'url'
+            ? { url: data.source?.trim() }
+            : { local_path: data.source?.trim() }
+        await delay(500)
```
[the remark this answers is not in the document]
peppescg: it is needed for showing a better experience during the api call showing the loading state

renderer/src/common/components/settings/registry/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+    {
+      message: 'Registry must be a .json file',
+      path: ['source'],
+    }
+  )
+  .refine(
+    (data) => {
+      if (data.type === 'url' && data.source) {
+        if (
+          !data.source.startsWith('http://') &&
```
eleftherias: It needs to be https or it will fail when trying to fetch the registry. This is a bug in the backend that we don't validate it.

peppescg: got it

peppescg: @eleftherias do I need to add the `--allow-private-i` checkbox in ui in order to allow http 🤔 wdyt?

eleftherias: I think it will pollute the UI, but it's up to you. I think for dev purposes you could just point to a file, instead of using localhost on http.

peppescg: ok let's leave it as is, we can refactor this page in a second step

renderer/src/common/components/settings/registry/registry-type-field.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <FormField
+      control={form.control}
+      name="type"
+      render={({ field }) => (
+        <FormItem className="w-full">
+          <FormLabel required>Registry Type</FormLabel>
+          <FormDescription>
+            Choose between ToolHive default registry, a custom remote registry
+            (HTTP/HTTPS URL), or a custom local registry file.
```
eleftherias:
```suggestion
            (HTTPS URL), or a custom local registry file.
```

eleftherias: This one snuck in. I removed my suggestion because if it's merged then I can't review this PR

eleftherias: Don't merge

peppescg: ops forgot it, thanks

PR #720, opened 2025-08-08
fix: switch renovate to track docker images instead of github releases

JAORMX:
Switches Renovate configuration from tracking GitHub releases to tracking Docker container images at `ghcr.io/stacklok/toolhive`.

## Changes:
- Changed datasource from `github-releases` to `docker`
- Updated `depNameTemplate` to use the full Docker image path `ghcr.io/stacklok/toolhive`
- Removed unnecessary `packageNameTemplate` and `versioningTemplate`
- Added `extractVersionTemplate` to handle the `v` prefix in Docker tags
- Updated regex to handle versions with or without the `v` prefix
- Simplified `managerFilePatterns` pattern

This should properly track ToolHive Docker images which are tagged as `v0.2.5`, etc.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -9,14 +9,13 @@
     {
       "customType": "regex",
       "description": "Update ToolHive version in constants.ts",
-      "managerFilePatterns": ["^utils/constants\\.ts$"],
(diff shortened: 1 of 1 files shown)


peppescg (approved):
🚀

Issue #742, opened 2025-08-14
[Bug] Registry URL validation is inconsistent between ToolHive and ToolHive Studio


nick-terry:
### Bug Description

The Registry form schema enforces a requirement that the registry URL end with the `.json` suffix. ([code](https://github.com/stacklok/toolhive-studio/blob/0cbbde7effb67777d5ab4e4fde3b382cd28fc580/renderer/src/common/components/settings/registry/schema.ts#L20-L31)) This is inconsistent with the ToolHive CLI, which imposes so such requirement. This can prevent the ToolHive Studio user from configuring the registry URL to a value which functions correctly otherwise.

### Steps to Reproduce

1. Consider a URL which returns a valid ToolHive registry, but does not have a `.json` suffix
2. Run the command `thv config set-registry {{url}}` to set the registry URL
3. Go to Settings (⚙️ icon) > Registry and check the `Registry URL` field for validation.

### Expected Behavior

The validation of the `Registry URL` field succeeds, with no error message shown. The user should be able to configure a new Registry URL which does not end with `.json` via the UI.

### Actual Behavior

The validation of the `Registry URL` field does not succeed, as shown by the error message. If the field is validated (i.e. the URL ends with `.json`), then the user will be unable to configure a new Registry URL which does not end with `.json` via the UI.

### Priority

Low

### Environment

- OS: macOS 15.3.1 
- Macbook Pro M1 Max
- App Version 0.4.0

### Additional Context

<img width="1032" height="681" alt="Image" src="https://github.com/user-attachments/assets/867a4165-d24a-45d4-a231-cd3fadc720a1" />

samuv:
thank you @nick-terry  I will take care of it

samuv:
the fix has been merged into main, it should be included in the next release

PR #746, opened 2025-08-18
fix(settings): remove json ext validation from registry remote URL

samuv:
This is relate to the [issue](https://github.com/stacklok/toolhive-studio/issues/742), We don’t need to enforce .json in this case, only when we’re working directly with a file

### DIFF
--- a/renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
+++ b/renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
@@ -65,6 +65,7 @@ describe('RegistryTab', () => {
       url: 'https://domain.com/registry.json',
     },
     {
+      // reproduces bug: https://github.com/stacklok/toolhive-studio/issues/742
       description: 'without .json extension',
       url: 'https://domain.com/registry',
     },


kantord:
approved, though: in this case it could be wise to add a test case that verifies this behavior and link the original issue. that way, if someone wants to change this behavior in the future, they will get a test failure and a direct link to further context

samuv:
> approved, though: in this case it could be wise to add a test case that verifies this behavior and link the original issue. that way, if someone wants to change this behavior in the future, they will get a test failure and a direct link to further context

Yeah good point, I just checked and we have the unit test for registry-tab, let me extended with and without .json

renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
```diff
@@ -59,7 +59,16 @@ describe('RegistryTab', () => {
...
   })
 
-  it('handles remote registry configuration', async () => {
+  it.each([
+    {
+      description: 'with .json extension',
+      url: 'https://domain.com/registry.json',
+    },
+    {
+      description: 'without .json extension',
```
kantord, line 69:
```suggestion
      // reproduces bug: https://github.com/stacklok/toolhive-studio/issues/742
      description: 'without .json extension',
```

samuv: What’s the benefit? The issue is already attached to the PR that was linked in the squash merge commit.
I usually try to keep the codebase leaner.

But we can add it anyway if you’d prefer.

samuv: added in any case

PR #748, opened 2025-08-18
feat(feature-flags): implementation with electron store and binded to the Window

samuv:
Replaces the placeholder that we had for the feature flag implementation with a electron-store solution that provides persistent storage. The new system stores flags securely in the main process using electron-store, exposes them through IPC to the renderer, and includes a synchronous `window.FeatureFlag` API for easy testing in developer tools (`FeatureFlag.playground.enable()`). 

I added already the playground flag and with this implementation if we are going to move it in settings we can do easily

https://github.com/user-attachments/assets/0037ec41-8f22-4b7d-b654-6af99932a03a

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -16,7 +16,6 @@ export default {
     '@/*': ['renderer/src/*'],
   },
   ignore: [
-    'renderer/src/feature-flags/index.ts',
     'renderer/src/route-tree.gen.ts',
     'renderer/src/routes/store.tsx',
     'renderer/src/routes/clients.tsx',


main/src/main.ts
```diff
@@ -704,3 +711,20 @@ ipcMain.handle('dialog:select-folder', async () => {
   if (result.canceled || result.filePaths.length === 0) return null
   return result.filePaths[0]
 })
+
+// Feature flag IPC handlers
+ipcMain.handle('feature-flags:get', (_event, key: FeatureFlagKey): boolean => {
```
kantord, line 716: I guess the setting could also be implemented using a single IPC handler like this

```ts
setFeatureFlags({ foo: true, bar: false })
```

Where the implementation would basically match this pseudo code:

```ts
function setFeatureFlags(newFlags) {
    store.set('flags', {
       ...store.get('flags'),
       ...newFlags,
    })

    return store.get('flags')
}
```

_techinicall_ you could call this with an empty object and that would also give you the "get all" functionality

kantord: regarding **getting** I am tempted to think that IPC handlers may not be the best solution for this? I mean, the values could be pre-filled in the `window` object somehow instead :thinking: I guess when the value of a feature flag changes, we can do a full reload of the window anyways, just to also simplify the logic

samuv: Good point, preloading flags into window at startup would definitely simplify reads, and I agree it could work if we’re okay with reloading the whole window after any change.

That said, I went with IPCs because we need persistence across sessions (via electron-store), and this setup makes it easy to expose flags to users via a settings UI later on. In that scenario, reloading the window after every change would feel a bit harsh, especially if we want to toggle things on/off and see the effect immediately.

The IPC API also gives us flexibility, for example, the renderer can update flags or fetch them at any time, without having to reload. It’s a bit more boilerplate, but it keeps things decoupled and makes the logic easier to extend

kantord (approved):

PR #762, opened 2025-08-20
refactor: reuse array field component for environment variables

kantord:
a good part of this is vibe coded, see thread about it: https://stacklok.slack.com/archives/C0958A56ZEE/p1755677759722769

This PR:

- removes some duplicated logic by rolling out reusable field from the volume mount feature for secrets and env vars (only for custom servers)
- as a result, makes the app look more visually consistent, because the visual implementation for those features was outdated

### Before

Screenshot_select-area_20250820114200

### After

Screenshot_select-area_20250820114110


renderer/src/common/components/secrets/secret-store-combobox.tsx
```diff
@@ -0,0 +1,97 @@
...
+                ? data.keys
+                    .sort((a, b) => (a.key ?? '')?.localeCompare(b.key ?? ''))
+                    .map((secret) => (
+                      <CommandItem
+                        key={secret.key}
+                        value={secret.key}
+                        className="font-mono"
+                        onSelect={async (selected) => {
+                          onChange(selected)
+                          await delay(150)
```
samuv, line 77: why do we need a delay ?

samuv (approved):

renderer/src/common/components/secrets/secret-store-combobox.tsx
kantord: sorry, I forgot to reply. The answer is I have no idea, this was part of the original implementation and I just blindly copied it.

you can see it here in the current implementation: https://github.com/stacklok/toolhive-studio/blob/c902b314e76685bd468d37aa9ce8fa1c552ceb18/renderer/src/common/components/secrets/form-combobox-secrets-store.tsx#L103

samuv: maybe we can remove it?

kantord: we can try, but I would separate that effort

PR #765, opened 2025-08-20
fix: add query all to get on graceful-exit

samuv:
as per title


samuv:
ops, I don't think we need this

PR #770, opened 2025-08-21
feat(graceful-exit): improve with better status checking and testing

samuv:
This PR enhances the `graceful-exit` functionality by updating the polling logic to properly handle all final server statuses that now we have with the latest API update (stopped, error, unknown, unhealthy) instead of just checking for running servers. Added also some unit tests

### DIFF
--- a/main/src/tests/graceful-exit.test.ts
+++ b/main/src/tests/graceful-exit.test.ts
@@ -104,7 +104,7 @@ describe('graceful-exit', () => {
       { name: 'server2', status: 'stopped', port: 3002 },
     ]
 
-    it('should handle no running servers gracefully', async () => {
+    it('handles no running servers gracefully', async () => {
       mockGetApiV1BetaWorkloads.mockResolvedValue(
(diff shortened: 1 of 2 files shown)


main/src/tests/graceful-exit.test.ts
```diff
@@ -0,0 +1,351 @@
...
+
+      expect(mockLog.info).toHaveBeenCalledWith(
+        'Still waiting for 1 servers to reach final state: server1(stopping)'
+      )
+      expect(mockLog.info).toHaveBeenCalledWith(
+        'All servers have reached final state'
+      )
+    })
+
+    it('should respect polling intervals', async () => {
```
kantord: my nit pick is this 

```suggestion
    it('respects polling intervals', async () => {
```

kantord (dismissed):

main/src/tests/graceful-exit.test.ts
samuv: ops yes, I change the naming

PR #782, opened 2025-08-25
feat(feature-flag): Add new page and starting logic for customize tools 

samuv:
as per title, I just opened this under a feature flag, this is still WIP, it introduces:

- dedicated page for customizing tools
- dedicated IPC function to retrieve the tools (temporary), reusing the logic we had in the playground while waiting for the dedicated endpoint from thv
- components for showing the tool list
- hook for the update (not working at the moment)
- feature flag (customize_tools)

With the flag enable:

https://github.com/user-attachments/assets/f122689f-a68a-480d-b03e-171dfd263b55

### DIFF
--- a/main/src/utils/mcp-tools.ts
+++ b/main/src/utils/mcp-tools.ts
@@ -1,0 +1,133 @@
+import {
+  experimental_createMCPClient as createMCPClient,
+  type experimental_MCPClientConfig as MCPClientConfig,
+  type Tool,
+} from 'ai'
+import { Experimental_StdioMCPTransport as StdioMCPTransport } from 'ai/mcp-stdio'
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js'
+import type { CoreWorkload } from '@api/types.gen'
+import log from '../logger'
+
+// Interface for MCP tool definition from client
+export interface McpToolDefinition {
+  description?: string
+  inputSchema?: Tool['inputSchema']
+}
+
+// Type guard to check if an object is a valid MCP tool definition
+export function isMcpToolDefinition(obj: Tool): obj is McpToolDefinition {
+  if (!obj || typeof obj !== 'object' || obj === null) return false
+
+  const tool = obj
+
+  // Description should be string if present
+  if (
+    'description' in tool &&
+    tool.description !== undefined &&
+    typeof tool.description !== 'string'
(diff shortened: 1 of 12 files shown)


main/src/utils/mcp-tools.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { CoreWorkload } from '@api/types.gen'
+import log from '../logger'
+
+// Interface for MCP tool definition from client
+export interface McpToolDefinition {
+  description?: string
+  inputSchema?: Tool['inputSchema']
+}
+
+// Type guard to check if an object is a valid MCP tool definition
```
kantord: this type of comment could be a jsdoc comment, or even removed

main/src/utils/mcp-tools.ts
```diff
@@ -0,0 +1,133 @@
...
+  experimental_createMCPClient as createMCPClient,
+  type experimental_MCPClientConfig as MCPClientConfig,
+  type Tool,
+} from 'ai'
+import { Experimental_StdioMCPTransport as StdioMCPTransport } from 'ai/mcp-stdio'
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js'
+import type { CoreWorkload } from '@api/types.gen'
+import log from '../logger'
+
+// Interface for MCP tool definition from client
```
kantord:
```suggestion
```

probably redudant, but if it's important, it should be jsdoc

main/src/utils/mcp-tools.ts
```diff
@@ -0,0 +1,133 @@
...
+      ) {
+        return false
+      }
+    }
+  }
+
+  return true
+}
+
+// Create transport configuration based on workload type
```
kantord:
```suggestion
```

seems like a redundant comment

main/src/utils/mcp-tools.ts
samuv: they are not really important at the moment

main/src/utils/mcp-tools.ts
samuv: same here

main/src/utils/mcp-tools.ts
samuv: sometimes I ask AI to move some code or change one thing and this is the result :D 

removed in any case

PR #783, opened 2025-08-25
feat: allow user to create groups

kantord:
This adds the "create group" feature. This feature is still behind the `groups` feature flag, just like all other groups related features.

### video

https://github.com/user-attachments/assets/a12939bd-2332-4f15-a4f3-33e9bfa3771b

### Implementation

For implementing a modal, I added a helper hook similar to `useConfirm` that makes it easier to describe the workflow. I chose to implement it in such a way that this hook basically handles the state management and such boilerplate. But I chose to not have it "generate" a form, since this would be extra logic and would only create limitations.

So this way the form can be "passed in" and fully customized, but also easy to refactor (for instance, easy to convert to an independent page instead of a popup/modal)

~The controversial part is that I used formik, instead of RHF. This is because we had a bad experience with RHF for our use cases. @peppescg had good experiences with formik, so I decided that I don't want to add another usecase for RHF and I just introduced formik instead.~

Note: instead of a dropdown button, I used a simple button for now, because special features (like importing a group) are not yet supported in this pr


renderer/src/common/contexts/prompt/form-prompt-dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface ReactHookFormPromptDialogProps {
+  isOpen: boolean
+  config: ReactHookFormPromptConfig<Record<string, unknown>>
+  onSubmit: (data: Record<string, unknown>) => void
+  onCancel: () => void
+  onOpenChange: (open: boolean) => void
+}
+
+export function ReactHookFormPromptDialog({
```
peppescg: The name ReactHook is measleading (and probably react redundant 🤔  ) considering that it is not a hook

I would use something like `FormDialog` if the goal is to have a generic dialog form

renderer/src/common/contexts/prompt/provider.tsx
```diff
@@ -0,0 +1,66 @@
+import { useState, type ReactNode } from 'react'
+import { PromptContext, type ReactHookFormPromptConfig } from '.'
+import { ReactHookFormPromptDialog } from './form-prompt-dialog'
+
+export function PromptProvider({ children }: { children: ReactNode }) {
```
peppescg, line 5: I am missing the term prompt 🤔 
Anyway why a context/provider approach is needed?

renderer/src/common/contexts/prompt/form-prompt-dialog.tsx
kantord: true

peppescg (commented):
I am missing the use of context/provider here, which kind of issue is solving 🤔 ? 
Is the idea to have only one unique component replacing all the forms that are currently included in dialog ( run custom/registry mcp forms?)

the hook use-prompt is a bit misleading cause want to provide only the configuration form

renderer/src/common/contexts/prompt/provider.tsx
kantord: this hook provider basically implements a shadcn ui equivalent of the p[rompt function in JavaScript](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt). That is where the naming comes from, it just keeps effectively the same usage and naming, but allows us to use shadcn components and any react node for the texts. The only actual difference is that this implementation uses a Promise, not a hard blocker.


The context provider simply moves the prompt()-related state management into a single place. This makes sense in my opinion because:

- Otherwise there isn't a very trivial way to avoid code duplication for this logic without requiring a more complex component structure of components. Since there are actually a few edge cases here, just copy pasting the code would be dangerous and even just being able to test the edge cases in isolation is valuable
- There can only be one prompt visible at the same time, so that is another reason to have it in a single place
- Because useConfirm already uses this exact same approach and it has been working well. We also used the same approach in multiple projects and never had any issues with it as far as I know 

Also, while I didn't include it in this PR (to keep it smaller), it will give us similar refactor opportunity that we had with notify and confirm. For instance, we could convert `useToastMutation` into a more generic "mutation flow" handler, for instance, we could even include the input there:

```tsx
// pseudo code
const handleCreateGroup = useMutationFlow({
  prompt: {
     fieldName: "Name",
     fieldType: "text",
     initialValue: "",
  },
  mutationFn: postApiV1BetaGroups,
  successMessage: name => `Group ${name} successfully created`,
})

// ...etc...

<Button onClick={handleCreateGroup} />
```

**not sure if we will do this refactor, but I think it's a good way to demonstrate what kind of flexibility having the context provider gives**

peppescg: but it is a dialog component not a prompt 🤔

kantord: it's only a technical difference. well, I'm not sure how `prompt()` works on Mac OS, but on Linux Chrome, Firefox, etc it opens a modal dialog

kantord: so difference is only shadcn dialog vs native dialog. Basically `<button />` vs. `<Button />`

renderer/src/common/hooks/__tests__/use-prompt.test.tsx
```diff
@@ -0,0 +1,388 @@
...
+  }
+
+  return (
+    <div data-testid={testId}>
+      <button onClick={handleClick} data-testid="trigger-button">
+        {buttonLabel}
+      </button>
+      {result !== undefined && (
+        <div data-testid="result">
+          {result === null ? 'Cancelled' : `Result: ${JSON.stringify(result)}`}
```
peppescg: is the JSON.stringify a leftover?

renderer/src/common/contexts/prompt/provider.tsx
peppescg: I am missing the piece of code where it is using the native prompt, cause I see only the Dialog from shadcn 🤔

peppescg: what I mean, this is a simple dialog 

Screenshot 2025-08-28 at 16 59 13

kantord: no no, what I mean is, this implements something that is equivalent in usage/business logic to the native prompt, but implements it using shadcnui to match the design.

Simple pseudo code to compare

*Native*

```tsx
const groupName = prompt("Please enter group name", initialValue)

if (groupName) {
   createGroupMutation.mutate(groupName)
}
```

*Our implementation*

```tsx
const prompt = usePrompt()
const groupName = await prompt("Please enter group name", initialValue)

if (groupName) {
   createGroupMutation.mutate(groupName)
}
```

peppescg:
a couple of issues:
- I see an hard reload after the first click on the created group, this is weird, the group sidebar group should be decoupled from its content IMO
- the create button should be disabled in case of invalid form

https://github.com/user-attachments/assets/accfcfe5-f676-42a9-908c-99e74ea14b4b

kantord:
> a couple of issues:
> 
> * I see an hard reload after the first click on the created group, this is weird, the group sidebar should be decoupled from the its content IMO
> * the create button should be disabled in case of invalid form
> 
>  Kapture.2025-08-28.at.17.00.30.mp4

yeah, I noticed the hard reload as well, so I was going to look at why it's happening, but I guess it's not introduced in this PR and does not affect users who don't have the flag enabled. I will definitely address it before we enabled the feature flag.

Create button should be disabled for invalid form, that is a good point! Let me address that right now

renderer/src/common/contexts/prompt/provider.tsx
peppescg: I am honestly a bit confused, it seems unnecessary complication cause it is a simple dialog with a form that includes an input field. The things that confuse me are, the needs of use a context/provider for showing a simple dialog at root level and the term prompt 🤔 that is misleading comparing the component that is used under the hood

renderer/src/common/hooks/__tests__/use-prompt.test.tsx
kantord: more like some hacky stuff than leftover, but I removed it anyway

peppescg (approved):

PR #799, opened 2025-08-28
chore: trigger mirror-to-gh-pages action after release completed

peppescg:
Based on [Running a workflow based on the conclusion of another workflow](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow) we can run `mirror-to-gh-pages` action after the release gh action is completed and assets generated

### DIFF
--- a/.github/workflows/mirror-to-gh-pages.yml
+++ b/.github/workflows/mirror-to-gh-pages.yml
@@ -23,7 +23,12 @@ jobs:
       - name: Get release info
         id: release
         run: |
-          TAG="${{ github.event.inputs.tag || github.event.release.tag_name }}"
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG="${{ github.event.inputs.tag }}"
+          else
+            # Get the latest release tag when triggered by workflow_run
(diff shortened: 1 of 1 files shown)


samuv (approved):

.github/workflows/mirror-to-gh-pages.yml
```diff
@@ -15,11 +17,18 @@ permissions:
...
     steps:
       - name: Get release info
         id: release
         run: |
-          TAG="${{ github.event.inputs.tag || github.event.release.tag_name }}"
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            TAG="${{ github.event.inputs.tag }}"
+          else
+            # Get the latest release tag when triggered by workflow_run
+            TAG=$(gh api "repos/${{ github.repository }}/releases/latest" | jq -r '.tag_name')
```
samuv, line 30: or we could just make the tag mandatory in the workflow_dispatch 🤔

samuv (approved):

.github/workflows/mirror-to-gh-pages.yml
peppescg: yep, the cons that I see is that we manually need to set the tag but we already know that we want to do it for the latest one

PR #811, opened 2025-08-29
fix(network-isolation): skip hosts, ports if network isolation is not enabled

peppescg:
Currently the validation is always applied also if the network isolation is not enabled, we can skip the network isolation validation in case it is disabled.

Other fix:
- Fix fields format for `allowedHosts` and `allowedPorts` in case of empty values for both registry and custom form
- Fix custom form validation message for docker image, introducing union type 

**Registry form**

https://github.com/user-attachments/assets/13b574bc-93a4-4984-96b5-89857bed183c

**Custom form**

https://github.com/user-attachments/assets/9c8e7ef2-e4cb-4ff9-b032-caf387f9eb70

### DIFF
--- a/renderer/src/features/mcp-servers/lib/form-schema-run-mcp-server-with-command.ts
+++ b/renderer/src/features/mcp-servers/lib/form-schema-run-mcp-server-with-command.ts
@@ -1,8 +1,8 @@
 import type { CoreWorkload } from '@api/types.gen'
 import z from 'zod/v4'
 
-const getCommonFields = (workloads: CoreWorkload[]) =>
-  z.object({
+export const getFormSchemaRunMcpCommand = (workloads: CoreWorkload[]) => {
+  const baseFields = z.object({
     name: z
       .union([z.string(), z.undefined()])
       .transform((val) => val ?? '')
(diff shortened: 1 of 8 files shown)


kantord:
in the first video, something weird happens around 0:16, no? like it feels like clicking the network isolation toggle makes the error disappear, but does not disable network isolation?

peppescg:
> in the first video, something weird happens around 0:16, no? like it feels like clicking the network isolation toggle makes the error disappear, but does not disable network isolation?

😆 nope I deleted the `/` and after clicked on the switch component, it works properly

https://github.com/user-attachments/assets/90dbbbda-ab3d-4967-a5bf-0da8541a5afb

renderer/src/features/mcp-servers/lib/form-schema-run-mcp-server-with-command.ts
```diff
@@ -41,32 +41,20 @@ const getCommonFields = (workloads: CoreWorkload[]) =>
...
+        })
+      )
+      .optional(),
+    allowedPorts: z
+      .array(
+        z.object({
+          value: z.string(),
+        })
+      )
+      .optional(),
```
samuv, line 57: are we passing those in the payload even if networkIsolation is disabled, is that why they’re defined in baseFields in any case?

renderer/src/features/registry-servers/lib/get-form-schema-run-from-registry.ts
```diff
@@ -64,76 +64,98 @@ export function getFormSchemaRunFromRegistry({
...
-      .min(1, 'Server name is required')
-      .refine(
-        (value) => !workloads.some((w) => w.name === value),
-        'This name is already in use'
-      ),
-    cmd_arguments: z.array(z.string()).optional(),
-    secrets: z
-      .object({
-        name: secretNameSchema,
+  return z
```
samuv, line 67: not the scope of this PR, but for the future we should definitely find a way to create a composer for this validation based on the form type and centralize the testing in one place

samuv (approved):
code looks good!

renderer/src/features/mcp-servers/lib/form-schema-run-mcp-server-with-command.ts
peppescg: there is a function before the api call, `prepareCreateWorkloadData` with its unit test that filter out network isolation fields in case it is disabled.
The reason why they are defined in the base field is cause they are defined in the defaultValue in the form state

renderer/src/features/registry-servers/lib/get-form-schema-run-from-registry.ts
peppescg: Agreed. So, there are a couple of stuff here that are a mess, first of all duplication of the validation, we are using the same validation and logic in two different files, custom run and registry form, we could unify and avoid duplication, and because I wasted time this morning on adding the same validation two times I think it is time to refactor and unify. Having a composer by formType will simplify the logic. I created this [issue](https://github.com/stacklok/toolhive-studio/issues/813), I am going to work on this soon

PR #812, opened 2025-08-29
fix: smooth navigation between groups

kantord:


renderer/src/features/mcp-servers/components/groups-manager/index.tsx
```diff
@@ -1,15 +1,13 @@
...
+import { useParams } from '@tanstack/react-router'
 import { AddGroupButton } from './add-group-button'
 import { GroupList } from './group-list'
 
 export function GroupsManager(): ReactElement {
-  const router = useRouterState({ select: (s) => s.location.search }) as Record<
-    string,
-    unknown
-  >
+  const params = useParams({ from: '/group/$groupName' })
```
peppescg, line 9: does it should have a `default` name for default group?

renderer/src/routes/group.$groupName.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const [isRunWithCommandOpen, setIsRunWithCommandOpen] = useState(false)
+  const { mutateAsync, isPending } = useMutationRestartServerAtStartup()
+  const hasProcessedShutdown = useRef(false)
+
+  // Redirect to default group if groups feature is disabled
+  useEffect(() => {
+    if (!showSidebar) {
+      navigate({ to: '/' })
+    }
+  }, [navigate, showSidebar])
```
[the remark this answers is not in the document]
peppescg: does it make sense 🤔 it seems so

peppescg (commented):
the video seems work as expected 🚀 
I left a couple of comments

renderer/src/routes/group.$groupName.tsx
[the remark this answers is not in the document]
kantord: hmm actually, let me see, maybe this entire logic can be removed at this point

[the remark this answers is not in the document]
kantord: yup, actually, this entire logic can be removed, it works fine without it

renderer/src/features/mcp-servers/components/groups-manager/index.tsx
kantord: not sure what you mean? :thinking:

kantord: you just mean 

```tsx
  const currentGroupName = params.groupName ?? 'default'
```

? 

if that's the case I guess the value should always be provided through the URL so it should not be necessary :thinking:

peppescg: yep right, I mean if there isn't a groupName it should be default? I don't know if there is something specifically to do, maybe not

kantord: basically, in the url the group name is obligatory, so here it is guaranteed by the type system & the router.

there is actually a logic to automatically add it to the url tho, here:

https://github.com/stacklok/toolhive-studio/pull/812/files#diff-f3e475ba88d6c938d6af252bfcb65a1c08124ec9d256e359b5324db71e0a1ce0R3-R8

peppescg: right, the default group name

PR #820, opened 2025-09-01
feat: add existing server to existing group

kantord:
https://github.com/user-attachments/assets/d4242a4d-2ede-4f6c-b3c4-11ca947c264c


kantord:
:hankey: everything is failing now that I accepted the suggestion by github copilot ahahah let me fix that

edit: pushed a fix

renderer/src/features/mcp-servers/components/add-server-to-group-menu-item.tsx
```diff
@@ -0,0 +1,71 @@
...
+    )
+
+    if (result) {
+      try {
+        await updateGroupMutation.mutateAsync({
+          workloadName: serverName,
+          groupName: result.value,
+        })
+      } catch (error) {
+        console.error('Failed to copy server to group:', error)
```
peppescg, line 56: nit: we don't need console.log anymore, we need to decide if this could be an error log.error that our middleware needs to intercept, or a warn or info...take a look to the other errors handle

renderer/src/features/mcp-servers/hooks/use-groups.ts
```diff
@@ -0,0 +1,19 @@
...
+  return useQuery({
+    queryKey: ['api', 'v1beta', 'groups'],
+    queryFn: async () => {
+      const response = await getApiV1BetaGroups({
+        parseAs: 'text',
+        responseStyle: 'data',
+      })
+      const parsed =
+        typeof response === 'string' ? JSON.parse(response) : response
+      return parsed as V1GroupListResponse
```
peppescg, line 15: why this JSON.parse  🤔 ?

PR #821, opened 2025-09-01
feat(feature-flag): implement editing with custom server form

samuv:
in this PR:
- added a feature flag for editing
- adjusted the custom server form to be reused for editing
- adjusted the height of the custom form to avoid overflow

the APIs aren’t consistent enough, I’ll do a second step once the fixes are done/released

missing:

- pass the group on editing
- add optimistic updates when fetching the status while polling
- handle errors on GET so the edit form doesn’t break if the workload configuration isn’t available (edge case)

### DIFF
--- a/renderer/src/features/mcp-servers/components/form-fields-run-mcp-command.tsx
+++ b/renderer/src/features/mcp-servers/components/form-fields-run-mcp-command.tsx
@@ -22,11 +22,14 @@ import { CommandArgumentsField } from '@/common/components/workload-cmd-arg/comm
 
 export function FormFieldsRunMcpCommand({
   form,
+  isEditing = false,
 }: {
   form: UseFormReturn<FormSchemaRunMcpCommand>
+  isEditing?: boolean
 }) {
   const typeValue = form.watch('type')
   const protocolValue = form.watch('protocol') ?? 'npx'
+  const transportValue = form.watch('transport')
 
   return (
     <>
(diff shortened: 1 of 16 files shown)


renderer/src/features/mcp-servers/components/form-fields-run-mcp-command.tsx
```diff
@@ -103,8 +107,14 @@ export function FormFieldsRunMcpCommand({
             </div>
             <FormControl>
               <Select
-                onValueChange={field.onChange}
-                defaultValue={field.value}
+                onValueChange={(value) => {
+                  field.onChange(value)
+                  // Automatically set target_port to 0 for stdio transport
+                  if (value === 'stdio') {
```
peppescg, line 113: Is correct that selecting the other transport the target port is 0?

Screenshot 2025-09-02 at 09 45 23
Screenshot 2025-09-02 at 09 46 05

samuv: It’s 0 because of the default value of stdio that gets carried over, we could reset it on other selection tho, but the field is mandatory with that transport, and without the correct port it would fail

peppescg: yep my assumption was to reset in the other cases, in order to let the user specify it

PR #823, opened 2025-09-02
feat: implement group-specific design for client management

kantord:
https://github.com/user-attachments/assets/cafb55d5-f800-4115-9f36-3d190afb706a

This PR implements the group-specific design for client management. NOTE: this is not behind a feature flag. That is because the new design is compatible with the feature flag being disabled, i. e. users can use the new design to enable clients even before we release the rest of the group-related features.


renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
```diff
@@ -0,0 +1,324 @@
...
+  }
+  server.events.on('request:start', onStart)
+  return {
+    get: () => records,
+    stop: () => {
+      server.events.removeListener('request:start', onStart)
+    },
+  }
+}
+
```
peppescg, line 48: it seems weird to see this implementation here. I like the record payload that we had in Minder, I will add the same feature here in a dedicated util file. I mean it is a middleware

kantord: yeah, I was thinking also about simply copying the entire msw handling logic from Minder. that could include the automatic mocks even :thinking: 

What do you think about moving that existing system into this repo in a separate PR? Then we can simplify a few of the existing mock-based tests as well :thinking:

renderer/src/features/clients/hooks/use-manage-clients.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  const { data: groupsData } = useQuery({
+    queryKey: ['api', 'v1beta', 'groups'],
+    queryFn: async () => {
+      const response = await getApiV1BetaGroups({
+        parseAs: 'text',
+        responseStyle: 'data',
+      })
+      const parsed =
+        typeof response === 'string' ? JSON.parse(response) : response
```
peppescg: why JSON.parse is needed?

renderer/src/features/clients/hooks/use-manage-clients.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    const currentClients = await getApiV1BetaClients({
+      parseAs: 'text',
+      responseStyle: 'data',
+    })
+
+    const parsed =
+      typeof currentClients === 'string'
+        ? currentClients
+          ? JSON.parse(currentClients)
```
peppescg: the same above, the API should already support this

renderer/src/features/clients/hooks/use-manage-clients.ts
kantord: it's the same bug that I still need to figure out/raise to BE (seen elsewhere in the codebase)  it seems like sometimes the response is not automatically parsed and it a string instead of an object :thinking: 

must be connected to a mime type issue

renderer/src/features/clients/hooks/use-manage-clients.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  const removeClientFromGroup = async (
+    clientType: string,
+    groupName: string
+  ) => {
+    // Find the client in the installed clients list
+    const client = installedClients.find((c) => c.client_type === clientType)
+    if (!client) {
+      throw new Error(`Client ${clientType} is not installed or available`)
```
peppescg, line 127: is this error catched?

renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
peppescg: yep makes sense to do it in a dedicated Pr

renderer/src/features/clients/hooks/use-manage-clients.ts
peppescg: the mime type is per endpoint in BE? cause we solved it months ago, it is weird to see this here 🤔 
I mean the client api is the same of the previous clients page version, and it didn't have the `JSON.parse` wrapper I think, so it should be something related to this implementation?

kantord: let me check because I think we even updated the thv multiple times since I originally added that code

renderer/src/features/clients/hooks/use-manage-clients.ts
kantord: seems like not, good catch. albeit this should never happen in real life, i had to simulate it to see what happens :smile_cat: 

let me see where I can add error handling for this

renderer/src/features/clients/hooks/use-manage-clients.ts
kantord: cool, seems like i could remove this and everything works just fine!

renderer/src/features/clients/hooks/use-manage-clients.ts
kantord: ok cool, it's trivial to just reuse `useToastMutation` for this, so pretty much zero changes were needed.

Success behavior after this:


https://github.com/user-attachments/assets/02be2397-34b9-4902-856f-3a3d01012f83


Failure behavior after this:


https://github.com/user-attachments/assets/affcf0dc-86f3-4279-8d83-81b8c7814c79

kantord: (again, i had to simluate this error because normally it would never happen)

peppescg:
Screenshot 2025-09-08 at 15 01 30

there is an issue with the dialog, the content is not scrollable

kantord:
> Screenshot 2025-09-08 at 15 01 30
> there is an issue with the dialog, the content is not scrollable

oh wow. let me fix that in a quick follow up PR? (it is because of a generic wrapper, not something introduced in this PR)

PR #824, opened 2025-09-02
chore: bump thv to v0.2.15

kantord:


eleftherias:
@kantord FYI we ended up releasing 0.2.15 because 0.2.14 was buggy

kantord:
> @kantord FYI we ended up releasing 0.2.15 because 0.2.14 was buggy

:eyes:

PR #825, opened 2025-09-03
refactor: form schema validation

peppescg:
In order to accomplish a refactor that affects form schema validation, utility function and components in order to avoid duplication code, I will split the effort into several PRs.

This first PR will affect only schema validation, create a common schema validation reused for local mcp and registry mcp forms.

Basically the common schema is:
`form-schema-mcp.ts` -> `renderer/src/common/lib/form-schema-mcp.ts`

The registry form schema is defined here `renderer/src/features/registry-servers/lib/form-schema-registry-mcp.ts` reusing some common fields

The local mcp form schema is defined  here `renderer/src/features/mcp-servers/lib/form-schema-local-mcp.ts`.

In next PR I will introduce remote mcp, so I started to define local vs remote mcp also for form schema and components, but I would avoid to do all the changes in this PR, so I am trying to reduce as much has possible the effort

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -23,6 +23,8 @@ export default {
     'renderer/src/common/hooks/use-servers.ts',
     'renderer/src/common/lib/utils.ts',
     'renderer/src/common/components/ui/**',
+    'renderer/src/features/mcp-servers/components/local-mcp/**',
+    'renderer/src/common/lib/form-schema-mcp.ts',
     'renderer/src/vite-env.d.ts',
     'renderer/src/types/global.d.ts',
     'renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts',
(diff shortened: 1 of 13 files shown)


renderer/src/features/mcp-servers/lib/orchestrate-run-local-server.tsx
```diff
@@ -129,42 +117,49 @@ function transformTypeSpecificData(
...
-  data: FormSchemaRunMcpCommand,
+  data: FormSchemaLocalMcp,
   secrets: SecretsSecretParameter[] = []
 ): V1CreateRequest {
   const request = transformTypeSpecificData(data)
 
-  request.cmd_arguments = data.cmd_arguments || []
-  request.env_vars = mapEnvVars(data.envVars)
-  request.secrets = secrets
+  // Extract and transform fields with proper typing
```
kantord, line 125: nit pick, i guess comments that explain what the code does are not that useful. if there is anything to explain here, it should be maybe _why_ we need to do this

kantord (approved):

renderer/src/features/mcp-servers/lib/orchestrate-run-local-server.tsx
peppescg: yep it makes sense, let me grab this into the next PR, cause I refactor this file in the remote mcp PR

PR #835, opened 2025-09-05
refactor: mcp local server form

peppescg:
In order to split the effort related to refactor form and remote mcp, in this PR I rename a couple of component and removed the deprecated `dialog-form-run-mcp-command.tsx` one.


renderer/src/common/components/empty-state.tsx
```diff
@@ -6,11 +6,16 @@ const actionsStyle = cva('mx-auto mt-8', {
     actions: {
       1: '',
       2: 'grid grid-cols-2 gap-2',
+      3: 'grid grid-cols-3 gap-2',
```
kantord, line 9: could this just be replaced with a flex layout, removing the need to have a different grid pattern based on the number of actions?

in that case also actions could be a single react node, not an array :thinking: (we can omit `null` from ReactNode to force at least one item)

kantord (approved):

renderer/src/common/components/empty-state.tsx
peppescg: I added this cause we don't have yet the empty state that includes remote mcp from figma. So I let this change as last, I honestly don't like to have 3 buttons in place, it seems redundant honestly, I will circle back with James this as last item to solve

kantord: makes sense

PR #839, opened 2025-09-08
test: add mocker system from minder

kantord:
fixes https://github.com/stacklok/toolhive-studio/issues/851 

### This PR

- this PR takes the old mocker helpers/mock system from Minder Frontend :smile_cat: based on a comment from a previous groups-related PR
- this means that now, when there is not a custom-made MSW mock for endpoints, a new one is generated. The new mock will always use fake data in a structure that matches the JSON schema provided by the API. This mock can still be manually customized, but the process of this manual customization will be greatly simplified as you don't need to worry about the data structure
- this also allows us to use the same `recordRequests()` function to easily make assertions on the behavior of mutations
- The PR includes an example of a custom mock that was removed and automatically replaced by the automatic mock
- It also replaces changes the assertions in the "Manage Clients" button tests to use `recordRequests()` 

### Improvement opportunities
- we could generate typescript mocks instead of JSON. This was already identified as an improvement in Minder. What this would do is allow us to type-check the mocks + we would get autocomplete for them and the possibility to include some business logic and comments. Not something urgent, but I guess this refactor could even be fully vibe-coded. Still, too much for this first PR
- The mocks are just generated using a generic data faker, but we could actually pass the schema and the automated mock through an LLM to finalize it. This way, given the extra context from the schema documentation the LLM could probably generate perfect or almost perfect mocks instead of adding a bunch of lorem ipsum. This could be probably done using `gh` cli and the copilot endpoints so that extra configuration/api keys are not needed. Still, this is too much for this first PR
- I think some things, like file name/directory structure could be improved
- Pretty much all existing mocks could be replaced by the automated mocks, so we could probably remove the custom_mocks file. But that would be too big of a step for a single PR imo


peppescg (approved):
🚢

PR #844, opened 2025-09-09
fix: trigger mirror after build_and_release action

peppescg:


peppescg:
> yeah, in retrospect a workflow call makes more sense with the async nature of the release. I thought “completed” also took into account publishing the assets, but it’s independent

I forgot this PR 😆, I mean to ask a review. Let's see if it works

PR #873, opened 2025-09-15
test: implement type-checked test fixtures

kantord:
a little refactor I did yesterday for the auto mocking system. What it includes:

- type checking even for auto-generated fixtures. This means that you get appropriate type failures for the mocks when BE introduces changes to the schema and you don't need to rely on test failures to know that you need to update a fixture.
- improve the folder structure for fixtures

#### Old folder structure

Screenshot_select-area_20250916111808

#### New folder structure

Screenshot_select-area_20250916111351

#### Type checkable automatic mocks

Example:

```tsx
import type { GetApiV1BetaGroupsResponse } from '@api/types.gen'

export default {
  groups: [
    { name: 'default', registered_clients: ['client-a'] },
    { name: 'Research team', registered_clients: ['client-b'] },
    { name: 'Archive', registered_clients: [] },
  ],
} satisfies GetApiV1BetaGroupsResponse
```


renderer/src/common/mocks/fixtures/groups_name/delete.ts
```diff
@@ -0,0 +1,3 @@
+import type { DeleteApiV1BetaGroupsByNameResponse } from '@api/types.gen'
+
+export default '' satisfies DeleteApiV1BetaGroupsByNameResponse
```
peppescg, line 3: the empty string is weird 🤔

renderer/src/common/mocks/mockTemplate.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// Small utility to render the TypeScript module for generated mock fixtures.
+// It conditionally adds a type import and `satisfies` clause when a response
+// type name is provided.
+
+export function buildMockModule(
+  payload: any,
```
peppescg: I would avoid any, it would be replace with unknown?

renderer/src/common/mocks/fixtures/groups_name/delete.ts
kantord: I believe that is actually what the real API responds with.

renderer/src/common/mocks/mockTemplate.ts
kantord: :eyes:

kantord: fixed. also converted the comment above into a docstring

renderer/src/common/mocks/fixtures/groups_name/delete.ts
kantord: let me check in real life

kantord: yup, this is 100% the actual response from BE. an empty string

kantord: Screenshot_select-area_20250916131016

PR #874, opened 2025-09-15
feat(playground): persist thread on storage and adjust tools

samuv:
This is the first implementation for handling persistent threads, we will save the playground chat in the file system, it's a little bit verbose on IPC side but I will clean up in the next PR

I also adjusted the connection of the MCP tools for the remote ones

https://github.com/user-attachments/assets/c3c2e99d-b777-4e71-a0b7-f81033f27154

### DIFF
--- a/main/src/chat/index.ts
+++ b/main/src/chat/index.ts
@@ -1,7 +1,9 @@
 // Export all chat-related functionality
 export * from './types'
 export * from './providers'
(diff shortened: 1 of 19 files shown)


peppescg (approved):
Tested the remote mcp with OAuth and OIDC and works great. Good Job, great refactor, clean

PR #884, opened 2025-09-16
feat: make the custom mcp-form group-aware

kantord:
https://github.com/user-attachments/assets/87664b2b-594c-48cf-8bae-93167686b293

note: does not have or need a feature flag. when the user does not have feature flags enabled, simply all mcps will go to the default group


renderer/src/features/mcp-servers/sub-pages/customize-tools/page.tsx
```diff
@@ -75,7 +75,11 @@ export function CustomizeToolsPage() {
         { data: formData },
         {
           onSuccess: () => {
-            checkServerStatus({ serverName, isEditing: true })
+            checkServerStatus({
+              serverName,
+              groupName: workload?.group || 'default',
```
JAORMX, line 80: nit: should we have a constant for `default`?

peppescg:
Not related to your PR, but I think we should have the delete group button also if we don't have mcp

Screenshot 2025-09-16 at 16 09 26

Screenshot 2025-09-16 at 16 10 53

peppescg (approved):

renderer/src/features/mcp-servers/sub-pages/customize-tools/page.tsx
kantord: yeah but for the most part, actually the ideal solution here would be to not have a fallback at all :thinking:  however, for some reason, the API marks this as optional. :thinking: 

my understanding is that there is no logical reason for this to be optional :thinking:   in real life, this fallback should not really happen either. best would be to address it in the API I guess ahahahah

PR #892, opened 2025-09-17
chore: improve thv and secrets logs

peppescg:


main/src/toolhive-manager.ts
```diff
@@ -128,7 +128,7 @@ export async function startToolhive(tray?: Tray): Promise<void> {
       `--port=${toolhivePort}`,
     ],
     {
-      stdio: 'ignore',
+      stdio: ['ignore', 'pipe', 'pipe'],
```
samuv: I think we can use only `pipe`

main/src/toolhive-manager.ts
```diff
@@ -139,6 +139,28 @@ export async function startToolhive(tray?: Tray): Promise<void> {
...
+  // Capture and log stdout
+  if (toolhiveProcess.stdout) {
+    log.info(`[ToolHive] Capturing stdout enabled`)
+    toolhiveProcess.stdout.on('data', (data) => {
+      const output = data.toString().trim()
+      if (output) {
+        log.info(`[ToolHive stdout] ${output}`)
+      }
+    })
+  }
```
samuv: IDK how much value we will getting from this, maybe is better only `stderr`

renderer/src/routes/__root.tsx
```diff
@@ -34,7 +34,10 @@ async function setupSecretProvider(queryClient: QueryClient) {
         await createEncryptedProvider()
       }
     })
-    .catch(createEncryptedProvider)
+    .catch(() => {
+      log.info('Error setting up secret provider, creating encrypted provider')
```
samuv: 💣 maybe we could also print the error message, the API might return some extra info we can use for debugging

peppescg: right I forgot to add it

main/src/toolhive-manager.ts
peppescg: the first pipe is for stdout the last one is for the error, if we want to catch only errors we can remove the second one

main/src/toolhive-manager.ts
peppescg: let's start with error in case we will add the other logs later

main/src/toolhive-manager.ts
samuv: right, I was thinking of just putting 'pipe' as a string, but we can avoid having that info in the object, makes sense

PR #893, opened 2025-09-17
feat: allow installing registry mcps to any group

kantord:
https://github.com/user-attachments/assets/daf3b5ed-db7d-4e72-9f58-7a8bd856c3eb

behavior when feature flag is disabled: 

Screenshot_select-area_20250917134223


renderer/src/features/registry-servers/components/form-run-from-registry/index.tsx
```diff
@@ -136,10 +138,14 @@ export function FormRunFromRegistry({
       {
         server,
         data,
+        groupName: (data.group ?? 'default') as string,
```
samuv: why the type assertion is necessary?

kantord: you are right, it's not needed. removed it

samuv:
I think you need a rebase?

kantord:
> I think you need a rebase?

done

PR #897, opened 2025-09-18
refactor: reimplement useConfirm using usePrompt

kantord:
it is possible to reimplement useConfirm using usePrompt, since usePrompt is basically just a more generalized version of useConfirm. This way, we can further reduce the complexity and reuse more code, plus remove a context provider.


samuv (approved):
Nice!

PR #898, opened 2025-09-18
feat(toolhive-manager): add sentry on thv process exit

samuv:
## Changes
- **Wrap ToolHive startup in Sentry scope** for better crash context and debugging
- **Add process exit crash reporting** with conditional logic to avoid noise:
  - Report crashes when ToolHive exits unexpectedly (`fatal` severity)
  - Skip reporting during normal app shutdown or restart operations
- **Enhanced debugging context** with breadcrumbs for process lifecycle and stderr output

### Implementation Details
- Uses `Sentry.withScope()` to capture startup context and process metadata
- Captures stderr output as breadcrumbs for better crash diagnosis
- Conditional reporting based on `isRestarting` flag and app quitting state

### DIFF
--- a/main/src/tests/toolhive-manager.test.ts
+++ b/main/src/tests/toolhive-manager.test.ts
@@ -30,6 +30,12 @@ vi.mock('../system-tray')
 vi.mock('../logger')
 vi.mock('@sentry/electron/main', () => ({
   captureMessage: vi.fn(),
+  withScope: vi.fn((callback) => {
+    const mockScope = {
+      addBreadcrumb: vi.fn(),
+    }
+    callback(mockScope)
+  }),
 }))
 
 const mockSpawn = vi.mocked(spawn)
@@ -201,7 +207,7 @@ describe('toolhive-manager', () => {
       )
       expect(mockCaptureMessage).toHaveBeenCalledWith(
         `Failed to start ToolHive: ${JSON.stringify(testError)}`,
-        'error'
+        'fatal'
       )
       expect(mockUpdateTrayStatus).toHaveBeenCalledWith(mockTray, false)
     })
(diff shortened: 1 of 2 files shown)


main/src/tests/toolhive-manager.test.ts
```diff
@@ -221,6 +233,74 @@ describe('toolhive-manager', () => {
       expect(isToolhiveRunning()).toBe(false)
     })
 
+    it('captures Sentry message when process exits unexpectedly', async () => {
+      mockGetQuittingState.mockReturnValue(false)
+
+      const startPromise = startToolhive(mockTray)
+
+      await vi.advanceTimersByTimeAsync(50)
```
kantord, line 242:
```suggestion
      // Advancing the timer actually allows the promise to resolve
      await vi.advanceTimersByTimeAsync(50)
```

main/src/tests/toolhive-manager.test.ts
```diff
@@ -221,6 +233,74 @@ describe('toolhive-manager', () => {
...
+      await startPromise
+
+      // Clear previous calls
```
kantord:
```suggestion
```

main/src/toolhive-manager.ts
```diff
@@ -104,65 +105,81 @@ async function findFreePort(
...
-  )
 
-  log.info(`[startToolhive] Process spawned with PID: ${toolhiveProcess.pid}`)
+    // Capture and log stderr
+    if (toolhiveProcess.stderr) {
+      log.info(`[ToolHive] Capturing stderr enabled`)
+      toolhiveProcess.stderr.on('data', (data) => {
+        const output = data.toString().trim()
+        if (output) {
+          scope.addBreadcrumb({
```
peppescg: I would add the `log.error(`[ToolHive stderr] ${output}`)` as well in order to have the correct flow, and also everything visible and not scrubbed

samuv: added back filtering for `A new version of ToolHive is available`, since it’s not really valuable even as breadcrumbs

Issue #899, opened 2025-09-18
Remote MCP server changes


eleftherias:
- nit: "OAuth2" should be "OAuth 2.0"
- Callback port should always be optional
- For client secret, it always gives me the error `failed to resolve OAuth client secret: secret not found: oauth_client_secret`, even if I give it the secret `secret123`. The error goes away if I create a secret with the name `oauth_client_secret`, but I assume it should be able to use any secret and this is a bug.
- Client secret should be optional for both OIDC and OAuth 2.0
- PKCE should be available for both OIDC and OAuth 2.0
- Should PKCE be a checkbox instead of a true/false dropdown?

peppescg:
fixed https://github.com/stacklok/toolhive-studio/pull/900

eleftherias:
@peppescg These items weren't updated with your PR

- nit: "OAuth2" should be "OAuth 2.0"
- Callback port should always be optional
- Client secret should be optional for both OIDC and OAuth 2.0
- PKCE should be available for both OIDC and OAuth 2.0
- Should PKCE be a checkbox instead of a true/false dropdown?

peppescg:
> [@peppescg](https://github.com/peppescg) These items weren't updated with your PR
> 
> * nit: "OAuth2" should be "OAuth 2.0"
> * Callback port should always be optional
> * Client secret should be optional for both OIDC and OAuth 2.0
> * PKCE should be available for both OIDC and OAuth 2.0
> * Should PKCE be a checkbox instead of a true/false dropdown?

Thanks Ria!

>Callback port should always be optional

the callback port is mandatory for client dynamic registration

> Client secret should be optional for both OIDC and OAuth 2.0

I got authentication issues if I don't pass secrets, expliciting that it is missing

> PKCE should be available for both OIDC and OAuth 2.0

for oauth is by default true, it should be possible to set false and oauth should works the same?

>Should PKCE be a checkbox instead of a true/false dropdown?

that makes sense, agreed

peppescg:
this is if I don't pass the callback_port for dynamic client registration auth

<img width="799" height="327" alt="Image" src="https://github.com/user-attachments/assets/3efa0696-2f99-44cd-be92-9d281fc52fc2" />

eleftherias:
> the callback port is mandatory for client dynamic registration

Ok thanks, I hadn't tried the dynamic client registration. For OIDC and OAuth it should be optional.

> I got authentication issues if I don't pass secrets, expliciting that it is missing

[...]

PR #900, opened 2025-09-18
feat: remote mcp on registry

peppescg:


peppescg:
Fixes https://github.com/stacklok/toolhive-studio/issues/899

PR #909, opened 2025-09-19
fix: remote mcp doc link and validation

peppescg:
A couple of fix:
- "OAuth2" should be "OAuth 2.0"
- Callback port should be optional for oauth/oidc ( it depends from mcp server, it is needed to github one but not for some Google client)
- Client secret should be optional for both OIDC and OAuth 2.0 ( the same above it depends from mcp server )
- PKCE should be available for both OIDC and OAuth 2.0
- replace PKCE from select to checkbox

Screenshot 2025-09-19 at 16 33 21


samuv (approved):
code looks good! 👍🏼

PR #924, opened 2025-09-30
feat(playground): add traces on streaming and user actions

samuv:
as per title, with this PR we’re adding more visibility into the playground by adding traces. This will help us better understand usage patterns and diagnose issues with this new feature.

### DIFF
--- a/main/src/chat/streaming.ts
+++ b/main/src/chat/streaming.ts
@@ -5,12 +5,13 @@ import {
   convertToModelMessages,
   createIdGenerator,
 } from 'ai'
+import * as Sentry from '@sentry/electron/main'
+import type { LanguageModelV2Usage } from '@ai-sdk/provider'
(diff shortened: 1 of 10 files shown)


main/src/chat/streaming.ts
```diff
@@ -21,210 +22,332 @@ export async function handleChatStreamRealtime(
...
+        // Get MCP tools if enabled
+        const {
+          tools: mcpTools,
+          clients: mcpClients,
+          enabledTools,
+        } = await createMcpTools()
+
+        try {
+          const result = streamText({
+            model,
```
peppescg, line 57: nit: I would probably split the callback part in a separate functions for readability

samuv: yeah I think you're right, I'm a already planning a refactoring for this part

PR #929, opened 2025-09-30
feat(playground): add tooltip on attachment and remove accept

samuv:
I wrongly assume that `accept` had the logic for specify different data types, I remove the prop so user can upload any kind of files

I also added a tooltip with the name of file

example

Screenshot 2025-09-30 at 14 41 29

### DIFF
--- a/renderer/src/features/chat/components/chat-input-prompt.tsx
+++ b/renderer/src/features/chat/components/chat-input-prompt.tsx
@@ -1,4 +1,5 @@
 import { useState, useEffect, useRef } from 'react'
+import type { ChatStatus, FileUIPart } from 'ai'
 import {
   PromptInput,
   PromptInputActionAddAttachments,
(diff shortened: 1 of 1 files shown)


renderer/src/features/chat/components/chat-input-prompt.tsx
```diff
@@ -165,7 +177,6 @@ export function ChatInputPrompt({
 
   return (
     <PromptInput
-      accept="image/*,.pdf,.txt,.md,.json,.xml,.yaml,.yml,.csv,.js,.ts,.jsx,.tsx,.py,.java,.cpp,.c,.h,.html,.css,.scss,.less,.sql,.sh,.bat,.docx,.doc,.rtf"
       onError={console.error}
```
peppescg, line 180: Should we show error message in case of uploaded fail?

samuv: I don't have a generic error right now, I have it only for the streaming, I would need a little more time for that

PR #944, opened 2025-10-01
feat: add instace_id to sentry tracking

peppescg:
In order to track the instance_id, we can pass it to sentry traces using `custom.user_id` attr

### DIFF
--- a/main/src/main.ts
+++ b/main/src/main.ts
@@ -46,7 +46,7 @@ import {
   getToolhiveMcpPort,
 } from './toolhive-manager'
 import log from './logger'
-import { getAppVersion, isOfficialReleaseBuild } from './util'
+import { getAppVersion, getInstanceId, isOfficialReleaseBuild } from './util'
 import { delay } from '../../utils/delay'
 import {
   initAutoUpdate,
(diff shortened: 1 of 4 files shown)


kantord (approved):

main/src/main.ts
```diff
@@ -116,7 +116,32 @@ const store = new Store<{
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
   tracesSampleRate: 1.0,
+  // It will send errors, exceptions and captured messages to Sentry only if the user has enabled telemetry
   beforeSend: (event) => (store.get('isTelemetryEnabled', true) ? event : null),
+  // It will send transactions to Sentry only if the user has enabled telemetry
+  beforeSendTransaction: async (transaction) => {
```
samuv, line 122: I think transactions are tied to the error/event, or maybe I misunderstood. But from what I get from the documentation for our use case, since we’re using spans, we should hook into `beforeSendSpan` if we want this as an attribute of the span.

[ref](https://docs.sentry.io/platforms/javascript/tracing/span-metrics/#adding-metrics-to-all-spans)

peppescg: I tested manually `beforeSendTransaction` and it works , `beforeSend` works only for error and caputreMessage

peppescg: and we are using transaction specifically. The diffe between those are that `beforeSendSpan` cannot be drop but only modified.

nit: I discovered now that we were always sending tracing, so I covered the tracing with `isTelemetryEnabled` flag as well, cause `beforeSend` is only for error, exception and captureMessage

Screenshot 2025-10-01 at 15 57 10

samuv: ohhh transaction parameters are passed down to the span so we can have this in both places, even better then

samuv (approved):

PR #960, opened 2025-10-03
fix(tray): avoid passing tray ref and fix update on exit during restarting

samuv:
Refactor tray status management to avoid passing tray references directly, centralizing status updates through updateTrayStatus() which now automatically accesses the tray instance, and fix test assertions to match the new function signatures.

Previously, the old process exit callback would incorrectly set tray status to false during restart, even after the new process was already running successfully

### DIFF
--- a/main/src/tests/auto-update.test.ts
+++ b/main/src/tests/auto-update.test.ts
@@ -237,7 +237,7 @@ describe('auto-update', () => {
         expect(vi.mocked(dialog.showMessageBox)).toHaveBeenCalled()
         expect(vi.mocked(stopAllServers)).toHaveBeenCalled()
         expect(vi.mocked(stopToolhive)).toHaveBeenCalled()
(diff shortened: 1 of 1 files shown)


main/src/toolhive-manager.ts
```diff
@@ -169,14 +163,14 @@ export async function startToolhive(tray?: Tray): Promise<void> {
...
-      if (tray) updateTrayStatus(tray, false)
+      updateTrayStatus(false)
     })
 
     toolhiveProcess.on('exit', (code) => {
       log.warn(`ToolHive process exited with code: ${code}`)
       toolhiveProcess = undefined
-      if (tray) updateTrayStatus(tray, false)
       if (!isRestarting && !getQuittingState()) {
+        updateTrayStatus(false)
```
samuv, line 173: the fix is pretty much this, set the status to false only if is not restarting or quitting

peppescg (approved):

PR #985, opened 2025-10-06
feat: allow specifying group for remote servers

kantord:


renderer/src/features/mcp-servers/components/remote-mcp/dialog-form-remote-mcp.tsx
```diff
@@ -132,16 +136,35 @@ export function DialogFormRemoteMcp({
...
+      : {}),
   })
 
+  useEffect(() => {
+    if (isEditing || !isOpen) return
+    form.setValue('group', groupName, {
+      shouldDirty: false,
+      shouldTouch: false,
+    })
+  }, [form, groupName, isEditing, isOpen])
```
peppescg: why it is needed 🤔 ?

renderer/src/features/mcp-servers/components/remote-mcp/dialog-form-remote-mcp.tsx
```diff
@@ -154,7 +177,7 @@ export function DialogFormRemoteMcp({
           onSuccess: () => {
             checkServerStatus({
               serverName: data.name,
-              groupName,
+              groupName: form.getValues('group') || groupName,
```
peppescg: it should be data.group 🤔 ? it is inside the form definition

renderer/src/features/mcp-servers/components/remote-mcp/dialog-form-remote-mcp.tsx
```diff
@@ -177,7 +200,10 @@ export function DialogFormRemoteMcp({
         { data },
         {
           onSuccess: () => {
-            checkServerStatus({ serverName: data.name, groupName })
+            checkServerStatus({
+              serverName: data.name,
+              groupName: form.getValues('group') || groupName,
```
peppescg: above question

renderer/src/features/registry-servers/components/dialog-form-remote-registry-mcp.tsx
```diff
@@ -122,6 +129,14 @@ export function DialogFormRemoteRegistryMcp({
       : {}),
   })
 
+  useEffect(() => {
+    if (!isOpen) return
+    form.setValue('group', 'default', {
+      shouldDirty: false,
+      shouldTouch: false,
+    })
+  }, [form, isOpen])
```
peppescg: same above useEffect question on custom form

renderer/src/features/registry-servers/lib/orchestrate-run-remote-registry-server.tsx
```diff
@@ -47,6 +47,7 @@ export function convertCreateRequestToFormData(
       name: s.name || '',
       value: { secret: s.default || '', isFromStore: false },
     })),
+    group: 'default',
```
peppescg, line 50: shouldn't be `createRequest.group` ?? 'default' 🤔

kantord: maybe I am misunderstanding something, but this just sets up a safe default value for the form that will not break anything (default group always exists)

peppescg: this is the function for edit mcp `https://github.com/stacklok/toolhive-studio/blob/8140fc88a45d7348454127309f9782971a1092ab/renderer/src/features/registry-servers/lib/orchestrate-run-remote-registry-server.tsx#L17` so I was thinking that the group should be derived from the api

kantord: hmm but it seems like this is only used when creating a new worklaod from the registry :thinking:  in that case the API does not yet have a group, the user must pick it :thinking:  I don't see a usage where this is used to edit an existing server :thinking:

peppescg: it is called here `renderer/src/features/mcp-servers/components/remote-mcp/dialog-form-remote-mcp.tsx:133` and it is considering the edit
```
const editingFormData =
    isEditing &&
    convertCreateRequestToFormData(existingServer, availableSecrets)
```

peppescg: you're right, I was confusing cause they have the same name

peppescg:
there is still a test failing due a timeout

kantord:
> there is still a test failing due a timeout

so that's suddenly a new thing :smiley_cat:  let me check

kantord:
> > there is still a test failing due a timeout
> 
> so that's suddenly a new thing 😺 let me check

i think that was just flaky because the very long work flow. I can never make it fail locally, it seems like it just depends on how fast the test can run I guess. so all I did was change the max time for this one test case

peppescg (approved):

PR #1002, opened 2025-10-07
refactor: remove unneeded env vars and secrets field from remote form

peppescg:
Env vars and secrets fields are not necessary for remote MCP, cause they are tight to local mcp.


renderer/src/features/mcp-servers/lib/orchestrate-run-remote-server.tsx
```diff
@@ -50,10 +47,9 @@ export function prepareCreateWorkloadData(
  * Transforms form data into an update request object
  */
 export function prepareUpdateWorkloadData(
```
samuv, line 49: I think this function is not called anywhare if we remove envs and secrets

peppescg: it is used here `renderer/src/features/mcp-servers/hooks/use-update-remote-server.tsx:33`

PR #1003, opened 2025-10-07
test: add tracing for auto, manual update

peppescg:
### Summary

Added comprehensive Sentry tracing for the entire auto-update lifecycle to improve observability, error tracking, and analytics.

### Complete traced flow:

Manual update span.op

```
update.manual
├─ update.init 
  ├─ update.checking
  ├─ update.available
  ├─ update.downloaded
  ├─ update.user_decision
  └─ update.install 
```
Auto update span.op
```
update.init 
  ├─ update.checking
  ├─ update.available
  ├─ update.downloaded
  ├─ update.user_decision
  └─ update.install 
```
All phases are traced in Sentry with proper span hierarchy

### DIFF
--- a/main/src/tests/auto-update.test.ts
+++ b/main/src/tests/auto-update.test.ts
@@ -11,7 +11,6 @@ import {
   setAutoUpdateEnabled,
   manualUpdate,
   getLatestAvailableVersion,
-  type ReleaseInfo,
 } from '../auto-update'
 
 vi.mock('electron', () => {
@@ -988,7 +987,7 @@ describe('auto-update', () => {
         const originalPlatform = process.platform
         Object.defineProperty(process, 'platform', { value: 'darwin' })
 
-        const mockResponse: ReleaseInfo = {
+        const mockResponse = {
           tag: 'v1.5.0',
           prerelease: false,
           published_at: '2024-01-01',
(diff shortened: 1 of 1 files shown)


main/src/auto-update.ts
```diff
@@ -399,65 +730,139 @@ function isCurrentVersionPrerelease(
...
+            error instanceof Error
+              ? error.message
+              : `Unknown error during manual update ${JSON.stringify(error)}`,
+        })
+        throw error
+      } finally {
+        finish()
+      }
+    }
+  )
```
kantord, line 760: i guess this file is starting to grow  a bit too big, it's already automatically hidden by github 😸

peppescg: right, after added the sentry logic for sure...the issue is that the entire update (auto/manual) flow is in this page. I refactored in this PR all the event handlers related to autoUpdater in separated fn for better readability, but I would handle all the update logic in this page. Let me see if I can move out into a util fn some helpers fn

kantord: also I guess it seems like a lot of these attributes are common in multiple places?

```
attributes: {
        update_flow: 'true',
        'analytics.source': 'tracking',
        'analytics.type': 'event',
        current_version: getAppVersion(),
        update_state: updateState,
        toolhive_running: `${isToolhiveRunning()}`,
      },
    },
```

I guess just having a function like `getCommonSpanAttributes()` or sth like that could reduce the code a bit (probably not enough to completely solve the problem)

peppescg: I moved to a util fn the fetch assets releases logic

kantord (approved):

PR #1008, opened 2025-10-08
feat: resolve conflicts when copying mcp to group

kantord:


renderer/src/features/mcp-servers/components/card-mcp-server/server-actions/items/add-server-to-group-menu-item.tsx
```diff
@@ -46,11 +96,91 @@ export function AddServerToGroupMenuItem({
...
+      return // User cancelled
+    }
+
+    const groupName = groupResult.value
+
+    let lastRejectedName: string | null = null
+    let customName = `${serverName}-${groupName}`
+    let toastId: string | number | undefined
+
+    while (true) {
```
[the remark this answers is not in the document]
kantord: fair point, pushed a fix

renderer/src/features/mcp-servers/components/card-mcp-server/server-actions/items/add-server-to-group-menu-item.tsx
```diff
@@ -46,11 +96,91 @@ export function AddServerToGroupMenuItem({
...
+            toast.dismiss(toastId)
+          }
+          lastRejectedName = customName
+          continue
+        }
+
+        if (toastId) {
+          toast.dismiss(toastId)
+        }
+        toast.error(errorMessage || 'Failed to copy server to group')
```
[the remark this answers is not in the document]
kantord: fair point, fixed

renderer/src/features/mcp-servers/components/card-mcp-server/server-actions/items/add-server-to-group-menu-item.tsx
```diff
@@ -46,11 +96,98 @@ export function AddServerToGroupMenuItem({
...
+
+        if (toastId) {
+          toast.dismiss(toastId)
+        }
+        toast.error(
+          errorMessage ||
+            'An unexpected error occurred while copying the server to the group'
+        )
+        return false
+      }
```
samuv: I think this logic would be better placed in a dedicated hook rather than directly in the component, what do you think?

kantord: yeah

kantord: I was hesitating a bit, because I get a certain "sixth sense" feeling that we almost kinda need a bigger abstraction for this that could be reused across features. Basically, something like 

```tsx
const { mutate: saveFooBar } = useMutationFlow({
  mutation,
  inputForm: etcetera1,
  confirmation: { /* etc. */ },
```

but you are right, for now it's a better idea to just move this logic to its own dedicated hook, and think about a bigger abstraction later.

kantord: moved to a dedicated hook like you suggested, plus, reused useTransition to simplify the hook (remove manual transition tracking). I am still thinking about further simplification for this kind of workflow, but for this PR it should do it. :+1:

Issue #923, opened 2025-09-30
[Task] verify logging/tracking for the groups feature


kantord:
### Priority

Medium

### Task Type

Feature

PR #1018, opened 2025-10-09
chore: implement group related events

kantord:
## New Events:
  1. Group created - Tracks group creation with existing count
  2. Group deleted - Tracks deletion with remaining count
  3. Group navigated - Tracks transitions between default/custom groups
  4. Server copied to group - Tracks successful copy with retry metrics
  5. Server copy cancelled - Tracks abandonment at different stages
  6. Manage clients opened - Tracks dialog open with client counts
  7. Group clients updated - Tracks bulk client changes with change breakdown

  ## Enhanced Events:
  - All workload started events now include isDefaultGroup
  - All client register/unregister events now include isDefaultGroup

fixes https://github.com/stacklok/toolhive-studio/issues/923

### DIFF
--- a/renderer/src/features/clients/components/manage-clients-button.tsx
+++ b/renderer/src/features/clients/components/manage-clients-button.tsx
@@ -48,7 +48,7 @@ export function ManageClientsButton({
       Object.values(defaultValues).filter(Boolean).length
 
     trackEvent('Manage clients opened', {
-      is_default_group: groupName === 'default',
+      is_default_group: String(groupName === 'default'),
       installed_clients_count: installedClients.length,
       registered_clients_count: registeredClientsCount,
     })
--- a/renderer/src/features/mcp-servers/components/groups-manager/group-list.tsx
+++ b/renderer/src/features/mcp-servers/components/groups-manager/group-list.tsx
@@ -10,8 +10,8 @@ interface GroupListProps {
 export function GroupList({ apiGroups, currentGroupName }: GroupListProps) {
   const handleGroupClick = (toGroupName: string) => {
     trackEvent('Group navigated', {
-      from_is_default_group: currentGroupName === 'default',
-      to_is_default_group: toGroupName === 'default',
+      from_is_default_group: String(currentGroupName === 'default'),
+      to_is_default_group: String(toGroupName === 'default'),
     })
   }
 
--- a/renderer/src/features/mcp-servers/hooks/use-mutation-delete-group.ts
+++ b/renderer/src/features/mcp-servers/hooks/use-mutation-delete-group.ts
@@ -15,7 +15,7 @@ export function useMutationDeleteGroup() {
     onSuccess: () => {
       const remainingGroupsCount = (groupsData?.groups?.length ?? 1) - 1
       trackEvent('Group deleted', {
-        is_default_group: false,
+        is_default_group: 'false',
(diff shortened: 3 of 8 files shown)


renderer/src/features/clients/components/manage-clients-button.tsx
```diff
@@ -43,6 +44,15 @@ export function ManageClientsButton({
   })
 
   const handleManageClients = async () => {
+    const registeredClientsCount =
+      Object.values(defaultValues).filter(Boolean).length
+
+    trackEvent('Manage clients opened', {
+      isDefaultGroup: groupName === 'default',
+      installedClientsCount: installedClients.length,
+      registeredClientsCount,
```
samuv: I think we need snake case here too

renderer/src/features/mcp-servers/components/card-mcp-server/server-actions/items/add-server-to-group-menu-item.tsx
```diff
@@ -46,6 +47,9 @@ export function AddServerToGroupMenuItem({
     )
 
     if (!groupResult) {
+      trackEvent('Server copy cancelled', {
+        cancelledAt: 'group_selection',
```
samuv: here

renderer/src/features/mcp-servers/components/groups-manager/group-list.tsx
```diff
@@ -1,12 +1,20 @@
...
 interface GroupListProps {
   apiGroups: Array<{ name?: string; registered_clients?: string[] }>
   currentGroupName: string
 }
 
 export function GroupList({ apiGroups, currentGroupName }: GroupListProps) {
+  const handleGroupClick = (toGroupName: string) => {
+    trackEvent('Group navigated', {
+      fromIsDefaultGroup: currentGroupName === 'default',
+      toIsDefaultGroup: toGroupName === 'default',
```
samuv: here

renderer/src/features/mcp-servers/hooks/use-mutation-delete-group.ts
```diff
@@ -2,14 +2,22 @@ import { useQueryClient } from '@tanstack/react-query'
...
   const queryClient = useQueryClient()
   const navigate = useNavigate()
+  const { data: groupsData } = useGroups()
 
   return useToastMutation({
     ...deleteApiV1BetaGroupsByNameMutation(),
     onSuccess: () => {
+      const remainingGroupsCount = (groupsData?.groups?.length ?? 1) - 1
+      trackEvent('Group deleted', {
+        is_default_group: false,
```
peppescg: all the booleans in Sentry are parsed correctly (Sentry use 1|0 as bool), you need to cast to string "false" in order to let Sentry convert as bool, in this way it will be easy to filter attr by `key is true` rather than `key is 1`

kantord: so basically  you prefer to use string rather than number there?

peppescg: it is necessary for helping filter attr, you can check the sentry dashboard in order to understand

kantord: thanks for the feedback! pushed the fix

PR #1019, opened 2025-10-09
feat(editing): optimistic update, centralize modal logic and merge similiar hooks

samuv:
This PR refactors the server editing functionality with the following improvements:

### Centralized Dialog Management
- Introduced a context-based solution to render a single edit modal instance at the page level
- Previously, each server card rendered its own modal instance that would unmount when the dropdown closed
- This reduces DOM overhead and improves performance

### Merged Duplicate Hooks
- Consolidated `useUpdateServer` and `useUpdateRemoteServer` into a single unified hook
- Used type guards and conditional typing to handle both remote and local server updates
- Centralizes the logic in one place, making it easier to maintain

### Optimistic Updates
- Added optimistic UI updates when triggering server edits
- The card now shows an "updating" status immediately, providing better user feedback

https://github.com/user-attachments/assets/cc25ab66-3732-4a46-9c9b-bef263c5e72a

### DIFF
--- a/renderer/src/common/mocks/fixtures/workloads_name_edit/post.ts
+++ b/renderer/src/common/mocks/fixtures/workloads_name_edit/post.ts
@@ -2,5 +2,5 @@ import type { PostApiV1BetaWorkloadsByNameEditResponse } from '@api/types.gen'
 
 export default {
   name: 'commodo',
-  port: -46951877,
+  port: 51877,
 } satisfies PostApiV1BetaWorkloadsByNameEditResponse


renderer/src/features/mcp-servers/components/__tests__/card-mcp-server.test.tsx
```diff
@@ -27,14 +28,16 @@ function createCardMcpServerTestRouter() {
...
     component: () => (
-      <CardMcpServer
-        name="postgres-db"
-        status="running"
-        statusContext={undefined}
-        url="http://localhost:8080"
-        transport="http"
-        group="default"
-      />
+      <EditServerDialogProvider>
```
kantord, line 31: kinda a random idea: we could actually use `usePrompt` for this entire thing, without the need for an addition provider 🤔 

I am not saying that you have to use it, and I'm not even sure that I would use it myself. Is it a good idea to use that hook for more complex scenarios, or just for simple cases? I originally designed it for simple cases, and have seen this same design used in simple cases, but there is no **technical** reason why we could not use it for the edit dialog.

Just putting the idea/possibility out there, not sure what you think about it

samuv: Did you mean for the entire form? IDK, I’d need to take another close look. Right now, the saved state is only for the modal, we just store a few things needed to render the modal and form

kantord: yup, basically it can take and display any RHF form component and handle the logic of showing a modal and the cancel/save buttons. but definitely does not need to be a change in this PR, just planting the idea 🌱

renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
```diff
@@ -284,4 +295,93 @@ describe('DialogFormRemoteMcp', () => {
...
+    })
+
+    // Wait for form to load with existing data
+    await waitFor(() => {
+      expect(screen.getByRole('textbox', { name: /url/i })).toHaveValue(
+        'https://old-api.example.com'
+      )
+    })
+
+    // Update the URL
```
kantord: nit pick, but this kind of comment is kinda redundant, because the code itself is already easy to read

samuv: yeah let's remove those

renderer/src/common/mocks/fixtures/workloads_name_edit/post.ts
```diff
@@ -0,0 +1,6 @@
+import type { PostApiV1BetaWorkloadsByNameEditResponse } from '@api/types.gen'
+
+export default {
+  name: 'commodo',
+  port: -46951877,
```
kantord:
```suggestion
  port: 51877,
```

does not really change anything, just making the mock data a little bit more realistic 😸

renderer/src/features/mcp-servers/hooks/use-mutation-update-workload.ts
```diff
@@ -0,0 +1,44 @@
...
+          server.name === path.name ? { ...server, status: 'updating' } : server
+        )
+        return {
+          ...old,
+          workloads: newWorkloads,
+        }
+      })
+      return { queryKey, previousServersList }
+    },
+    onError: (_error, _variables, context) => {
```
kantord, line 30: don't we need to notify the user about the error here, or re-throw so that it can be caught somewhere else? 🤔

samuv: it is handle in `updateServerMutation` with a dedicated callback, this part is only for the optimistic update

renderer/src/common/mocks/fixtures/workloads_name_edit/post.ts
samuv: ops, edited

renderer/src/features/mcp-servers/components/__tests__/card-mcp-server.test.tsx
samuv: yeah let's talk about this in a different effort

PR #1037, opened 2025-10-13
fix: cannot find image reference when copying remote server

kantord:
fixes the bug that was mentioned here: https://stacklok.slack.com/archives/C091GRYAG49/p1760022742218379?thread_ts=1760003349.102839&cid=C091GRYAG49

### DIFF
--- a/renderer/src/common/mocks/customHandlers/index.ts
+++ b/renderer/src/common/mocks/customHandlers/index.ts
@@ -66,32 +66,6 @@ export const customHandlers = [
     return HttpResponse.json({ workloads: filtered })
   }),
 
-  http.get(mswEndpoint('/api/v1beta/workloads/:name'), ({ params }) => {
-    const { name } = params
-
-    const server = getWorkloadByName(name as string)
-    if (!server) {
-      return HttpResponse.json({ error: 'Server not found' }, { status: 404 })
-    }
-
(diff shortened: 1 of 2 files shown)


renderer/src/features/mcp-servers/hooks/use-copy-server-to-group.ts
```diff
@@ -102,15 +102,48 @@ export function useCopyServerToGroup(serverName: string) {
...
+                              }
+                            : undefined,
+                          issuer: runConfig.remote_auth_config.issuer,
+                          oauth_params:
+                            runConfig.remote_auth_config.oauthParams,
+                          scopes: runConfig.remote_auth_config.scopes,
+                          skip_browser:
+                            runConfig.remote_auth_config.skipBrowser || false,
+                          token_url: runConfig.remote_auth_config.tokenURL,
+                          use_pkce: true,
```
peppescg: 🤔 this should be retrieved from runConfig as well, and having default `true` as default value

kantord: seems like RunnerRemoteAuthConfig does not have this field.

of course, we can just skip the field instead, but that would also lead to some weird behavior potentially?

peppescg (approved):

PR #1038, opened 2025-10-13
feat: list groups in registry

kantord:
Notes about testing et cetera:

- You will need a custom registry, because the default registry will not have groups
- An example of a custom registry was included in the repo to simplify this
- This is behind a feature flag. use `FeatureFlag.groups_in_registry.enable()`
- This is **NOT** part of the groups feature and the first groups release

screenshot-2025-10-14_16-49-58

What's included:
- All required logic to display groups on the registry page. Installing them is not yet possible, hence the need for the feature flag

How it looks when the feature flag is enabled AND using a custom registry with groups included:

https://github.com/user-attachments/assets/9196a283-7b69-40b9-b4cf-e813df618d36

when the feature flag is disabled, the groups are simply hidden. (Screenshot showing the same custom registry)

screenshot-2025-10-14_16-59-09

### DIFF
--- a/renderer/src/features/registry-servers/components/grid-cards-registry.tsx
+++ b/renderer/src/features/registry-servers/components/grid-cards-registry.tsx
@@ -65,7 +65,7 @@ export function GridCardsRegistry({
       {filteredAndSortedItems.length === 0 && (
         <div className="text-muted-foreground py-12 text-center">
           <p className="text-sm">
-            No registry items found matching the current filter
+            No servers or groups found matching the current filter
           </p>
         </div>
       )}


renderer/src/features/registry-servers/components/grid-cards-registry.tsx
```diff
@@ -35,28 +44,28 @@ export function GridCardsRegistryServer({
...
             }}
           />
         ))}
       </div>
-      {filteredAndSortedServers.length === 0 && (
+      {filteredAndSortedItems.length === 0 && (
         <div className="text-muted-foreground py-12 text-center">
           <p className="text-sm">
-            No registry servers found matching the current filter
+            No registry items found matching the current filter
```
peppescg: I would maintain `servers` here, cause at the end of the day a group is a list of servers, or saying `servers or groups`

kantord: So basically sth like

```suggestion
            No servers or groups found matching the current filter
```

peppescg: yep makes sense

peppescg: I was thinking that maybe the term groups without having groups available could be hide, but I don't thing at this stage it is so important, up to you

PR #1040, opened 2025-10-13
fix: http server not ready when app start

peppescg:
**Loading toolhive experience, waiting for http server**
Screenshot 2025-10-13 at 19 41 35

**Generic error after health check rety failed**
Screenshot 2025-10-13 at 17 31 50

**Full flow**

https://github.com/user-attachments/assets/6c342468-cdbd-4008-b28f-295a33b36ec6

### DIFF
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
-    "target": "ES2020",
+    "target": "ES2023",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
     "module": "ESNext",
     "skipLibCheck": true,
     /* Bundler mode */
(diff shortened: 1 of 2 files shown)


renderer/src/routes/__root.tsx
```diff
@@ -21,6 +21,25 @@ import '@fontsource/atkinson-hyperlegible/700-italic.css'
...
+      isToolhiveRunning: boolean
+      containerEngineAvailable: boolean
+    }
+  ) {
+    super(message)
+    this.name = 'HealthCheckError'
+    this.isToolhiveRunning = context.isToolhiveRunning
+    this.containerEngineAvailable = context.containerEngineAvailable
+  }
+}
```
samuv: Ehm, I’m not sure I’d extend Error here, I’d probably handle it a bit differently instead

peppescg: yep I don't like me either, I found another solution using ES2023 where you can configure a cause object

tsconfig.app.json
```diff
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
-    "target": "ES2020",
+    "target": "ES2023",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
peppescg, line 6: we already had the node one to ES2023, so I aligned both tsconfig using the same version.
I tested it within a make and run mcp, stop vs start, connect to client and secret provider, everything works as expected

samuv: nice!

PR #1043, opened 2025-10-14
fix: execute queries and mutations also if offline

peppescg:
https://github.com/user-attachments/assets/d979f3bb-3fa3-4c0f-8d57-23d63d83498a

### DIFF
--- a/renderer/src/common/lib/query-client.ts
+++ b/renderer/src/common/lib/query-client.ts
@@ -3,11 +3,11 @@ import { QueryClient } from '@tanstack/react-query'
 export const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      // Execute queries even when detects offline state
+      // Execute queries even when it detects offline state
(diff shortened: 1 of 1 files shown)


renderer/src/common/lib/query-client.ts
```diff
@@ -1,3 +1,14 @@
...
+  defaultOptions: {
+    queries: {
+      // Execute queries even when it detects offline state
+      networkMode: 'always',
+    },
+    mutations: {
+      // Execute mutations (POST/PUT/DELETE) even when it detects offline state
+      networkMode: 'always',
+    },
+  },
```
samuv, line 13: did we need to handle this also in the main process?

peppescg: we don't have a QueryClient configuration in main

samuv: good to know that the fetch client is not impacted

PR #1056, opened 2025-10-14
fix: container engine loop in case of mismatch with thv

peppescg:
Avoid connection refused loop in case of mismatch container engine check, showing generic error page, when the user can quit the app easily or restart the app with try again button

https://github.com/user-attachments/assets/4b37b022-5c77-4904-a0ec-099b3757ea7d

### DIFF
--- a/renderer/src/common/components/error/base-error-screen.tsx
+++ b/renderer/src/common/components/error/base-error-screen.tsx
@@ -9,6 +9,8 @@ import {
 } from '../ui/card'
 import { RefreshCw } from 'lucide-react'
 import type { ReactNode } from 'react'
+import { withMinimumDelay } from './utils'
+import log from 'electron-log/renderer'
 
 interface BaseErrorScreenProps {
   title: string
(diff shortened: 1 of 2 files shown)


renderer/src/common/components/error/__tests__/index.test.tsx
```diff
@@ -89,13 +102,19 @@ describe('Error', () => {
...
+      cause: { containerEngineAvailable: true },
+    })
 
     render(
       <QueryClientProvider client={queryClient}>
-        <ErrorComponent error={genericError} />
+        <ErrorComponent
+          error={
+            genericError as Error & {
+              cause?: { containerEngineAvailable?: boolean }
```
kantord, line 114: just one question... why cause is not an enum-like type for example?


```
type EngineErrorCause = "engine_not_available" | "whatever_other_cause"
```

I think that would also clarify in a single place what **other** causes are possible. Also it would avoid a possible future complication (like additional attributes for the `cause`)


Also I would probably make this error class also shared across component to avoid code duplication.


```
type EngineError<CauseType = EngineErrorCause> = Error & { cause: CauseType }
```

That way you would have a generic class, but whenever you need to specify that a component/function can only handle a specific case, you can still do things like `EngineError<'container_engine_not_available'>`

peppescg: That's only for tsc in a test. In the _root component we don't know which kind of errors we could have cause it comes from BE.

The logic about cause is only needed for matching a specific root use case, it could not be shared
``` 
throw new Error('Health check failed', {
        cause: {
          isToolhiveRunning,
          containerEngineAvailable: containerEngineStatus.available,
        },
      })
```

kantord: i guess I would not call that a "cause" I would maybe call it "details" or "context". 

but I guess it's nit picking

peppescg: cause comes from ES2022 it is js built-in fro customize error instance, I didn't create it 😆, but I agree that it is misleading

PR #1065, opened 2025-10-15
feat(tool-filtering): adjust the logic of filtering based on the registry

samuv:
This PR introduces tool filtering functionality, allowing users to customize which tools are enabled/disabled for their MCP servers and exposed in thv proxy.

**Current Scope:**
- Running servers installed from the registry
- Remote server/Custom servers (not yet supported)

**Key Features:**
- **Customize Tools page** with scrollable table displaying all available tools with descriptions
- **Toggle switches** for individual tool enable/disable
- **Tag drift detection** that alerts users when local image differs from registry version, with visual indicators comparing both tags
- **Dynamic height** - compact for few tools, scrollable for many, properly handles variable-length descriptions
- **Persistent preferences** with "Apply" and "Enable all tools" actions

**Technical Approach:**

We rely on the registry's tool list as the canonical source to show all available tools (including previously disabled ones), but since local Docker images may use different tags/versions than the registry, we can't guarantee the registry's tool list matches the locally installed version - so we display a tag drift alert when mismatches are detected and provide a reset option to handle this discrepancy.

This is the first implementation after deciding to use the registry's list of tools as the foundation for tool filtering.

https://github.com/user-attachments/assets/e98cf3c3-cf8e-4859-a1fe-635076405880

Updated Alert styling:

Screenshot 2025-10-16 at 12 35 19
Screenshot 2025-10-16 at 12 35 03

Added the alert when there is a mismatch of tool names from the registry:
(not a real example, just forced to see the alert)
Screenshot 2025-10-17 at 11 13 40

### DIFF
--- a/renderer/src/features/mcp-servers/components/customize-tools-table.tsx
+++ b/renderer/src/features/mcp-servers/components/customize-tools-table.tsx
@@ -22,6 +22,7 @@ import {
 } from '@/common/components/ui/alert'
 import { Badge } from '@/common/components/ui/badge'
 import { TriangleAlert } from 'lucide-react'
+import type { IsFromRegistryToolDiff } from '../hooks/use-is-server-from-registry'
 
 interface Tool {
   name: string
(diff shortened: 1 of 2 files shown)


renderer/src/features/mcp-servers/components/customize-tools-table.tsx
```diff
@@ -8,26 +8,94 @@ import {
...
 interface CustomizeToolsTableProps {
   tools: Tool[]
   isLoading?: boolean
-  onApply?: (enabledTools: Record<string, boolean>) => void
-  onCancel?: () => void
+  toolsDiff?: {
+    hasExactMatch: boolean
+    addedTools: string[]
+    missingTools: string[]
+  } | null
```
peppescg: it seems repeatable in multiple files, maybe it could be a shared type

samuv: I added a interface for it

PR #1066, opened 2025-10-15
fix: remote client_secret target value

peppescg:
The `client_secret` should point to the target provider secret key rather than the name.

**Edit gh remote mcp**

https://github.com/user-attachments/assets/23444f8c-e8b4-4caa-9951-1e476dea8b29

**Create gh remote mpc**

https://github.com/user-attachments/assets/a44365f6-a8cb-47fb-8623-9162acf52cb8

### DIFF
--- a/renderer/src/features/mcp-servers/lib/__tests__/orchestrate-run-local-server.test.tsx
+++ b/renderer/src/features/mcp-servers/lib/__tests__/orchestrate-run-local-server.test.tsx
@@ -1,4 +1,4 @@
-import { it, expect, vi, describe, beforeEach } from 'vitest'
+import { it, expect, describe } from 'vitest'
 import type { FormSchemaLocalMcp } from '../form-schema-local-mcp'
 import {
   prepareCreateWorkloadData,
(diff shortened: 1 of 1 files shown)


renderer/src/features/mcp-servers/lib/__tests__/orchestrate-run-remote-server.test.tsx
```diff
@@ -0,0 +1,285 @@
+import { it, expect, vi, describe, beforeEach } from 'vitest'
+import type { FormSchemaRemoteMcp } from '@/common/lib/workloads/remote/form-schema-remote-mcp'
+import {
+  prepareCreateWorkloadData,
+  prepareUpdateRemoteWorkloadData,
+  convertCreateRequestToFormData,
+} from '../orchestrate-run-remote-server'
+import type { V1CreateRequest, V1ListSecretsResponse } from '@api/types.gen'
+
+vi.mock('sonner', async () => {
```
kantord, line 10: we should probably not need to mock this in every single test file 🤔  but we can fix this in a follow up pr hahahahahaa

peppescg: I saw in some case we are testing the text of the notification, that IMO is too much considering considering that now we can record a mutation and check the success, so I agree with you

peppescg: @kantord we already have it in the `vitest.setup.ts` 😆  I removed it from my test

PR #1068, opened 2025-10-16
feat: release groups feature

kantord:


renderer/src/features/mcp-servers/components/local-mcp/dialog-form-local-mcp.tsx
```diff
@@ -288,7 +285,7 @@ export function DialogFormLocalMcp({
                   form={form}
                   isEditing={isEditing}
                   groupProps={{
-                    show: isGroupsEnabled,
+                    show: true,
```
kantord: i guess this is still somewhat redundant

PR #1088, opened 2025-10-17
feat: create meta-mcp route

kantord:
https://github.com/user-attachments/assets/2a1ac72b-9930-4890-9e9a-7c363ec3afed

fixes https://github.com/stacklok/toolhive-studio/issues/1100 

fixes https://github.com/stacklok/toolhive-studio/issues/1096


renderer/src/common/components/settings/tabs/general-tab.tsx
```diff
@@ -183,6 +239,7 @@ export function GeneralTab() {
         <AutoLaunchField />
         <TelemetryField />
         <QuitConfirmationField />
+        <ExperimentalFeaturesMasterToggle />
```
peppescg: I think you can remove this and leverage only the below one

PR #1090, opened 2025-10-17
chore: bump thv to 3.11

kantord:


api/generated/@tanstack/react-query.gen.ts
```diff
@@ -178,7 +178,6 @@ export const getApiOpenapiJsonQueryKey = (
 
 /**
  * Get OpenAPI specification
- *
```
samuv, line 181: I think you need to do a `pnpm i` first

kantord: you are right :+1: 

I was wondering why it happened...

peppescg (approved):

JAORMX (approved):

PR #1092, opened 2025-10-17
fix: mcp state during restart and stop

peppescg:
Fix a couple of bugs:
- polling status api on mcp restart
- invalidate mcp cache by group
- refetch workloads list per group after mcp update

https://github.com/user-attachments/assets/d9796326-8cb6-43e8-85a4-3db194e1cff7

### DIFF
--- a/renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
+++ b/renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
@@ -191,7 +191,7 @@ export function useMutationRestartServer({
           return statusResponses.map((response, index) => ({
             name: names[index],
             status: response.status || 'unknown',
-          })) as CoreWorkload[]
+          }))
         },
         [name],
         'running'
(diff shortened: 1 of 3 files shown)


renderer/src/features/mcp-servers/hooks/use-mutation-stop-server.ts
```diff
@@ -15,10 +15,16 @@ const getMutationData = (name: string) => ({
...
+  name,
+  group = 'default',
+}: {
+  name: string
+  group?: string
+}) {
   const queryClient = useQueryClient()
   const queryKey = getApiV1BetaWorkloadsQueryKey({
-    query: { all: true, group: 'default' },
+    query: { all: true, group: group },
```
samuv, line 27: this is a good catch

renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
```diff
@@ -178,10 +177,26 @@ export function useMutationRestartServer({
...
+            names.map((name) =>
+              queryClient.fetchQuery(
+                getApiV1BetaWorkloadsByNameStatusOptions({ path: { name } })
+              )
+            )
+          )
+          return statusResponses.map((response, index) => ({
+            name: names[index],
+            status: response.status || 'unknown',
+          })) as CoreWorkload[]
```
samuv: I think this is not needed anymore, can you quickly try to remove it, it should just work

peppescg: yes I copy pasted what we had from delete but we don't need a return value

samuv: Sorry, I meant only the type assertion

PR #1102, opened 2025-10-20
feat(tool-filtering): add remote server compatibility with the registry

samuv:
This add the matching logic for tool filtering with remote servers on the hook and adjusted the logic of the form when we are editing a remote server

https://github.com/user-attachments/assets/beaf5149-de2a-48a7-8786-7811cb5dc568

### DIFF
--- a/renderer/src/features/mcp-servers/sub-pages/customize-tools/__tests__/page.test.tsx
+++ b/renderer/src/features/mcp-servers/sub-pages/customize-tools/__tests__/page.test.tsx
@@ -1,6 +1,6 @@
 import { render, waitFor, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { describe, it, expect, beforeEach, vi } from 'vitest'
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { server as mswServer } from '@/common/mocks/node'
 import { http, HttpResponse } from 'msw'
 import { mswEndpoint } from '@/common/mocks/customHandlers'


renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts
```diff
@@ -35,29 +35,39 @@ export function useIsServerFromRegistry(serverName: string) {
...
-    ? serversList.find(
-        (item) =>
-          getImageNameWithoutTag(item.image) ===
-          getImageNameWithoutTag(workload?.image)
-      )
-    : undefined
+  const { servers: serversList = [], remote_servers: remoteServersList = [] } =
+    data || {}
+
+  // If workload has a URL defined, it's a remote server - match by URL
```
kantord, line 41: we need to flag this for BE. it's not the first time we are adding logic to detect remove server with this logic. IMO instead of this, we should ask BE to send sth like `"serverLocation": "remote"` or sth similar

renderer/src/features/mcp-servers/sub-pages/customize-tools/__tests__/page.test.tsx
```diff
@@ -0,0 +1,285 @@
...
+  createRootRoute,
+  createRoute,
+  Outlet,
+  Router,
+  RouterProvider,
+} from '@tanstack/react-router'
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
+import { CustomizeToolsPage } from '../page'
+
+const createLocalServerWorkload = (
```
kantord, line 22: you can probably import an auto generated/customized fixture and just add the overrides here? (It's nit pick tho)

samuv: Yeah, I’d take a look at what we have and extend the base scenario in the autogenerated code and just changing what we need

renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts
samuv: Yeah, I think we’ve already given the feedback, just waiting for the next iteration

Issue #510, opened 2025-07-14
[Bug]


jtenniswood:
### Bug Description

<img width="744" height="582" alt="Image" src="https://github.com/user-attachments/assets/da6da509-f4a6-4d91-bc77-5f5ff107109a" />

Fix alignment of the back button. This can be changed to just a link to avoid the button border issue

### Priority

Medium

michelemazzucco:
Hi 👋 

Opened a PR to fix this issue: https://github.com/stacklok/toolhive-studio/pull/1108

PR #1108, opened 2025-10-20
fix(ui): resolved back button alignment

michelemazzucco:
## Description
PR to fix this issue: https://github.com/stacklok/toolhive-studio/issues/510

### Additional Changes
I also added the `cursor-pointer` class to the button. While this is the default cursor for buttons in browsers, Electron appears to use the `default` cursor instead.

## Screenshots
CleanShot 2025-10-20 at 21 08 31@2x

## Extra / Suggestions
I tried to follow what was suggested in the issue, but there are two extra tweaks that I think would be nice to add:

### A. Color instead of underline
Instead of using an underline, I would use a color change instead. This works better when there's an icon contained in the button.

CleanShot 2025-10-20 at 21 11 50

### B. ChevronLeft vs ArrowLeftIcon
As you can see from the screenshots below, the `ChevronLeft` icon, due to its frame offset, contributes to creating a misalignment effect. The `ArrowLeftIcon` might help reduce this misalignment effect.

CleanShot 2025-10-20 at 21 20 04@2x
<br />
<br />
Please let me know what you think :)

### DIFF
--- a/renderer/src/features/mcp-servers/sub-pages/customize-tools/page.tsx
+++ b/renderer/src/features/mcp-servers/sub-pages/customize-tools/page.tsx
@@ -225,11 +225,11 @@ export function CustomizeToolsPage() {
       <div className="mb-2">
         <LinkViewTransition to={`/group/${workload?.group || 'default'}`}>
           <Button
-            variant="ghost"
+            variant="link"
             aria-label="Back"
             className="text-muted-foreground"
           >
-            <ChevronLeft className="size-5" />
+            <ChevronLeft className="size-4" />
             Back
           </Button>
         </LinkViewTransition>
--- a/renderer/src/features/mcp-servers/sub-pages/logs-page/index.tsx
+++ b/renderer/src/features/mcp-servers/sub-pages/logs-page/index.tsx
@@ -43,11 +43,11 @@ export function LogsPage() {
       <div className="mb-2">
         <LinkViewTransition {...backLink}>
           <Button
-            variant="ghost"
+            variant="link"
             aria-label="Back"
             className="text-muted-foreground"
           >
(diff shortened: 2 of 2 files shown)


samuv:
Thank you, @michelemazzucco ❤️   looks great!

Just one thing: we were thinking of going back to a regular button like this:

Registry Server Description

What do you think?

michelemazzucco:
> Thank you, @michelemazzucco ❤️ looks great!
> 
> Just one thing: we were thinking of going back to a regular button like this:
> 
> Registry Server Description
> What do you think?

I considered that option too, but I didn't commit to it because of the impact on the overall page hierarchy – i.e. it would give the back button more visibility/relevance than the one it might actually require.

But if you're happy with that approach, I can update it to use that variant instead 👀

samuv:
> > Thank you, @michelemazzucco ❤️ looks great!
> > Just one thing: we were thinking of going back to a regular button like this:
> > What do you think?
> 
> I considered that option too, but I didn't commit to it because of the impact on the overall page hierarchy – i.e. it would give the back button more visibility/relevance than the one it might actually require.
> 
> But if you're happy with that approach, I can update it to use that variant instead 👀

I need to validate this internally, I’ll circle back to you tomorrow

samuv:
@michelemazzucco let's go with your proposal, can you apply this to all the back button you find?

michelemazzucco:
> @michelemazzucco let's go with your proposal, can you apply this to all the back button you find?

I have updated the variant and icon size for all the back buttons I managed to find. For now, I kept the underline on `:hover` and did not change the arrow icons (i.e., kept the chevron).

peppescg:
thanks @michelemazzucco for the contribution 🙌 Very nice job!

PR #1113, opened 2025-10-21
fix(instance-id): use correct data path for all the OS

samuv:
As per the title, the data path for `thv` differs across operating systems compared to Electron’s `userData` folder.  
We need to handle this properly to retrieve the `updates.json` file.


peppescg (approved):
🚢

PR #1114, opened 2025-10-21
feat: init meta-optimizer group and meta-mcp workload on app bootstrap

peppescg:
## Meta Optimizer Initialization

Implemented automatic initialization of the meta-MCP server workload based on feature flags.

### Changes
- Added `initMetaOptimizer()` function that:
  - Checks both `EXPERIMENTAL_FEATURES` and `META_OPTIMIZER` feature flags
  - Auto-creates `mcp-optimizer---` group if missing
  - Auto-creates `meta-mcp` workload if missing
  - Fetches server config from registry
  - Handles errors gracefully with proper logging

### Implementation Details
- Uses `Promise.allSettled` for parallel feature flag checks
- Uses React Query's `queryClient.ensureQueryData` for efficient caching
- Treats 404 responses as valid "not found" scenarios
- Each function has granular error handling with electron-log

### Testing
Added comprehensive unit tests covering:
- Feature flag orchestration
- Group creation when missing
- Workload creation when missing
- Skip creation when resources already exist
- Error handling scenarios

https://github.com/user-attachments/assets/a511687e-8789-4561-97cd-d8af31aa897c

### DIFF
--- a/renderer/src/common/lib/meta-optimizer.ts
+++ b/renderer/src/common/lib/meta-optimizer.ts
@@ -8,6 +8,7 @@ import {
 } from '@api/sdk.gen'
 import type { V1CreateRequest } from '@api/types.gen'
 import {
+  getApiV1BetaGroupsQueryKey,
   getApiV1BetaRegistryByNameServersByServerNameOptions,
   getApiV1BetaWorkloadsByNameOptions,
 } from '@api/@tanstack/react-query.gen'
@@ -39,7 +40,7 @@ async function ensureMetaOptimizerWorkload() {
 async function createMetaOptimizerWorkload() {
   try {
     const workloadDetail = await ensureMetaOptimizerWorkload()
-    if (workloadDetail) {
+    if (workloadDetail?.group === MCP_OPTIMIZER_GROUP_NAME) {
       return workloadDetail
     }
 
@@ -103,7 +104,7 @@ async function createMetaOptimizerGroup() {
   }
 
   await queryClient.invalidateQueries({
-    queryKey: ['api', 'v1beta', 'groups'],
+    queryKey: getApiV1BetaGroupsQueryKey(),
   })
 
   // Create workload after group creation succeeds
@@ -112,8 +113,8 @@ async function createMetaOptimizerGroup() {
 
 async function ensureMetaOptimizerGroup() {
   try {
-    const rawGroups = await queryClient.ensureQueryData({
-      queryKey: ['api', 'v1beta', 'groups'],
+    const rawGroups = await queryClient.fetchQuery({
+      queryKey: getApiV1BetaGroupsQueryKey(),
       queryFn: async () => {
         const response = await getApiV1BetaGroups()
 
@@ -130,11 +131,12 @@ async function ensureMetaOptimizerGroup() {
     const metaOptimizerGrp = rawGroups?.groups?.find(
       (group) => group.name === MCP_OPTIMIZER_GROUP_NAME
     )
+
     if (!metaOptimizerGrp) {
       return await createMetaOptimizerGroup()
     }
 
-    return metaOptimizerGrp
+    return await createMetaOptimizerWorkload()
   } catch (error) {
     log.error('[ensureMetaOptimizerGroup] Error checking group:', error)
     return undefined
--- a/renderer/src/common/lib/__tests__/meta-optimizer.test.ts
+++ b/renderer/src/common/lib/__tests__/meta-optimizer.test.ts
@@ -126,7 +126,10 @@ describe('Meta Optimizer', () => {
         ),
         // Mock workload already exists
         http.get(mswEndpoint('/api/v1beta/workloads/:name'), () =>
-          HttpResponse.json({ name: META_MCP_SERVER_NAME })
+          HttpResponse.json({
+            name: META_MCP_SERVER_NAME,
+            group: MCP_OPTIMIZER_GROUP_NAME,
+          })
         )
       )
 


renderer/src/routes/__root.tsx
```diff
@@ -183,6 +184,9 @@ export const Route = createRootRouteWithContext<{
       })
     }
   },
-  loader: async ({ context: { queryClient } }) =>
-    await setupSecretProvider(queryClient),
+  loader: async ({ context: { queryClient } }) => {
+    await setupSecretProvider(queryClient)
+    // avoiding blocking the UI render we can init the meta optimizer in the background
```
kantord: i would probably add sth here about the implications. does this mean that this logic will run in the background while the app is initialized, i. e. it might not be immediately available.

I would probably add something like:

"Similarly to how the group restart logic works, this will happen in the background and might not be immediately available"

renderer/src/common/lib/meta-optimizer.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    const isExperimentalEnabled =
+      experimentalFeaturesEnabled.status === 'fulfilled' &&
+      experimentalFeaturesEnabled.value === true
+
+    const isOptimizerEnabled =
+      metaOptimizerEnabled.status === 'fulfilled' &&
+      metaOptimizerEnabled.value === true
+
+    if (isExperimentalEnabled && isOptimizerEnabled) {
```
kantord, line 167: I would probably prefer the early-exit/guard pattern here:

```tsx
if (!isExperimentalEnabled) {
  return
}

// ... etc ...

if (!isOptimizerEnabled) {
  return
}

// ... etc ...

// (no if here)
await checkMetaOptimizerGroup()
await createMetaOptimizerWorkload()
```

kantord: but it is also somewhat odd that the function is called `initMetaOptimizer` but as a matter of fact, it does not necessarily initialize it. 

I would think about just having a function like this:


```
function isMetaOptimizerEnabled() {
  if (!isExperimentalEnabled) {
    return false
   }

  // ... etc ...

  if (!isOptimizerEnabled) {
    return false
  }


  return true
}
```


then when calling it, the fact that it is not *always* initialized is immediately obvious:


```tsx

if (isMetaOptimizerEnabled()) {
    await initMetaOptimizer()
}
```


but of course, these are nitpicks, feel free to ignore it in the interest of time, or just create an issue for post-release cleanup and put them there

renderer/src/common/lib/meta-optimizer.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  await queryClient.invalidateQueries({
+    queryKey: ['api', 'v1beta', 'groups'],
+  })
+
+  return response.data
+}
+
+async function checkMetaOptimizerGroup() {
```
kantord:
```suggestion
async function ensureMetaOptimizerGroup() {
```

renderer/src/common/lib/meta-optimizer.ts
```diff
@@ -0,0 +1,171 @@
...
+    const body: V1CreateRequest = {
+      name: META_MCP_SERVER_NAME,
+      image: server.image,
+      transport: server.transport,
+      env_vars: {},
+      secrets: [],
+      cmd_arguments: [],
+      network_isolation: false,
+      volumes: [],
+      group: MCP_OPTIMIZER_GROUP_NAME,
```
kantord, line 69: I guess here you could actually do sth like
```suggestion
      group: await ensureMetaOptimizerGroup(),
```

so basically the syntax automatically makes sure that these functions are always calledc in the correct order

renderer/src/common/lib/meta-optimizer.ts
```diff
@@ -0,0 +1,171 @@
...
+      return undefined
+    }
+
+    // Log unexpected errors
+    log.error('[getMetaOptimizerWorkload] Error fetching workload:', error)
+    return undefined
+  }
+}
+
+async function createMetaOptimizerWorkload() {
```
kantord, line 40:
```suggestion
async function ensureMetaOptimizerWorkload() {
```

renderer/src/common/lib/meta-optimizer.ts
peppescg: so basically the `isExperimentalEnabled` check will be disabled after Wed, simplify the logic here, the reason is that we don't want to show the meta-mcp in experimental feature in production in case of a release before the meta-optimizer release. So I guess we can avoid to do this refactor right now, cause potentially tomorrow in a release PR I will remove this check. Does it make sense to you?

renderer/src/routes/__root.tsx
peppescg: yep 👍

renderer/src/common/lib/meta-optimizer.ts
peppescg: mm I am not agree with this cause if we renamed a GET from `checkMetaOptimizerGroup` to `ensureMetaOptimizerGroup`, I would avoid to use the same logic for a POST. Also because we have `createMetaOptimizerGroup` and I cannot rename it in the same way

renderer/src/common/lib/meta-optimizer.ts
peppescg: 🤔 I am missing something with your suggestion. This is only the group name for the API, it is static on the constant, we don't need to do async stuff here

kantord: the point is that this way whenever you call the function that creates the Workload, it will inherently ensure that the group is already in place. Hence, the correct order of calling the two functions is automatically ensured & usage is simplified. 

More future proof, because the syntax itself guarantees the correct order

kantord (dismissed):
💯 


happy to merge as is if you don't agree with my last suggestion. it's a nit pick detail.

kantord (dismissed):

renderer/src/common/lib/meta-optimizer.ts
peppescg: The issue is that we need to handle the failures from group creation and avoid the workload creation in that specific case. Aslo it would be better to avoid calling other api calls if the group creation fails, and we can do this before.

PR #1116, opened 2025-10-21
feat(tool-filtering): add toggle all on table header, adjust apply button disable logic

samuv:
1. Add a switch toggle all on the table header for easily toggle all the tools
2. Adjust the Apply button logic to not save if there is no tools selected
3. Replaces the "Enable all tools" button with a "Cancel" button that navigates back
4. Fix back button on the page that should take in consideration the groups button
5. Add skeleton loading to table header
6. Adjust the switch position

https://github.com/user-attachments/assets/cc7f29b9-3c28-48c6-963b-5d617d11e782

### DIFF
--- a/renderer/src/features/mcp-servers/sub-pages/customize-tools/__tests__/page.test.tsx
+++ b/renderer/src/features/mcp-servers/sub-pages/customize-tools/__tests__/page.test.tsx
@@ -161,10 +161,10 @@ describe('Customize Tools Page - Converter Function Selection', () => {
         expect(screen.getByText('tool1')).toBeInTheDocument()
       })
 
-      const enableAllButton = screen.getByRole('button', {
-        name: /enable all tools/i,
+      const applyButton = screen.getByRole('button', {
+        name: /apply/i,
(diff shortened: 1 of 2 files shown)


renderer/src/features/mcp-servers/components/customize-tools-table.tsx
```diff
@@ -212,35 +232,65 @@ export function CustomizeToolsTable({
...
+                    <Switch
+                      className="ml-1"
+                      checked={isAllToolsEnabled}
+                      onCheckedChange={(checked) =>
+                        handleAllToolsToggle(checked)
+                      }
+                    />
+                  </TableHead>
+                  <TableHead className="text-muted-foreground text-xs">
+                    Tool
```
peppescg, line 272: nit: IMO it would more clear if the text would be Enable all tools

samuv: that is the tool name column, for the enable all tools we have directlu the switch in the table head

Issue #1097, opened 2025-10-17
[Task] Handle advanced meta-mcp configuration


peppescg:
### Description

we should have a dropdown in order to handle advanced configuration:
- logs
- restarts meta-mcp
- customize meta-mcp config ( network isolation should be hidden)

### Priority

Medium

### Task Type

Feature

PR #1120, opened 2025-10-21
feat(mcp-optimizer): allow user to edit workload

kantord:
fixes: https://github.com/stacklok/toolhive-studio/issues/1097 

https://github.com/user-attachments/assets/d07d9de5-14c1-4cdc-a7d9-c1923deb6e21

### DIFF
--- a/renderer/src/routes/__tests__/mcp-optimizer.test.tsx
+++ b/renderer/src/routes/__tests__/mcp-optimizer.test.tsx
@@ -516,20 +516,17 @@ it('radio button updates after editing ALLOWED_GROUPS via Customize Configuratio
     ).not.toBeInTheDocument()
   })
 
-  await waitFor(
-    () => {
-      const productionRadio = screen.getByRole('radio', {
-        name: /production/i,
-      }) as HTMLInputElement
-      const defaultRadio = screen.getByRole('radio', {
-        name: /default/i,
-      }) as HTMLInputElement
+  await waitFor(() => {
+    const productionRadio = screen.getByRole('radio', {
+      name: /production/i,
+    }) as HTMLInputElement
+    const defaultRadio = screen.getByRole('radio', {
+      name: /default/i,
+    }) as HTMLInputElement
 
-      expect(productionRadio).toBeChecked()
-      expect(defaultRadio).not.toBeChecked()
-    },
-    { timeout: 5000 }
-  )
+    expect(productionRadio).toBeChecked()
+    expect(defaultRadio).not.toBeChecked()
+  })
 })
 
 it('clicking Meta-MCP logs in Advanced menu navigates to logs page', async () => {


renderer/src/features/mcp-servers/components/local-mcp/dialog-form-local-mcp.tsx
```diff
@@ -183,12 +185,23 @@ export function DialogFormLocalMcp({
...
               serverName: data.name,
               groupName: data.group || groupName,
               isEditing,
             })
+            const queryKey = getApiV1BetaWorkloadsByNameQueryKey({
+              path: { name: data.name },
+            })
+
+            // Wait for the backend to persist changes before invalidating
+            await new Promise((resolve) => setTimeout(resolve, 500))
```
peppescg: I don't think this is correct, if the issue is to wait the `checkServerStatus` before making the invalidateQueries per serverName, I guess it is valuable to add directly in the `use-check-server-status` at this row:

`renderer/src/common/hooks/use-check-server-status.tsx`

We are already invalidate the workload list, I think it makes sense do the same for this other query

renderer/src/features/meta-mcp/components/group-selector-form.tsx
```diff
@@ -29,11 +30,18 @@ export function GroupSelectorForm({
...
+      selectedGroup: '',
     },
   })
 
+  // Reset form when defaultSelectedGroup changes to ensure radio buttons update
+  useEffect(() => {
+    form.reset({
+      selectedGroup: defaultSelectedGroup ?? '',
+    })
+  }, [defaultSelectedGroup, form])
```
peppescg: this is not needed, I handle this in my PR. This is the LLM that for fixing test is changing behavior 😆 
you only need to continue to use `values` rather than defaultValues, or having both, but at the same time values is the only one needed

renderer/src/features/mcp-servers/components/local-mcp/dialog-form-local-mcp.tsx
kantord: seems like this is not enough. i mean, i removed this logic and it does not work

kantord: but the timeout itself miight not be needed, let me check that one specifically

kantord: yup, the timeout is not needed

renderer/src/routes/__tests__/mcp-optimizer.test.tsx
```diff
@@ -440,6 +440,98 @@ it('Manage Clients button sends API requests to the correct mcp-optimizer group'
...
+        name: /production/i,
+      }) as HTMLInputElement
+      const defaultRadio = screen.getByRole('radio', {
+        name: /default/i,
+      }) as HTMLInputElement
+
+      expect(productionRadio).toBeChecked()
+      expect(defaultRadio).not.toBeChecked()
+    },
+    { timeout: 5000 }
```
peppescg: is this a leftover? we shouldn't need this

PR #1145, opened 2025-10-22
refactor: create a constant for the meta mcp docker image name

kantord:


kantord:
> TBH, I think we should rely on the registry image instead of using a constant

or the whole registry entry should be a constant?

samuv:
> > TBH, I think we should rely on the registry image instead of using a constant
> 
> or the whole registry entry should be a constant?

the entire registy is always the same for each thv version, so it is! alredy :)

kantord:
> > > TBH, I think we should rely on the registry image instead of using a constant
> > 
> > 
> > or the whole registry entry should be a constant?
> 
> the entire registy is always the same for each thv version, so it is! alredy :)

yup but i don't know if i want to connect a hardcoded mcp to an external dependency? :thinking:

PR #1147, opened 2025-10-22
feat(mcp-optimizer): enable host networking when native containers are used

kantord:
with this, MCP optimizer will work flawlessly on Linux for the vast majority of users

### DIFF
--- a/renderer/src/common/lib/constants.ts
+++ b/renderer/src/common/lib/constants.ts
@@ -3,27 +3,3 @@ export const META_MCP_SERVER_NAME = 'toolhive-mcp-optimizer'
 export const ALLOWED_GROUPS_ENV_VAR = 'ALLOWED_GROUPS'
 export const MCP_OPTIMIZER_REGISTRY_SERVER_NAME = 'mcp-optimizer'
 export const DEPRECATED_MCP_OPTIMIZER_REGISTRY_SERVER_NAME = 'meta-mcp'
-
-/**
- * Heuristic: are we likely running with a native container engine on the same machine?
- *
- * Rationale:
- * - On most Linux hosts, container engines (Docker/Podman) run natively on the same
- *   network stack as the host. In this case, the MCP Optimizer container needs host
- *   networking so it can talk to the ToolHive API on the host directly. We also set
- *   `TOOLHIVE_HOST=127.0.0.1` so the container resolves the API via loopback.
- * - On macOS/Windows, Docker Desktop typically runs in a VM. In that setup, host
- *   networking does not provide access to the macOS/Windows host. Instead, Docker
- *   exposes a special bridge (e.g. `host.docker.internal`) and/or forwarded ports,
- *   so we must NOT request host networking or override the host address.
- *
- * Why this is a heuristic:
- * - "Linux" is used here as a proxy for "native container engine on the same host".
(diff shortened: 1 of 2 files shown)


kantord:
@claude please do a deep review of this pr

renderer/src/common/lib/constants.ts
```diff
@@ -3,3 +3,14 @@ export const META_MCP_SERVER_NAME = 'toolhive-mcp-optimizer'
...
+/**
+ * Determines if the platform is probably using native containers.
+ * Native containers mean Docker runs directly on the host OS (Linux),
+ * as opposed to running in a VM (macOS/Windows with Docker Desktop).
+ * When native containers are used, host networking mode allows direct
+ * access to the host's network stack.
+ */
+export function isProbablyUsingNativeContainers(): boolean {
+  return window.electronAPI.isLinux
+}
```
peppescg: I would remove this from here, this is only for constant, and considering that this fn is used only on a single file (`renderer/src/features/meta-mcp/hooks/use-create-optimizer-workload.ts`) you can call directly`window.electronAPI.isLinux`

renderer/src/features/meta-mcp/hooks/use-create-optimizer-workload.ts
```diff
@@ -106,16 +107,33 @@ export function useCreateOptimizerWorkload() {
     optimized_workloads: string[]
   }) => {
     if (!isMetaOptimizerEnabled) return
+
+    // On platforms with native containers (Linux), use host networking mode
+    // to allow the container to access the host's ToolHive API
+    const useHostNetworking = isProbablyUsingNativeContainers()
```
peppescg: wdyt to replace `useHostNetworking` with `needsHostNetworking`, the prefix `use` is misleading in react considering the hook

kantord: great point

renderer/src/common/lib/constants.ts
kantord: I don't agree with directly calling `isLinux`, because that does not correctly capture the idea that `isLinux` is used as a heuristic to determine the location of the container engine :crying_cat_face: this is not strictly speaking the fully correct logic. but we could not yet identify a more correct way to do it. Nevertheless, the fully correct implementation would probably not even include `isLinux`

peppescg: but this is what you're doing here, you are calling isLinux, and also you put a function inside a constant file, at least you cna move this fn  where it is used, cause it is used only in a specific file, so it shouldn't be in common, cause it is not shared with multiple files

kantord: true, I can move it into the same file. and I can improve the documentation to actually explain what is happening there

peppescg: @kantord I was wondering if does it make sense to specify in the documentation this use case 👍

kantord: pushed

peppescg (approved):
🚢

renderer/src/common/lib/constants.ts
peppescg: sorry, I meant in our documentation repo not only in the code

PR #1148, opened 2025-10-22
feat: handle clients on optimizer page submit, remove manage clients

peppescg:
Handle Optimizer clients directly leveraging the selected group, we can derive the clients from the selected group.
Changing group from Optimizer and submitting the form, will apply changes following the group clients selected.
In the specific group page, updating the clients will update the Optimizer clients as well

https://github.com/user-attachments/assets/08b1d2f8-7f88-4b7e-8b78-1eaa2672d39d

### DIFF
--- a/renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
+++ b/renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
@@ -33,7 +33,8 @@ describe('ManageClientsButton – BDD flows', () => {
     vi.mocked(useFeatureFlag).mockReturnValue(true)
     vi.mocked(useMcpOptimizerClients).mockReturnValue({
       saveGroupClients: saveGroupClientsMock,
-    })
+      restoreClientsToGroup: vi.fn(),
+    } as unknown as ReturnType<typeof useMcpOptimizerClients>)
     vi.clearAllMocks()
   })
 


renderer/src/features/clients/hooks/use-manage-clients.ts
```diff
@@ -33,7 +33,7 @@ export function useManageClients(groupName: string) {
   const queryClient = useQueryClient()
 
   const { data: groupsData } = useQuery({
-    queryKey: ['api', 'v1beta', 'groups'],
+    queryKey: getApiV1BetaClientsQueryKey(),
```
samuv: I think this is not correct, getApiV1BetaGroupsQueryKey?

renderer/src/features/clients/hooks/use-manage-clients.ts
samuv: Should this be updated as well?

renderer/src/features/clients/hooks/use-manage-clients.ts
peppescg: ooh autocomplete...thanks

renderer/src/features/clients/hooks/use-manage-clients.ts
peppescg: yep

samuv (approved):
Tested the flow and it seems consistent 💪🏼 
Clients are correct in the various use cases

PR #1169, opened 2025-10-24
fix(mcp-optimizer): only show enabled MCPs

kantord:
https://github.com/user-attachments/assets/e5bc6c79-27c1-47aa-923b-061880ddacb7


peppescg (approved):
thanks 🙌

PR #1174, opened 2025-10-24
fix: refine labels and messages

danbarr:
Some refinements/polishing of labels and messages related to MCP Optimizer

Signed-off-by: Dan Barr <6922515+danbarr@users.noreply.github.com>


samuv (approved):
Nice, thank you @danbarr

PR #1178, opened 2025-10-24
fix: remove optimizer workload if optimized group is deleted

peppescg:
Remove optimizer workload in case of optimized group deletion. Remove the clients from optimized group as well.

**Issue react-query cache for optimizer workload**
There a bunch of query on optimizer workload using `gcTime: 0`, this can cause issue trying to invalidate it, cause the query could be not present in the cache or refetching it, `retry` is by default 3, could add the query after invalidate it.
In order to remove this issue, I removed `gcTime` and also set `retry: false`.

For fixing a cache issue into optimizer page I invoked a reset of the query, by optimizer workload name, on `beforeLoad` in optimizer page tanstack/router level. In this way we are sure to get always the fresh data.

______

https://github.com/user-attachments/assets/5d575f6b-4263-4072-9284-2013bf39cee3

### DIFF
--- a/renderer/src/features/meta-mcp/components/group-selector-form.tsx
+++ b/renderer/src/features/meta-mcp/components/group-selector-form.tsx
@@ -37,9 +37,12 @@ type FormSchema = z.infer<typeof formSchema>
 export function GroupSelectorForm({
   groups,
 }: GroupSelectorFormProps): ReactElement {
-  const { data: metaMcpConfig } = useMetaMcpConfig()
+  const { data: metaMcpConfig, isError: isMetaMcpConfigError } =
+    useMetaMcpConfig()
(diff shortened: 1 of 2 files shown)


renderer/src/routes/group.$groupName.tsx
```diff
@@ -124,12 +124,13 @@ function GroupRoute() {
                   <DropdownMenuRunMcpServer
                     openRunCommandDialog={setServerDialogOpen}
                   />
-                  <ManageClientsButton
-                    isOptimizedGroupName={isOptimizedGroupName}
-                    groupName={groupName}
-                  />
                 </>
               )}
+              <ManageClientsButton
```
samuv, line 129: 👌🏼

PR #1206, opened 2025-10-30
chore: simplify running with different thv binary

kantord:
lets you connect to a custom `thv serve` using `THV_PORT`

screenshot-2025-11-03_10-51-01

### DIFF
--- a/renderer/src/common/components/custom-port-banner.tsx
+++ b/renderer/src/common/components/custom-port-banner.tsx
@@ -34,9 +34,9 @@ export function CustomPortBanner() {
 
   return (
     <Alert
-      variant="warning"
       className="fixed bottom-4 left-1/2 z-50 w-auto max-w-[95vw]
-        -translate-x-1/2"
+        -translate-x-1/2 border-yellow-200 bg-yellow-50 text-yellow-900
+        dark:border-yellow-800 dark:bg-yellow-950 dark:text-yellow-100"
     >
       <AlertTriangle />
       <AlertDescription className="flex items-start gap-2">
--- a/renderer/src/common/components/ui/alert.tsx
+++ b/renderer/src/common/components/ui/alert.tsx
@@ -11,8 +11,6 @@ const alertVariants = cva(
         default: 'bg-card text-card-foreground',
         destructive:
           'text-destructive bg-card [&>svg]:text-current *:data-[slot=alert-description]:text-destructive/90',
-        warning:
-          'border-yellow-200 bg-yellow-50 text-yellow-900 dark:border-yellow-800 dark:bg-yellow-950 dark:text-yellow-100',
       },
     },
     defaultVariants: {
--- a/renderer/src/routes/__root.tsx
+++ b/renderer/src/routes/__root.tsx
@@ -54,9 +54,7 @@ function RootComponent() {
   return (
     <>
       {!isShutdownRoute && <TopNav />}
-      {!isShutdownRoute &&
-        import.meta.env.DEV &&
-        !!import.meta.env.THV_PORT && <CustomPortBanner />}
+      {!isShutdownRoute && import.meta.env.DEV && <CustomPortBanner />}
       <Main>
         <Outlet />
         <Toaster


kantord:
@claude  can you please do a deep review of this pr

kantord:
@claude can you please do a deep review of this pr

kantord:
Hi @claude Thanks! Kept this dev‑only and minimal: .thv_bin is now just a plain path (empty = default), watcher is debounced and cleaned up, and version
  reads thv version with a short timeout (no misleading fallback). Skipped extra robustness/tests since this is dev‑only and non‑critical; happy to
  revisit if it proves flaky.

Anything else you still think is critical?

main/src/toolhive-manager.ts
```diff
@@ -103,6 +110,20 @@ async function findFreePort(
...
+    if (isUsingCustomPort()) {
+      const customPort = parseInt(process.env.THV_PORT!, 10)
+      if (isNaN(customPort)) {
+        log.error(
+          `Invalid THV_PORT environment variable: ${process.env.THV_PORT}`
+        )
+        return
+      }
+      toolhivePort = customPort
+      toolhiveMcpPort = undefined
```
peppescg: I wouldn't avoid to set this as undefined, the changes should only related to toolhivePort 🤔

renderer/src/routes/__root.tsx
```diff
@@ -53,6 +54,7 @@ function RootComponent() {
   return (
     <>
       {!isShutdownRoute && <TopNav />}
+      {!isShutdownRoute && import.meta.env.DEV && <CustomPortBanner />}
```
peppescg, line 57: I am not totally convinced putting a warning for this use case, cause we already know that we are connecting a different thv binary, but anyways, I would also check directly here `import.meta.env.THV_PORT` if it is not set we don't need to render the component at all

kantord: good point

main/src/toolhive-manager.ts
kantord: ah, the actual problem here is that for the thv mcp to work we need to set a port.... which means you have to start the thv process in a special way to begin with 😸 so I guess the real solution is to get it from a second ENV var.

peppescg: I would avoid to set another env var for it,  `--experimental-mcp-port=${toolhiveMcpPort}`, let's assume that it is up to us passing the flag where we are running thv serve, but anyway it is a small nit.

README.md
```diff
@@ -94,6 +94,26 @@ Kubernetes Operator. Learn more in the
...
+
+2. Set the `THV_PORT` environment variable and start the dev server:
+   ```bash
+   THV_PORT=50000 pnpm start
+   ```
+
+The UI will display a warning banner showing the HTTP address it's connecting to
+when using a custom port. This only works in development mode; packaged builds
+always use the embedded binary.
+
```
peppescg: I would probably leverage only the /doc/README file that is specilized on developer guide

main/src/toolhive-manager.ts
kantord: I mean you kinda have to pass the env var, there is no other way to make all features (playground) fully work. I tested it, and it fully works only if you pass this env var (of course). although, the overall app does not break if you don't pass it.

I also realized that MCP optimizer won't work if you don't start the server in the correct port range, because they will only do port scanning on a limited range. So I added this to the readme as well

README.md
kantord: true

renderer/src/common/components/custom-port-banner.tsx
```diff
@@ -0,0 +1,55 @@
...
+    Promise.all([
+      window.electronAPI.isUsingCustomPort(),
+      window.electronAPI.getToolhivePort(),
+    ])
+      .then(([usingCustom, toolhivePort]) => {
+        setIsCustomPort(usingCustom)
+        setPort(toolhivePort)
+      })
+      .catch((error: unknown) => {
+        console.error('Failed to get custom port info:', error)
```
kantord: use log instead of console.log

renderer/src/common/components/ui/alert.tsx
```diff
@@ -11,6 +11,8 @@ const alertVariants = cva(
         default: 'bg-card text-card-foreground',
         destructive:
           'text-destructive bg-card [&>svg]:text-current *:data-[slot=alert-description]:text-destructive/90',
+        warning:
```
kantord: remove warning style from here and instead override in the component itself (because the warning is not an official design)

PR #1228, opened 2025-11-03
feat(playground): add ollama

samuv:
as per title, this add ollama as a provide for the playground.

This is not the final implementation, I'm using the api key store for saving the endpoint of Ollama, is just for testing the integration


main/src/chat/providers.ts
```diff
@@ -228,6 +234,27 @@ export const CHAT_PROVIDERS: ChatProvider[] = [
...
+    name: 'Ollama',
+    models: CHAT_PROVIDER_INFO.find((p) => p.id === 'ollama')?.models || [],
+    createModel: (modelId: string, apiKey: string) => {
+      // For Ollama, the apiKey field can be used to specify a custom base URL
+      // If empty or not a URL, default to localhost:11434
+      const baseURL =
+        apiKey &&
+        (apiKey.startsWith('http://') || apiKey.startsWith('https://'))
+          ? apiKey
+          : 'http://localhost:11434'
```
JAORMX, line 248: should this URI be configurable as well?

main/src/chat/providers.ts
```diff
@@ -228,6 +234,27 @@ export const CHAT_PROVIDERS: ChatProvider[] = [
...
+  {
+    id: 'ollama',
+    name: 'Ollama',
+    models: CHAT_PROVIDER_INFO.find((p) => p.id === 'ollama')?.models || [],
+    createModel: (modelId: string, apiKey: string) => {
+      // For Ollama, the apiKey field can be used to specify a custom base URL
+      // If empty or not a URL, default to localhost:11434
+      const baseURL =
+        apiKey &&
+        (apiKey.startsWith('http://') || apiKey.startsWith('https://'))
```
JAORMX, line 246: why would the API key start with `http://`

main/src/chat/providers.ts
samuv: It would be, at the moment it’s just a hacky implementation because we’re using the same storage we use for API keys 🙈
Screenshot 2025-11-03 at 10 03 02

main/src/chat/providers.ts
samuv: I’d need to extend the schema of those configurations to include a separate key for the provider URL.  
This is just the quickest and hackiest way to test it for now

kantord:
A "use default" button would be nice ( just fill in the default value)

but also, is there an endpoint that is cheap to query (no need to load stuff into GPU) that we can use to discover if ollama is running? if so, we could just make a single request to the default endpoint if ollama is not yet configured. then, if it was successful, we could auto-configure it :thinking:

kantord:
I tested it locally and works really well!

peppescg:
tested with qwen3 and it works pretty well!

samuv:
closing this in favor of https://github.com/stacklok/toolhive-studio/pull/1238

samuv:
> A "use default" button would be nice ( just fill in the default value)
> 
> but also, is there an endpoint that is cheap to query (no need to load stuff into GPU) that we can use to discover if ollama is running? if so, we could just make a single request to the default endpoint if ollama is not yet configured. then, if it was successful, we could auto-configure it 🤔

Yeah, I was thinking the same, just auto-configure it.  But at the same time, I don’t want to add something the user might not want.  For example, if they’re running an instance on a local network, they might prefer using an IP instead of `localhost`.  
It feels safer to let the user decide rather than configuring it automatically.

PR #1230, opened 2025-11-03
feat(customize-tools): add tool overide to customize table

samuv:
## Overview

Adds the ability to override tool names and descriptions in the customize tools table. Users can customize how tools appear, with changes tracked locally before being applied to the server configuration.

https://github.com/user-attachments/assets/ecafcc12-33c9-4123-ad5f-ed8778271555

https://github.com/user-attachments/assets/9933b9d4-6ae7-49ea-8121-6f57859137c0

https://github.com/user-attachments/assets/24379f00-5b98-4392-8faf-edcc12b12a8a

### Default Display
- **No filter, no overrides**: All tools from registry/server appear with their original names and descriptions
- All tools are enabled by default

### Tool Filtering (Allowlist)
- **Filter applied**: All tools appear in the table, but switches are enabled/disabled based on the allowlist
- Tools not in the filter appear but are disabled
- Enables selective tool usage while maintaining visibility

### Name Override
- **Name override applied**: Tool appears with the overridden name; original name is hidden
- When drift occurs (not all tools enabled), the override name is included in the `tools` array, not the original name
- Original name tracked as `originalName` for reference

### Description Override
- **Description-only override**: Tool keeps its original name but displays the overridden description
- Original description shown as helper text in edit modal (unless override exists)
- Works independently of name overrides

### Combined Overrides with Filtering
- **Filter + description override**: All tools appear, switches respect the filter, descriptions show overrides
- Override name takes precedence when checking filter allowlist

### Edit Workflow
1. Click "Edit" → Modal opens with current tool name/description
2. Modify fields → Changes tracked locally (not saved yet)
3. Save → Local override applied (table updates immediately)
4. Apply → Changes sent to server via `tools_override` field

### Local Override Persistence
- After saving a local override, reopening the edit dialog shows the local override values
- Local overrides take priority over saved overrides until "Apply" is clicked
- Original values displayed as helper text for reference

### Tool Building (`buildToolsWithOverrides`)
Merges registry tools, server tools, and overrides:
- **Name overrides**: Original tool excluded, appears with new name
- **Description overrides**: Tool name unchanged, description replaced
- **Filter**: Enabled state based on display names (not original names)
- **Drift**: Tools missing from registry/server still appear

### State Management (`useCustomizeToolsTable`)
- Tracks `enabledTools` (toggle switches), `toolsOverride` (local edits), and `editState` (modal)
- Detects changes by comparing local vs saved overrides and enabled state vs initial state
- Navigation blocker prevents leaving with unsaved changes

### Override Logic
- Local overrides take priority over saved overrides
- Resetting to original explicitly clears saved overrides
- Empty overrides (both name and description empty) are filtered out
- When description override exists, original description becomes unavailable (N/A)

## UI Features

- **Override indicator**: Wrench icon (orange = unsaved, primary = saved)
- **Expandable text**: Long descriptions truncated with "Show more/less"
- **Edit modal**: Shows original values as helper text, current values in inputs

### DIFF
--- a/renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
+++ b/renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
@@ -190,6 +190,16 @@ export function useCustomizeToolsTable({
     })
   }
 
+  const resetEditState = () => {
+    setEditState({
+      isOpen: false,
+      tool: null,
+      name: '',
+      description: '',
+      hasOverrideDescription: false,
+    })
+  }
+
   const handleSaveToolOverride = () => {
     const { tool } = editState
     if (!tool) return
@@ -251,26 +261,14 @@ export function useCustomizeToolsTable({
       })
     }
 
-    setEditState({
-      isOpen: false,
-      tool: null,
-      name: '',
-      description: '',
-      hasOverrideDescription: false,
-    })
+    resetEditState()
   }
 
   const handleCloseEditModal = () => {
     trackEvent('Customize Tools: close edit tool modal', {
       tool_name: editState.tool?.name || '',
     })
-    setEditState({
-      isOpen: false,
-      tool: null,
-      name: '',
-      description: '',
-      hasOverrideDescription: false,
-    })
+    resetEditState()
   }
 
   const handleNameChange = (name: string) => {
--- a/renderer/src/features/mcp-servers/lib/tool-override-utils.ts
+++ b/renderer/src/features/mcp-servers/lib/tool-override-utils.ts
@@ -216,30 +216,29 @@ function mergeField(
     return newValue
   }
 
-  // Field wasn't changed - preserve previous local value if it exists and differs from saved
-  // This should happen even if there are no other changes (prevLocalOverride needs to be preserved)
-  if (prevLocalValue !== undefined) {
-    const shouldPreserve = shouldPreserveField(prevLocalValue, existingValue)
-    if (shouldPreserve) {
-      return prevLocalValue
-    }
-    // If prevLocalValue is same as existingValue and there are other changes,
-    // we still need to preserve it because we're updating the override
-    // This happens when toolsOverride was initialized from overrideTools
-    if (hasAnyChanges && existingValue !== undefined) {
-      return existingValue
-    }
-    // If prevLocalValue is same as existingValue and no other changes, return undefined
-    // to avoid creating redundant override
-    return undefined
+  // Field wasn't changed - handle preservation logic
+  // No previous local value - preserve existing saved override only if there are other changes
+  if (prevLocalValue === undefined) {
+    return hasAnyChanges && existingValue !== undefined
+      ? existingValue
+      : undefined
   }
 
-  // If no previous local value but there's an existing saved override, preserve it
-  // only if there are other changes (to avoid preserving saved overrides when nothing changed)
+  // There's a previous local value - check if it should be preserved
+  const shouldPreserve = shouldPreserveField(prevLocalValue, existingValue)
+  if (shouldPreserve) {
+    return prevLocalValue
+  }
+
+  // prevLocalValue is same as existingValue
+  // If there are other changes, preserve existingValue (updating the override)
+  // This happens when toolsOverride was initialized from overrideTools
   if (hasAnyChanges && existingValue !== undefined) {
     return existingValue
   }
 
+  // No other changes and prevLocalValue matches existingValue - return undefined
+  // to avoid creating redundant override
   return undefined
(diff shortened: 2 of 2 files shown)


renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
```diff
@@ -0,0 +1,302 @@
...
+          }
+
+          // Tool set is the same but isInitialEnabled might have changed
+          // (e.g., when a filter is applied/removed). Since the key changed, reset to new initial state.
+          // This ensures that when a filter changes, tools are enabled/disabled correctly.
+          toolsInitializedRef.current = true
+          previousToolNamesRef.current = currentToolNames
+          return initialState
+        })
+      }
```
peppescg, line 128: it is a bit hard to read this, considering that it is also inside a useEffect

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
```diff
@@ -0,0 +1,302 @@
...
+      })
+    }
+
+    setEditState({
+      isOpen: false,
+      tool: null,
+      name: '',
+      description: '',
+      hasOverrideDescription: false,
+    })
```
peppescg: considering that below there is the same reset, I would create a fn `resetEditState` avoiding duplicate it

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
samuv: This handles the case where tools change after loading.  It’s an edge case, but it can happen.  
It’s not exported to an external function because it mostly uses two refs and a state setter.

renderer/src/features/mcp-servers/lib/tool-override-utils.ts
```diff
@@ -0,0 +1,354 @@
...
+      // If there's a saved override (existingDescription differs from original),
+      // we need to explicitly set description to original to override the saved one
+      override.description =
+        existingDescription !== originalDescription
+          ? originalDescription
+          : undefined
+    } else {
+      override.description = editedDescription
+    }
+  }
```
peppescg, line 183: I guess it is possible to simplify a bit avoiding those if/else nested improving the readability

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
samuv: yeah it makes sense

renderer/src/features/mcp-servers/lib/tool-override-utils.ts
samuv: yeah I think we can revisit this and avoid few if

peppescg:
Screenshot 2025-11-03 at 16 49 49

we are not waiting that the server is up and running after the edit, and the status 🤔

peppescg:
is it ok that after override the `convert_time` tool, I see my change `_test` but I also see the original one? 🤔 


Screenshot 2025-11-03 at 16 52 42

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
peppescg: yep I meant it could be inside a fn in the same file, only for readability...but it is not a blocker at all

samuv:
> 
> we are not waiting that the server is up and running after the edit, and the status 🤔

I think this is related to the [Issue](https://github.com/stacklok/toolhive/issues/2350) on thv side

renderer/src/features/mcp-servers/lib/tool-override-utils.ts
samuv: done

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
samuv: done

renderer/src/features/mcp-servers/hooks/use-customize-tools-table.ts
samuv: BTW, this is only needed because we don’t have a dedicated endpoint that give us those info. Most of this logic can be removed once we have one in place.

samuv:
@claude do a review on this PR

> is it ok that after override the `convert_time` tool, I see my change `_test` but I also see the original one? 🤔
> 

this part was fixed [here](https://github.com/stacklok/toolhive/pull/2386), we just need to wait the last thv, you can test it running thv main locally

PR #1232, opened 2025-11-03
feat(chat): add export functionality for chat conversations

ptelang:
Adds ability to export chat history in multiple formats (Markdown, JSON, and Text) to enable users to save and share their conversations. The export feature includes:

- Export dropdown menu in chat interface
- Comprehensive formatting for each export type
- Tool call information and metadata preservation
- Automatic file download with timestamped filenames

### DIFF
--- a/renderer/src/features/chat/components/chat-interface.tsx
+++ b/renderer/src/features/chat/components/chat-interface.tsx
@@ -6,6 +6,10 @@ import {
   MessageCircleMore,
   ChevronDown,
   Trash2,
+  Download,
+  FileText,
+  FileJson,
+  File,
 } from 'lucide-react'
 import { ChatMessage } from './chat-message'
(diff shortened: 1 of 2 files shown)


samuv:
@ptelang how important is it to support multiple download formats?  
If we stick with the original JSON file, we can support both export and chat import.  But if we introduce multiple formats, that wouldn’t be possible, which is why I was waiting a bit before introducing this feature

ptelang:
> @ptelang how important is it to support multiple download formats? If we stick with the original JSON file, we can support both export and chat import. But if we introduce multiple formats, that wouldn’t be possible, which is why I was waiting a bit before introducing this feature

If a customer wants to export and feed the conversation back to another LLM, markdown format will be more suitable. So it will be good to support at least those JSON and markdown formats for export.

PR #1234, opened 2025-11-03
feat: implement group detail page in registry

kantord:
## Registry group

screenshot-2025-11-04_14-06-42

(for testing info see: https://github.com/stacklok/toolhive-studio/pull/1038)

## Registry server (unchanged!)

this is unchanged, but the file was affected by the pr. since we don't have visual regression testing, I'm just including a screenshot just in case

screenshot-2025-11-04_14-08-25

## Empty state

there is a fallback behavior for empty groups - this is of course not a "valid" use case, but there is nothing stopping the user from creating a registry JSON with an empty group, so we need to handle it. for course in this case, there is no install button

screenshot-2025-11-04_14-05-17

### DIFF
--- a/renderer/src/routes/(registry)/registry-group_.$name.tsx
+++ b/renderer/src/routes/(registry)/registry-group_.$name.tsx
@@ -95,7 +95,7 @@ export function RegistryGroupDetail() {
           </div>
         </>
       ) : (
-        <Alert className="mt-6">
+        <Alert className="mt-6 max-w-2xl">
           <Info />
           <AlertDescription>
             This group does not have any servers.


kantord:
@claude can you please do a deep review of this pr

kantord:
@claude thanks. the non functional button is not an issue, that won't be implemented in this pr. don't worry, the groups-in-registry feature is behind a feature flag, and users won't be able to see this page. 

Regarding the hardcoded registry: currently, the UI does not support using multiple registries at the same time: we can just switch what the registry is, and that will apply to the entire app. The implementation that we have there is, to my understanding, consistent with what we can see elsewhere in the app. 

not having a specialized 404 page here is acceptable because this is an error that should never happen in practice: this is an electron app, not a web app, and we only link to to this page when the group actually exists. we still have a generic error handling system in the app, so that should cover for such rare error cases

for now, we will not have a navigation to the server detail page. 


addressed the 'empty state' issue. can you re-review with that in mind?

renderer/src/routes/(registry)/registry-group_.$name.tsx
```diff
@@ -0,0 +1,107 @@
...
+              </TableBody>
+            </Table>
+          </div>
+          <Separator className="my-6" />
+          <div className="flex gap-5">
+            <Button variant="default">
+              <Wrench className="size-4" />
+              Create group
+            </Button>
+          </div>
```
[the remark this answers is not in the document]
kantord: no need, this is behind a feature flag

peppescg:
should the width be cut? it seems weird to have it full with without any long text

Screenshot 2025-11-04 at 14 29 32

renderer/src/features/registry-servers/components/grid-cards-registry.tsx
```diff
@@ -68,10 +68,17 @@ export function GridCardsRegistry({
...
+                navigate({
+                  to: '/registry-group/$name',
+                  params: { name: item.name! },
+                })
+              } else {
+                navigate({
+                  to: '/registry/$name',
+                  params: { name: item.name! },
+                })
+              }
```
peppescg: the only change is on `to`

```suggestion
            navigate({
                  to: item.type === 'group' ? '/registry/$name' : '/registry-group/$name',
                  params: { name: item.name! },
                })
```

renderer/src/routes/(registry)/registry-group_.$name.tsx
```diff
@@ -0,0 +1,107 @@
...
+        title={name}
+        badges={
+          <Badge
+            variant="secondary"
+            className="bg-foreground/5 w-fit rounded-full px-2.5 py-0.5"
+          >
+            Group
+          </Badge>
+        }
+        description={group?.description ?? undefined}
```
peppescg: `undefined` is needed? I think `group?.description` returns already undefined

kantord:
@peppescg  there you go

screenshot-2025-11-04_16-10-46

peppescg (approved):

PR #1236, opened 2025-11-04
feat: improve ux when update not yet available from server

peppescg:
## Issue
When a new release is published, GitHub Pages updates immediately but the Electron update server (`update.electronjs.org`) takes ~10 minutes to sync. This caused confusion when users clicked "Download" right after seeing a new version available, stacking in the page without any feedback.

## Solution
- Added `not-available` state to track when update check finds no update
- Show "Checking..." button text for both `checking` and `not-available` states
- Display toast notification: *"Update not yet available. The update server is still processing the latest release. Please try again in a few minutes."*

## Changes
- **Main**: Added `not-available` to `UpdateState` type, modified `handleUpdateNotAvailable()` 
- **Preload**: Updated type definitions to include new state
- **UI**: Added toast notification on state change, improved button text logic
- **Hook**: Updated `useCurrentUpdateState()` to handle new state
- **Tests**: Added coverage for `not-available` state behavior

## Result
Better UX when update server lags behind GitHub Pages, with clear user feedback instead of silent failure.

### DIFF
--- a/renderer/src/common/components/settings/tabs/__tests__/version-tab.test.tsx
+++ b/renderer/src/common/components/settings/tabs/__tests__/version-tab.test.tsx
@@ -314,6 +314,7 @@ describe('VersionTab', () => {
     expect(button).toBeDisabled()
 
     expect(toast.info).toHaveBeenCalledWith('Update not yet available', {
+      id: 'update-not-available',
       description:
         'The update server is still processing the latest release. Please try again in a few minutes.',
       duration: 6000,
--- a/renderer/src/common/components/settings/tabs/version-tab.tsx
+++ b/renderer/src/common/components/settings/tabs/version-tab.tsx
@@ -50,6 +50,7 @@ export function VersionTab({ appInfo, isLoading, error }: VersionTabProps) {
   useEffect(() => {
     if (updateState === 'not-available') {
       toast.info('Update not yet available', {
+        id: 'update-not-available',
         description:
           'The update server is still processing the latest release. Please try again in a few minutes.',
         duration: 6000,


peppescg:
@claude review the pr

renderer/src/common/components/settings/tabs/__tests__/version-tab.test.tsx
```diff
@@ -5,6 +5,13 @@ import { PromptProvider } from '@/common/contexts/prompt/provider'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import type { AppVersionInfo } from '@/common/hooks/use-app-version'
 import userEvent from '@testing-library/user-event'
+import { toast } from 'sonner'
+
+vi.mock('sonner', () => ({
```
kantord, line 10: probably not necessary to do a one-off mock here, i think this should already be mocked globally

kantord: (I guess that the mock could be importable, in order to be able to verify what it was called with)

peppescg: probably yes, I remember that we talked alrady of this 😆 let me remove it

renderer/src/common/components/settings/tabs/__tests__/version-tab.test.tsx
```diff
@@ -269,4 +277,119 @@ describe('VersionTab', () => {
...
+      </PromptProvider>
+    )
+
+    const button = screen.getByRole('button', { name: /Checking/i })
+    expect(button).toBeVisible()
+    expect(button).toBeDisabled()
+  })
+
+  it('shows "Downloading..." when update state is downloading', () => {
+    import.meta.env.MODE = 'production'
```
kantord, line 361: we could probably have a global helper function for this, or actually, is it wrong to make this the default behavior across all tests? 🤔

peppescg: nope cause we are using it dev vs prod in order to check the behavior, in a bunch of tests file

renderer/src/common/components/settings/tabs/__tests__/version-tab.test.tsx
peppescg: the mock is needed cause the test check specific info sonner notification, and there are a lot of tests with custom mocks, so we don't have right now it mocked globally

PR #1238, opened 2025-11-05
feat(playground): add Ollama and LM studio as local providers

samuv:
This PR adds support for local AI providers (Ollama and LM Studio) to ToolHive Studio, allowing users to run AI models locally without requiring cloud API keys.

https://github.com/user-attachments/assets/6135d084-0127-445c-9e75-7c992c6799d0

### What's New

- **Ollama Integration**: Connect to local Ollama instances with automatic model discovery
- **LM Studio Integration**: Connect to LM Studio's OpenAI-compatible API
- **Dynamic Model Discovery**: Automatically fetch available models from local servers with refresh capability
- **Provider-Specific UI**: Context-aware settings dialog with URL configuration for local providers

### Technical Changes

- Refactored provider settings to use discriminated unions (`endpointURL` for local servers, `apiKey` for cloud providers)
- Added type guards throughout for type-safe credential handling
- Enriched message metadata with `providerId` for accurate chat history icons
- Implemented graceful handling for providers that don't report token usage (LM Studio)
- Added NaN protection to token usage display
- Centralized constants and utilities to resolve circular dependencies

### User Experience

- Local providers are hidden when not configured or offline
- Refresh button validates server URLs before saving
- Clear messaging when token usage is unavailable
- Provider icons in chat history for easy identification

### Dependencies

- Added `@ai-sdk/openai-compatible` for LM Studio
- Added `ai-sdk-ollama` for native Ollama support

### DIFF
--- a/main/src/chat/constants.ts
+++ b/main/src/chat/constants.ts
@@ -166,4 +166,4 @@ export const CHAT_PROVIDER_INFO: ChatProviderInfo[] = [
       'zhipuai/glm-4.5-thinking',
     ],
   },
-]
+] as const satisfies ChatProviderInfo[]


samuv:
@claude please please review this PR, but keep the feedback concise

peppescg:
I think this warning is misleading, cause the user needs to click on refresh icon in order to check the connection but it doesn't mean that the model is not running. 

Probably I start automatically the check `onMouseLeave` or with a debounce if we want to show the warning

Screenshot 2025-11-05 at 14 09 13

peppescg:
navigating to playground there is always this issue, cause `baseUrl` is empty string

Screenshot 2025-11-05 at 14 14 21

samuv:
> I think this warning is misleading, cause the user needs to click on refresh icon in order to check the connection but it doesn't mean that the model is not running.
> 
> Probably I start automatically the check `onMouseLeave` or with a debounce if we want to show the warning
> 
good catch, I will check if the endpoint is valorized before

> navigating to playground there is always this issue, cause `baseUrl` is empty string
> 
yep, added the skip if baseURL is not valorized

main/src/chat/constants.ts
```diff
@@ -0,0 +1,165 @@
...
+
+      // Moonshot AI (Kimi) models
+      'moonshot/kimi-k2',
+
+      // Zhipu AI (GLM) models
+      'zhipuai/glm-4.5',
+      'zhipuai/glm-4.5-thinking',
+    ],
+  },
+]
```
peppescg: I would add `satisfies ChatProviderInfo[]` in order to enforce i literal types

main/src/chat/settings-storage.ts
```diff
@@ -76,28 +91,99 @@ const chatStore = new Store<ChatSettings>({
...
+export function getChatSettings(providerId: ProviderId): ChatSettingsProvider {
   try {
     const providers = chatStore.get('providers')
     if (isProvidersRecord(providers)) {
-      return providers[providerId] || { apiKey: '', enabledTools: [] }
+      const existing = providers[providerId]
+      if (existing) return existing
+
+      // Return default based on provider type
+      if (providerId === 'ollama' || providerId === 'lmstudio') {
```
peppescg: it seems repeatable a bunch of times
```
const LOCAL_PROVIDER_MODELS_ID= ['ollama', 'lmstudio']
...
if( LOCAL_PROVIDER_MODELS_ID.includes(providerId))
```

main/src/chat/streaming.ts
```diff
@@ -43,8 +44,15 @@ export async function handleChatStreamRealtime(
...
-        // Create AI model
-        const model = provider.createModel(request.model, request.apiKey)
+        // Create AI model using type guards for discriminated union
+        const model = isLocalServerRequest(request)
+          ? provider.createModel(request.model, request.endpointURL)
+          : hasApiKey(request)
+            ? provider.createModel(request.model, request.apiKey)
+            : (() => {
+                throw new Error('Invalid request: missing credentials')
+              })()
```
peppescg: I guess this could be simply 🤔 
```suggestion
            :  throw new Error('Invalid request: missing credentials')          
```

main/src/chat/types.ts
```diff
@@ -13,15 +14,24 @@ interface MessageMetadata {
...
+      enabledTools?: string[]
+    }
+  | {
+      chatId: string
+      messages: ChatUIMessage[]
+      provider: string
+      model: string
+      apiKey: string
+      enabledTools?: string[]
+    }
```
peppescg:
```suggestion
export type BaseChatRequest = {
  chatId: string
  messages: ChatUIMessage[]
  model: string
  enabledTools?: string[]
}

export type ChatRequest =
  | (BaseChatRequest & {
      provider: 'ollama' | 'lmstudio'
      endpointURL: string
    })
  | (BaseChatRequest & {
      provider: Exclude<string, 'ollama' | 'lmstudio'>
      apiKey: string
    })
```

preload/src/preload.ts
```diff
@@ -133,21 +133,41 @@ contextBridge.exposeInMainWorld('electronAPI', {
...
+            enabledTools?: string[]
+          }
+        | {
+            chatId: string
+            messages: ChatUIMessage[]
+            provider: string
+            model: string
+            apiKey: string
+            enabledTools?: string[]
+          }
```
peppescg: you can reuse the above type

renderer/src/features/chat/components/model-selector.tsx
```diff
@@ -35,17 +35,51 @@ export function ModelSelector({
...
+              provider: providerId,
+              model: modelId,
+              apiKey:
+                'apiKey' in providerSettings ? providerSettings.apiKey : '',
+              enabledTools: settings.enabledTools,
+            }
+
+      onSettingsChange(newSettings)
+    } catch (error) {
+      console.error('Failed to load provider settings:', error)
```
peppescg: log.error

main/src/chat/constants.ts
samuv: it should be defined as output:

```
CHAT_PROVIDER_INFO: ChatProviderInfo[]
```

main/src/chat/types.ts
samuv: it makes sense

preload/src/preload.ts
samuv: yep

main/src/chat/streaming.ts
samuv: yeah let me revisit that nested ternary is not that great

main/src/chat/constants.ts
peppescg: sorry it is an array it should be 
```
 as const satisfies readonly ChatProviderInfo[];
```
in this way you are enforcing all the singles nested prop
and after that remove the type annotation

peppescg: now when you do for instance
```
const test = CHAT_PROVIDER_INFO[0]?.id
```
test will be automatically inferred as 'openai' as only value

main/src/chat/settings-storage.ts
samuv: yep, I have centralized a little this logic

main/src/chat/streaming.ts
samuv: moved to a dedicated function

main/src/chat/constants.ts
samuv: why not, the only thing is not readonly because it mutate when we have the models for openrouter, ollama and lmstudio, but I added the as const ...

peppescg (approved):
nice job!
🚢

PR #1257, opened 2025-11-12
chore: handle toolhive upgrade using renovate

peppescg:
Renovate steps:
- upgrade thv constant
- call `pnpm:thv` that removes `/bin` folder and install the new thv binary
- call `pnpm:generate-client` that regenerates the api clients


renovate.json
```diff
@@ -11,6 +11,17 @@
...
       "prPriority": 10,
       "schedule": []
+    },
+    {
+      "matchDatasources": ["github-releases"],
+      "matchPackageNames": ["stacklok/toolhive"],
+      "prPriority": 10,
+      "schedule": [],
+      "postUpgradeTasks": {
+        "commands": ["pnpm run thv", "pnpm run generate-client"],
```
kantord, line 21: we can see the output of these commands in some log files, right? 

if that is the case, I would use


```suggestion
        "commands": ["pnpm run thv version", "pnpm run generate-client"],
```

just because that way we will see what version the binary reports in the logs

peppescg: we don't need to know the version cause we already manually upgrade it in the previous step...pnpm thv use the constant. And it is already printed inside `fetch-thv` script

kantord: makes sense

PR #1272, opened 2025-11-24
feat: show issuer url for dynamic client registration and oidc

peppescg:
Show the issuer url for remote mcp in these use cases:
- dynamic client registration 
- oidc

### DIFF
--- a/renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
+++ b/renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
@@ -37,14 +37,12 @@ const AUTH_FIELD_MATRIX = {
 type AuthFieldName =
   (typeof AUTH_FIELD_MATRIX)[keyof typeof AUTH_FIELD_MATRIX][number]
 
-const shouldShowField = (
-  authType: string | undefined,
-  fieldName: AuthFieldName
-) => {
-  if (!authType) return false
-  const fields = AUTH_FIELD_MATRIX[authType as keyof typeof AUTH_FIELD_MATRIX]
-  return (fields as readonly AuthFieldName[])?.includes(fieldName) ?? false
-}
+const shouldShowField =
+  (authType: string | undefined) => (fieldName: AuthFieldName) => {
+    if (!authType) return false
+    const fields = AUTH_FIELD_MATRIX[authType as keyof typeof AUTH_FIELD_MATRIX]
+    return (fields as readonly AuthFieldName[])?.includes(fieldName) ?? false
+  }
 
 export function FormFieldsAuth({
   authType,
@@ -53,9 +51,11 @@ export function FormFieldsAuth({
   authType: string | undefined
   form: UseFormReturn<FormSchemaRemoteMcp>
 }) {
+  const showField = shouldShowField(authType)
+
   return (
     <>
-      {shouldShowField(authType, 'callback_port') && (
+      {showField('callback_port') && (
         <FormField
           control={form.control}
           name="oauth_config.callback_port"
@@ -91,7 +91,7 @@ export function FormFieldsAuth({
         />
       )}
 
-      {shouldShowField(authType, 'issuer') && (
+      {showField('issuer') && (
         <FormField
           control={form.control}
           name="oauth_config.issuer"
@@ -121,7 +121,7 @@ export function FormFieldsAuth({
         />
       )}
 
-      {shouldShowField(authType, 'authorize_url') && (
+      {showField('authorize_url') && (
         <FormField
           control={form.control}
           name="oauth_config.authorize_url"
@@ -152,7 +152,7 @@ export function FormFieldsAuth({
         />
       )}
 
-      {shouldShowField(authType, 'token_url') && (
+      {showField('token_url') && (
         <FormField
           control={form.control}
           name="oauth_config.token_url"
@@ -183,7 +183,7 @@ export function FormFieldsAuth({
         />
       )}
 
-      {shouldShowField(authType, 'client_id') && (
+      {showField('client_id') && (
         <FormField
           control={form.control}
           name="oauth_config.client_id"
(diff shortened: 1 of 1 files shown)


renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
```diff
@@ -265,6 +265,95 @@ describe('DialogFormRemoteMcp', () => {
...
+        const clientIdInput = screen.getByPlaceholderText(
+          'e.g. 00000000-0000-0000-0000-000000000000'
+        )
+        await user.clear(clientIdInput)
+        await user.type(clientIdInput, 'oidc-client-id')
+      }
+
+      await user.click(screen.getByRole('button', { name: 'Install server' }))
+
+      await waitFor(() => {
```
kantord, line 327: i guess that we could use `recordRequests` in order to make an assertion about the "data" attribute on the actual request payload instead of mocking the mutation. that way probably the assertion can also be simplified (fewer "objectContaining" and such)

but it's more of a nitpick than anythign

renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
```diff
@@ -185,7 +185,7 @@ export function FormFieldsAuth({
         )}
       />
 
-      {authType === 'oidc' && (
+      {authType !== 'oauth2' && (
```
kantord: correct me if I'm wrong, but my understanding is that what is going on here is that different auth types require different fields 🤔 

so basically there is a boolean flag matrix of auth types vs. form fields.

this might be slightly more verbose, but what do you think about adding a slightly more self-documenting const that describes this configuration:

```tsx
  const AUTH_FIELD_MATRIX = {
    none: [],
    'oidc-dynamic': ['issuer', 'client_credentials', 'pkce'],
    oidc: ['issuer', 'client_credentials', 'pkce'],
    oauth2: ['oauth2_urls', 'client_credentials', 'pkce'],
  } as const
```

then it will be more clean at first glance what is the mapping of auth_type -> field plus the usage would also be more self explanatory:


```tsx
const visible_extra_fields = AUTH_FIELD_MATRIX[auth_type];
...
visible_extra_fields[auth_type].includes('client_credentials') && (
  <>
    ...etc...
  </>
) 
```

renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
peppescg: that was my idea at the origin, I don't remember if we didn't have it at that time, or if it was a mess to test using it, because there are a bunch of stuff in the middle. I will take a look

renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
peppescg: yep I agreed. 
I had a similar map for schema validation and it makes sense to refactor it for form fields as well


```
const OAUTH_VALIDATION_RULES = {
  oauth2: [
    {
      field: 'authorize_url',
      message: 'Authorize URL is required for OAuth 2.0',
      path: ['oauth_config', 'authorize_url'],
    },
    {
      field: 'token_url',
      message: 'Token URL is required for OAuth2',
      path: ['oauth_config', 'token_url'],
    },
    {
      field: 'client_id',
      message: 'Client ID is required for OAuth 2.0',
      path: ['oauth_config', 'client_id'],
    },
  ],
  oidc: [
    {
      field: 'issuer',
      message: 'Issuer URL is required for OIDC',
      path: ['oauth_config', 'issuer'],
    },
    {
      field: 'client_id',
      message: 'Client ID is required for OIDC',
      path: ['oauth_config', 'client_id'],
    },
  ],
}
```

kantord: great

renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
kantord: in this case it is good to list complaints about it, because the current solution is extremely minimalistic and we can definitely improve it.

peppescg: refactored it using the record request for the entire file

renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
peppescg: I completed the refactor, definitely better

renderer/src/features/mcp-servers/components/remote-mcp/form-fields-auth.tsx
```diff
@@ -13,306 +13,337 @@ import type { FormSchemaRemoteMcp } from '@/common/lib/workloads/remote/form-sch
...
+    'client_secret',
+    'scopes',
+    'use_pkce',
+  ],
+} as const
+
+type AuthFieldName =
+  (typeof AUTH_FIELD_MATRIX)[keyof typeof AUTH_FIELD_MATRIX][number]
+
+const shouldShowField = (
```
kantord: just one nitpick: if you turn this into a closure inside FormFieldsAuth, you can avoid having to pass authType, making usage simpler and reducing the chance for any mistake.


an alternative is to turn  it into a hook or a curried function, something like this:
 

```tsx
const useShouldShowField = (authType?: string) => {
  const shouldShowField = (fieldName: AuthFieldName) => {
    // ... logic ...
  }

  return { shouldShowField }
}
```

peppescg: makes sense currying it

PR #1314, opened 2025-11-28
fix: sentry deps vulnerabilities

peppescg:
Fix @sentry deps using overrides, cause `@sentry/electron` is not still including the fixed version


peppescg:
this pr needs `@sentry/core` to be bumped, so I need to close it and wait for @sentry/electron upgrade

PR #1340, opened 2025-12-04
test: restore playwright tests

kantord:
- restore playwright
- update existing tests to match the new behavior
- change test isolation method: use a special group that is recreated from scratch in the app fixture. this way we get pretty good test isolation in a way that is connected to the app fixture which guarantees that you cannot interact with the app in a non-isolated way accidentally


kantord:
@claude can you please do a deep review of this PR

kantord:
@claude can you please do another deep review of the pr with the new changes included?

PR #1407, opened 2025-12-17
fix: registry misconfiguration route error handling

peppescg:
Customizing registry server, introducing a misconfiguration in the url, can lead an error of the application that it is not possible to fix rather than close the loop. So it is needed an error handling in the registry route

https://github.com/user-attachments/assets/cd0d7350-2d1c-4946-a010-2da82f61d59e

### DIFF
--- a/renderer/src/routes/settings.route.tsx
+++ b/renderer/src/routes/settings.route.tsx
@@ -6,10 +6,6 @@ import { Route } from './settings'
 export default function SettingsRouteComponent() {
   const { tab } = useSearch({ from: Route.id })
 
-  if (typeof window === 'undefined' || !window.electronAPI) {
-    return null
-  }
-
   return (
     <>
       <TitlePage title="Settings" />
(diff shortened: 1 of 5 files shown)


renderer/src/routes/settings.route.tsx
```diff
@@ -0,0 +1,19 @@
+import { TitlePage } from '@/common/components/title-page'
+import { useSearch } from '@tanstack/react-router'
+import { SettingsTabs } from '@/common/components/settings/tabs/settings-tabs'
+import { Route } from './settings'
+
+export default function SettingsRouteComponent() {
+  const { tab } = useSearch({ from: Route.id })
+
+  if (typeof window === 'undefined' || !window.electronAPI) {
+    return null
```
kantord: hmm, this seems weird... my understanding is that under some circumstances this would lead to an an empty page basically?

I understand that those circumstances should not really happen in real life, but it is still weird to show an empty page for an impossible scenario 🤔  why not throw an error for example?

at least I'm missing a code comment explaining what this logic is intended to solve because it's not obvious

peppescg: I don't know it was there from the beginning, so I avoided to touch it, but IMO it shouldn't be needed 🤔

peppescg: let me remove it

kantord (approved):

PR #1438, opened 2026-01-07
fix: switch coverage provider from istanbul to v8 for coveralls

peppescg:
## Summary

Fixes Coveralls integration by switching coverage provider from `istanbul` to `v8`.

## Changes

- Replace `@vitest/coverage-istanbul` with `@vitest/coverage-v8`
- Update `vitest.config.ts` to use `v8` provider
- Add `reportsDirectory: './coverage'` for explicit output path

## Why

The `istanbul` provider was generating empty `lcov.info` files, causing Coveralls to report "Nothing to report".


peppescg:
@claude could you review

vitest.config.ts
```diff
@@ -29,30 +29,35 @@ export default defineConfig({
...
+        'main/src/main-window.ts',
+        'main/src/system-tray.ts',
+        'main/src/menu.ts',
+        'main/src/dock-utils.ts',
+        'main/src/app-state.ts',
+        'main/src/container-engine.ts',
+        'main/src/csp.ts',
+        'main/src/logger.ts',
+        'main/src/headers.ts',
+        'main/src/util.ts',
```
samuv: why we are excluding this from the coverage?

peppescg: ops I guess opus excluded too much files, I wanted to exclude only logger cps and headers. Let me fix it

PR #1441, opened 2026-01-08
fix: store fetched models after ollama refresh

kantord:
adds proper end-to-end testing to the playground feature using ollama. This copies the same solution that was [implemented in the cloud-ui](https://github.com/stacklok/toolhive-cloud-ui/pull/199). 

In other words, the playground test actually adds an ollama provider and uses a very small, but real model. 

Note: the test assertion is not very useful/strict right now: just asks the LLM to reply with a specific string. The plan is to replace that assertion in the next PR with something slightly more complex: a secret code retrieved from a dummy (remote) MCP. That will also achieve testing the tool call capability and remote MCPs.

the PR is mostly about adding the end to end test, but there was a bug with auto-selecting the ollama model on a fresh install (end to end test running in GHA) so it resulted in a buggy state that can only be reproduced in GHA (or presumable a fully fresh install). Fixed this, and for this fix, I also added some vitest tests, because of the fact that it's only possible to reliably repro it in GHA.

### DIFF
--- a/e2e-tests/playground.spec.ts
+++ b/e2e-tests/playground.spec.ts
@@ -31,7 +31,6 @@ async function waitForPlaygroundReady(window: Page): Promise<void> {
 async function openProviderSettingsDialog(window: Page): Promise<void> {
   await waitForPlaygroundReady(window)
 
-  // Two paths: "Configure your providers" button (no provider) or model selector menu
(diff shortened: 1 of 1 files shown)


kantord:
@claude please review this branch

PR #1443, opened 2026-01-08
ci: trigger client regeneration on @hey-api/openapi-ts updates

samuv:
Extends the Renovate post-upgrade workflow to automatically regenerate the API client when `@hey-api/openapi-ts` is updated.

### Changes
- Add `package.json` to workflow path triggers
- Check PR diff for `@hey-api/openapi-ts` changes before running regeneration
- Skip all setup steps if neither toolhive nor hey-api was updated
- Add `@hey-api/openapi-ts` to Renovate package rules for priority scheduling

This ensures the generated client stays in sync with the code generator version without requiring manual intervention.

### DIFF
--- a/.github/workflows/renovate-post-upgrade.yml
+++ b/.github/workflows/renovate-post-upgrade.yml
@@ -5,6 +5,7 @@ on:
     branches: [main]
     paths:
       - 'utils/constants.ts'
+      - 'package.json'
 
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


samuv:
Incorporate some feedback from copilot:

- Error handling: || exit 1 after gh pr diff - fails workflow if diff can't be retrieved
- Robust matching: grep -q instead of bash pattern matching - more portable

PR #1447, opened 2026-01-09
test(e2e): test remote servers using playground

kantord:
this replaces the local fetch tool call with a "remote" server running locally in the tool call test. This way, the existing playground test case can kill 2 birds with one stone: it tests not only that tool calls work, but also provides an end-to-end test of the remote servers.

Limitation worth noting: the auth flow is not being tested, because there does not seem to be a way to test it that is both fairly simple to implement and adds major value.


kantord:
@claude please do a deep review of this PR

kantord:
@claude can you please do a deep review of this branch?

PR #1455, opened 2026-01-13
feat(remote-mcp): support bearer token authentication

peppescg:
Support `bearer token` authentication in the remote mcp form.
Refactor the remote form fields between custom and registry into a common folder, removing duplicated code.
Hide `proxy_mode` if `bearer token` auth type is selected.

https://github.com/user-attachments/assets/a0418206-ab5d-4fc2-8cc5-58b8ca0bc8fe

**IMPORTANT**
For testing it you need to use the toolhive binary from main repository. I used the mcp from https://us.posthog.com/  creating a bearer token for testing the flow.
Needs a toolhive release before release the UI as well, there is a fix not release yet from toolhive binary

### DIFF
--- a/renderer/src/features/mcp-servers/hooks/use-run-remote-server.tsx
+++ b/renderer/src/features/mcp-servers/hooks/use-run-remote-server.tsx
@@ -20,6 +20,7 @@ interface UseRunRemoteServerProps {
   quietly?: boolean
 }
 
+/** Build the appropriate secrets based on auth type. */
 const buildAuthSecret = (data: FormSchemaRemoteMcp) => {
   const isDefaultAuthType = data.auth_type === REMOTE_MCP_AUTH_TYPES.None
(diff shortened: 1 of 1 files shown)


renderer/src/features/mcp-servers/hooks/use-run-remote-server.tsx
```diff
@@ -19,6 +20,18 @@ interface UseRunRemoteServerProps {
   quietly?: boolean
 }
 
+const getSecrets = (data: FormSchemaRemoteMcp) => {
```
kantord: the naming is slightly confusing as now we have `handleSecrets` and `getSecrets`, both slightly "generic" names.

peppescg: replaced with `buildAuthSecret`

kantord (approved):

PR #1465, opened 2026-01-15
fix(proxy_mode): show proxy mode only for stdio transport for local mcp

peppescg:
The `proxy_mode` works only with transport `stdio`, following this we are going to remove the `proxy_mode` value from the remote create/update payload, and sending it only in case of matching transport

https://github.com/user-attachments/assets/ef77184d-be1a-4c77-ba4d-3a5ea7d28397


samuv (approved):
Looks good!

PR #1467, opened 2026-01-15
test: fail tests when accidentally sending sentry events

kantord:
This code makes sure that it is not possible to generate sentry events form end-to-end tests. 

We are already not sending sentry events in end-to-end tests unless the end-to-end tests are run against production builds, so this problem was already rare and unlikely.

This PR just adds a final warning: if you accidentally run the tests and send Sentry events, you will get tests failures with an error explaining this. Under normal circumstances, the events will still be associated with your local user ID, so will still be filtered out anyway.

Running end-to-end tests against production builds remains possible as long as you build it with an empty Sentry DSN

### DIFF
--- a/renderer/src/common/hooks/__tests__/use-cleanup-meta-optimizer.test.ts
+++ b/renderer/src/common/hooks/__tests__/use-cleanup-meta-optimizer.test.ts
@@ -15,9 +15,8 @@ vi.mock('../use-feature-flag')
 vi.mock('@/features/mcp-servers/hooks/use-mutation-delete-group')
 
 const { useFeatureFlag } = await import('../use-feature-flag')
-const { useMutationDeleteGroup } = await import(
-  '@/features/mcp-servers/hooks/use-mutation-delete-group'
-)
+const { useMutationDeleteGroup } =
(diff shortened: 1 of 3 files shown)


kantord:
@claude can you please do a deep review of this branch please

e2e-tests/fixtures/electron.ts
```diff
@@ -109,6 +113,27 @@ export const test = base.extend<ElectronFixtures>({
...
+      try {
+        const Sentry = (await import('@sentry/electron/main')) as {
+          getCurrentHub?: () => {
+            getClient?: () => { getOptions?: () => { dsn?: string | null } }
+          }
+        }
+        const client = Sentry.getCurrentHub?.().getClient?.()
+        return client?.getOptions?.().dsn ?? null
+      } catch {
+        return null
```
peppescg, line 130: I usually don't like this dynamic import that AI does sometime, you can import normally at the top?

kantord: the thing is that it is being passed to the electron process so that we can look into what is inside the electron process. importing it at top level would mean it is importing sentry into the test runner, not the app

PR #1482, opened 2026-01-20
fix(registry-api): show error in case registry-api unavailable

peppescg:
### Problem

  After changing registry settings to an invalid value, we are not showing the error in place in the settings. The form also briefly flashed to "Default Registry" during submission.

  **Changes**

  1. Show errors after registry api url is changed
  2. Fresh data on navigation - Changed the registry route to always fetch fresh data instead of using
  cached values
  3. Clear servers cache - When registry settings change, the servers cache is now cleared to force a fresh
   fetch
  4. Consistent cache operations - Fixed cache updates to use correct type mappings, preventing UI flicker
  5. Proper error display - Errors from the API now properly trigger the error component

_____

https://github.com/user-attachments/assets/0e8639a4-a41a-465b-99dc-034dbc123af9

### DIFF
--- a/renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
+++ b/renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
@@ -530,14 +530,11 @@ describe('RegistryTab', () => {
       expect(screen.getByLabelText(/Registry Server API URL/i)).toBeVisible()
     })
 
-    // Enter API URL
     const apiUrlInput = screen.getByLabelText(/Registry Server API URL/i)
     await userEvent.type(apiUrlInput, apiUrl)
 
-    // Submit
     await userEvent.click(screen.getByRole('button', { name: 'Save' }))
 
-    // After mutation, form should still show the submitted values (not flash to default)
     await waitFor(() => {
       expect(screen.getByLabelText(/Registry Server API URL/i)).toHaveValue(
         apiUrl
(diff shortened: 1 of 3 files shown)


peppescg:
@claude succinct review

renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
```diff
@@ -448,4 +450,103 @@ describe('RegistryTab', () => {
...
+    await userEvent.click(screen.getByRole('combobox'))
+    await userEvent.click(
+      screen.getByRole('option', { name: 'Registry Server API' })
+    )
+
+    await waitFor(() => {
+      expect(screen.getByLabelText(/Registry Server API URL/i)).toBeVisible()
+    })
+
+    // Enter API URL
```
kantord: nit pick, but such comments basically just say what the code already says (not all the comments here are redundant though, "form should still show the submitted values" one is important for example because it adds context)

samuv:
Is it correct that validation is triggered immediately when you select something different from the registry?
Screenshot 2026-01-21 at 10 56 32

renderer/src/common/components/settings/tabs/__tests__/registry-tab.test.tsx
peppescg: agreed. LLM sometimes add no sense comments

peppescg:
> Is it correct that validation is triggered immediately when you select something different from the registry? Screenshot 2026-01-21 at 10 56 32


This was at the beginning, but I would prefer to show error message on submit. Let me replace the validation

PR #1486, opened 2026-01-20
test: refactor mock overrides to use new wrapper (third batch)

kantord:
Part of a series of PRs to refactor all tests to use the new mocking system:

- https://github.com/stacklok/toolhive-studio/pull/1474
- https://github.com/stacklok/toolhive-studio/pull/1476

related to this task: https://github.com/stacklok/toolhive-studio/issues/1343

It includes some changes to agent skills as well because I'm using the refactoring PRs in order to also progressively improve the quality of the agent skills.

PR #1474, opened 2026-01-19
test: refactor mock overrides to use new wrapper (first batch)

kantord:

Issue #1343, opened 2025-12-09
[Task] Replace tests that are mocking api or functions


peppescg:
### Description

- Remove unneeded mock per test (centralized the mock in `vitest.setup.ts` )
- Replace API mock with recordRequest
- create conveinient way to override the msw response

### Priority

Medium

### Task Type

Feature

PR #1497, opened 2026-01-21
test: refactor mock overrides to use new wrapper (fourth batch)

kantord:
Part of a series of PRs to refactor all tests to use the new mocking system:

- https://github.com/stacklok/toolhive-studio/pull/1474
- https://github.com/stacklok/toolhive-studio/pull/1476
- https://github.com/stacklok/toolhive-studio/pull/1486

related to this task: https://github.com/stacklok/toolhive-studio/issues/1343

It includes some changes to agent skills as well because I'm using the refactoring PRs in order to also progressively improve the quality of the agent skills.

### DIFF
--- a/renderer/src/common/mocks/customHandlers/index.ts
+++ b/renderer/src/common/mocks/customHandlers/index.ts
@@ -71,30 +71,6 @@ export const customHandlers = [
     }
   ),
 
-  http.get(mswEndpoint('/api/v1beta/workloads/:name/export'), ({ params }) => {
-    const { name } = params
-
-    const server = getWorkloadByName(name as string)
-    if (!server) {
-      return HttpResponse.json({ error: 'Server not found' }, { status: 404 })
-    }
-
-    return HttpResponse.json({
-      name: server.name,
-      image: server.package || 'ghcr.io/default/default:latest',
-      transport: 'stdio',
-      target_port: server.port || 0,
-      cmd_args: [],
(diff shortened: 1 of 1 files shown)


kantord:
@claude can you please do a deep review of this PR?

renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
```diff
@@ -73,7 +73,7 @@ describe('ManageClientsButton – BDD flows', () => {
       ],
     }))
 
-    mockedPostApiV1BetaClientsRegister.override(() => [])
+    mockedPostApiV1BetaClientsRegister.activateScenario('empty')
```
peppescg, line 76: Is the `empty` param necessary? it seems repeated but without a custom logic per test, so I was wondering maybe we can get rid of it?

peppescg (approved):

renderer/src/features/clients/components/__tests__/manage-clients-button.test.tsx
kantord: i mean, in this case it is not as obvious why it needs to be an "empty" scenario rather than a hardcoded return value because the hardcoded value is just an array, but in general the empty response is still more complex, like an object with different attributes and such.

so I guess we should make a decision not on this example but overall in the project :thinking:

PR #1518, opened 2026-01-23
feat: upgrade Streamdown and adjust minimum version of tanstack router

samuv:
as per title bumping Streamdown to the new major and move to the new plugin system

adding mermaid
Screenshot 2026-01-23 at 17 47 29

Screenshot 2026-01-23 at 17 46 21

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -124,10 +124,10 @@ importers:
         specifier: ^5.80.5
         version: 5.91.2(@tanstack/react-query@5.90.19(react@19.2.3))(react@19.2.3)
       '@tanstack/react-router':
-        specifier: '^1.154.12 '
+        specifier: ^1.154.12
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -123,11 +123,14 @@
...
+    "@streamdown/cjk": "^1.0.1",
+    "@streamdown/code": "^1.0.1",
+    "@streamdown/mermaid": "^1.0.1",
     "@tailwindcss/vite": "^4.1.8",
     "@tanstack/react-query": "^5.77.2",
     "@tanstack/react-query-devtools": "^5.80.5",
-    "@tanstack/react-router": "^1.120.11",
-    "@tanstack/react-router-devtools": "^1.120.11",
+    "@tanstack/react-router": "^1.154.12",
+    "@tanstack/react-router-devtools": "^1.154.12",
```
samuv, line 133: we already have the bump, but I added the minimum version here because of the bug that had

PR #1531, opened 2026-01-29
refactor(api): move api folder to common/api and update configs

samuv:
As per the title, this moves the `api` folder from the root into `common/` and updates the paths and aliases across the codebase.


peppescg (approved):
🚀

PR #1547, opened 2026-02-03
feat: show if current version is latest one

peppescg:
Align version and cli tabs on new rebranding style

Screenshot 2026-02-16 at 11 11 29Screenshot 2026-02-16 at 11 29 46
Screenshot 2026-02-16 at 11 29 56

### DIFF
--- a/renderer/src/common/components/settings/tabs/cli-tab.tsx
+++ b/renderer/src/common/components/settings/tabs/cli-tab.tsx
@@ -17,6 +17,7 @@ import { WrapperField } from './components/wrapper-field'
 import { Badge } from '../../ui/badge'
 import { Button } from '../../ui/button'
 import { Alert, AlertDescription, AlertTitle } from '../../ui/alert'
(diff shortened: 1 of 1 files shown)


samuv:
Instead of the text, wouldn’t it be better to use a green tag on the side that just says “latest”?  
I feel like the text in parentheses doesn’t have a lot of impact

peppescg:
> Instead of the text, wouldn’t it be better to use a green tag on the side that just says “latest”? I feel like the text in parentheses doesn’t have a lot of impact

yeah you're right I don't have a strong opinion.
I also forgot to share this PR 😆 cause I am not totally convinced about it (probably I would move the default update flow to manual update, cause also if the user do not wants to upgrade, the next time that will restart the app he will find the new one, so he doesn't have the full control...and moving the default flow to manual there will be always the notification and icon for new update available, and in this case you shouldn't need a latest label)

@jtenniswood we can add the a green tag on the right side with latest, wdyt?

PR #1574, opened 2026-02-06
build(publisher): add s3 as a publisher

samuv:
Add S3 as a secondary publish target and configure AWS OIDC authentication in the release workflow so that release artifacts are uploaded to the `toolhive-studio-releases` S3 bucket in addition to GitHub Releases

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -35,6 +35,7 @@ export default {
   ignoreDependencies: [
     '@electron-forge/maker-dmg', // Used indirectly in MakerDMGWithArch
     '@electron-forge/publisher-github',
+    '@electron-forge/publisher-s3',
     '@electron-forge/maker-base',
     '@tailwindcss/vite',
   ],


.github/workflows/on-release.yml
```diff
@@ -107,7 +108,14 @@ jobs:
         uses: ./.github/actions/setup-flatpak
         if: runner.os == 'Linux'
 
-      - name: Build and Publish to GitHub Release
+      - name: Configure AWS Credentials
+        if: vars.AWS_ROLE_ARN != ''
```
peppescg, line 112: shouldn't we block the `build and publish` job if we don't have `AWS_ROLE_ARN` ?

samuv: at the moment, if publishing to S3 fails, it’s not a blocker. I’d prefer to proceed with the release even if the upload fails

PR #1581, opened 2026-02-06
feat(mcp-servers): add one-click update for servers with registry version drift

samuv:
- Detects when a container-based MCP server is running an older image tag than what the registry lists and surfaces an update action in two places: an amber icon on the server card and an "Update to {tag}" item in the dropdown menu
- Extracts shared upgrade logic into a `useUpdateVersion` hook that confirms with the user, then sends the full current workload config with only the image replaced
- Refactors `useIsServerFromRegistry` to expose `matchedRegistryItem`, removing the redundant `useServerDetails` hook and its separate API call
- Deduplicates `useIsServerFromRegistry` usage per card by lifting it to `CardMcpServer` and passing results down via props

https://github.com/user-attachments/assets/86f69b29-f050-42d0-a7b1-ff2d56d36ee1

### DIFF
--- a/renderer/src/features/mcp-servers/components/card-mcp-server/index.tsx
+++ b/renderer/src/features/mcp-servers/components/card-mcp-server/index.tsx
@@ -85,6 +85,8 @@ function CardContentMcpServer({
   registryImage,
 }: CardContentMcpServerProps) {
   const isRunning = status === 'running'
+  const isUpdating = `${status}` === 'updating'
+  const isDeleting = `${status}` === 'deleting' || status === 'removing'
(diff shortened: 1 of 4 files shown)


JAORMX:
I think this is awesome. Kinda wanted to do it for the CLI / API too

jtenniswood:
Nice. does it need the icon when it's updating? Seems a bit redundant.

samuv:
> Nice. does it need the icon when it's updating? Seems a bit redundant.

I disabled the button icon during updates because the edit action it could fail. We could also hide it entirely, although the status in the update flow is still inconsistent (see this [issue](https://github.com/stacklok/toolhive/issues/2350)).

Issue #1076, opened 2025-10-17
[Task] Spike for validate deep-link architecture


samuv:
### Description

We want to support deep-linking that allows users to navigate directly to a specific page or trigger an action inside the desktop application.

Deep linking is expected to integrate cleanly with the Electron main process lifecycle and the existing TanStack Router-based navigation in the renderer.

The research team has already created a [POC](https://github.com/stacklok/toolhive-studio/issues/956) and a [Story](https://github.com/stacklok/toolhive-studio/issues/956).

With this task, we want to analyze implementation options and validate stability, extensibility, and error-handling for deep-link routing across cold and warm application starts.

**Scope note:** This task focuses on deep-link support in the desktop app only. Future CLI or Cloud UI integrations are out of scope, but the architecture should remain extensible enough to support additional link handlers over time.

---

### Acceptance Criteria

- [x] Architecture is **extensible** and **composable**, allowing new deep-link handlers to be added without modifying core dispatch logic.
- [x] Works seamlessly with **TanStack Router** for route resolution and navigation in the renderer.
- [x] Supports both **cold start** (app closed) and **warm start** (app already running) deep-link handling.
- [x] Implements OS-level protocol handling correctly:
  - [x] Registers the deep-link scheme via `app.setAsDefaultProtocolClient`
  - [x] Handles macOS deep links via `app.on('open-url')`
  - [x] Handles Windows/Linux deep links via `second-instance` + `process.argv` parsing
- [x] Deep-link events are safely queued until the renderer is ready (leveraging existing window readiness patterns such as `pollWindowReady`).
- [x] Handles **404** and **invalid links gracefully**, with clear user-facing recovery options.
- [x] Deep-link scheme and parsing logic are **documented**, **type-safe**, and validated.
- [ ] Architecture supports **multiple link patterns** (e.g., navigation routes vs action-style targets).
- [ ] Proper **error logging and telemetry** for broken or invalid links:
  - [ ] Structured logging via existing logging utilities
  - [ ] Sentry capture with relevant context (link payload, failure reason)
- [x] Includes at least one **proof-of-concept** demonstrating navigation from an external deep link to a TanStack route.
- [x] Design is **testable** and covered by minimal **unit and integration tests** (main + renderer).

---

### Priority

Medium

---

### Task Type

Feature

---

### Additional Context

Initial codebase research confirms:

- Single-instance mode is already implemented (`app.requestSingleInstanceLock`), but does not yet parse deep-link URLs.
[...]

PR #1584, opened 2026-02-09
feat: add initial deep link implementation

kantord:
fixes https://github.com/stacklok/toolhive-studio/issues/1076 

Technical document: https://docs.google.com/document/d/1Y-tYsIl98DPHKdCNsbuvtZeb0VUq41cTxQfGQElbS9c

Add deep link support via the toolhive-gui:// protocol. URLs like toolhive-gui://v1/open-registry-server-detail?serverName=fetch open the app   
(or focus it if already running) and navigate to the corresponding page.

For testing on MacOS, you probably have to make a production build and install it. There are probably some hacky ways to make such testing builds more automated, but I could not test that.

For Linux, a helper script was added: this script makes a production build and installs its .desktop file in such a way that deep links are immediately opened using that link.

You can use this js fiddle to find some links for testing: https://jsfiddle.net/deLzqvos/1/  (github markdown filters out custom protocols, so cannot be directly included in the PR description)

Still TODO:

- some testing of the "readiness" system to make sure it works reliably
- testing on Windows and Mac - need help with this
- testing Flatpak builds - I wanna get feedback on this branch before investing time into that

### DIFF
--- a/renderer/src/renderer.tsx
+++ b/renderer/src/renderer.tsx
@@ -115,8 +115,6 @@ if (!window.electronAPI || !window.electronAPI.getToolhivePort) {
   // Listen for deep link navigation events — the main process resolves
   // the deep link URL to a navigation target, so the renderer just navigates.
   const deepLinkCleanup = window.electronAPI.onDeepLinkNavigation((target) => {
-    // TODO: Remove this alert — temporary debug aid for Windows testing
-    alert(`[deep-link] target: ${target.to} ${JSON.stringify(target.params)}`)
     log.info(`[deep-link] Navigating to: ${target.to}`, target.params)
     router.navigate(target)
   })


kantord:
@claude deep review please

kantord:
@claude please do a deep review

kantord:
@claude please do a deep review

peppescg:
@kantord clicking on jsfiddle the `serverName=non-existent"` I got the general error page in the desktop app. I would expect the 404 maybe 🤔 , the http route in fact is 404, it should be a specific 404 registry page IMO

Screenshot 2026-02-10 at 10 35 13


Screenshot 2026-02-10 at 10 32 04

peppescg:
@kantord did you try with a build and the app closed?  
Creating a build in this way we can test production build and in Windows as well.
You need to rebase before the /build-test 😆  ( I would avoid to do it myself)

common/deep-links.ts
```diff
@@ -0,0 +1,92 @@
+// Deep link definitions — see design doc:
+// https://docs.google.com/document/d/1Y-tYsIl98DPHKdCNsbuvtZeb0VUq41cTxQfGQElbS9c
```
peppescg, line 2: I would remove this is an internal document, if oyu want to share technical decisione it would be better to have a rfc

main/src/deep-links/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  waitForMainWindowReady,
+} from '../main-window'
+import { parseDeepLinkUrl } from './parse'
+import {
+  DEEP_LINK_PROTOCOL,
+  showNotFound,
+  resolveDeepLinkTarget,
+} from '@common/deep-links'
+
+export { registerProtocolWithSquirrel } from './squirrel'
```
peppescg, line 14: it is already exported in the squirrel file
I don't know why sometimes the agent do this 😆

main/src/deep-links/squirrel.ts
```diff
@@ -0,0 +1,31 @@
...
+    const success = app.setAsDefaultProtocolClient(
+      DEEP_LINK_PROTOCOL,
+      updateExe,
+      ['--processStart', path.basename(process.execPath)]
+    )
+    log.info(
+      `[deep-link] Registered protocol via Squirrel Update.exe: ${success}`
+    )
+  } else {
+    const success = app.setAsDefaultProtocolClient(DEEP_LINK_PROTOCOL)
```
peppescg, line 37: this should be for mac right? if so I would rename the fn

main/src/main.ts
```diff
@@ -217,12 +222,24 @@ export async function blockQuit(source: string, event?: Electron.Event) {
 const gotTheLock = app.requestSingleInstanceLock()
 
 if (!gotTheLock) {
+  // This is the second instance (e.g. launched by a deep link or duplicate click).
+  // The first instance handles the deep link via the 'second-instance' event.
+  // We use process.exit(0) instead of app.quit() because quit() is async —
+  // its lifecycle events let app.whenReady() fire and briefly create a window.
   log.info('Another instance is already running. Exiting...')
-  app.quit()
+  process.exit(0)
```
peppescg, line 230: 🤔 I am not sure we need to do `process.exit(0)` also if `app.quit` is async I don't think it is an issue

kantord: on Mac, it's not strictly needed, because the "Single instance" workflow is enforced by the protocol handler system.

On Windows and Linux, a second instance is created. This second instance does not needs to be killed but does need a proper teardown (and actually, for whatever reason, a zombie window is created during the teardown process)

Perhaps I could add a comment that this is not needed on Mac, but removing it will cause issues on Linux and Windows

common/deep-links.ts
kantord: yes, I do want to share it. actually I'm thinking about [ADR](https://adr.github.io/) instead of RFC. 

I could convert the doc into an ADR markdown file and add it to this repo. That way it can also be referenced easily by agents

main/src/main.ts
peppescg: ah ok, you already tested the flow in Windows? I thought you didn't test there yet

kantord: nope, not yet tested on Windows, but this just follows from the information I found while researching (except for the zombie Window part, that did surprise me when testing on Linux, but it might either just be how it works, or maybe we actually create the window somewhere)

peppescg (approved):

PR #1591, opened 2026-02-10
docs: add design document about deep-links

kantord:
this removes references to a private document from the codebase. The full document (with all details, about 30 pages) will still probably be published later, but probably for the repo this is enough


peppescg (approved):
Why a ADR inside the docs?

PR #1604, opened 2026-02-11
feat(workload): improve update flow and detect envs drift

samuv:
With this we improving the upgrade flow
- Detect added/removed env vars between local config and registry when updating server versions
- Show drift details (new/removed variables with badges) in the update dialog
- No drift: simple confirm for quick upgrade. Drift detected: route to edit dialog with missing env vars pre-filled

Fix the problem that we have in after a simple upgrade, we were missing the polling

https://github.com/user-attachments/assets/cf34718e-5275-4ff6-be40-28facf6689d1

### DIFF
--- a/renderer/src/features/mcp-servers/components/drift-details.tsx
+++ b/renderer/src/features/mcp-servers/components/drift-details.tsx
@@ -1,6 +1,18 @@
 import { Badge } from '@/common/components/ui/badge'
 import type { EnvVarDrift } from '../lib/get-env-vars-drift'
 
+function VarTypeBadge({ secret }: { secret: boolean }) {
+  return secret ? (
+    <Badge variant="outline" className="px-1 py-0 text-[10px]">
+      secret
+    </Badge>
+  ) : (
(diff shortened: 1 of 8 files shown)


renderer/src/features/mcp-servers/components/drift-details.tsx
```diff
@@ -0,0 +1,67 @@
...
+                <code className="bg-muted rounded px-1 py-0.5">{v.name}</code>
+                {v.secret ? (
+                  <Badge variant="outline" className="px-1 py-0 text-[10px]">
+                    secret
+                  </Badge>
+                ) : (
+                  <Badge variant="secondary" className="px-1 py-0 text-[10px]">
+                    env
+                  </Badge>
+                )}
```
peppescg: this section here seem to be repeated below in the removed ones, you can abstract into a component

samuv: yep, fixed

PR #1633, opened 2026-02-13
fix(polling): reconcile workload status for CLI-initiated operations

### LINKED ISSUE #1129: [Bug] reconcile workload status using polling in case of mcp stopped by CLI
### Bug Description

if we stop all the mcp ( or one mcp ) from CLI, the UI will use polling in order to update the status.
The same for restart, if we restart an mcp from CLI, we are going to see starting in the UI until a refresh of the page.

This is because we are not start polling based on status for workload list in the mcp servers page

____
Restart
We continue to see starting until we reload the page.
Image
____
Stopping
Some mcp needs more time to be stopped and there is an inconsistent status between CLI and UI

Image

### Steps to Reproduce

_No response_

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Priority

Medium

### Environment

_No response_

### Additional Context

_No response_

samuv:
- Add `useAutoResumePolling` hook that automatically polls servers stuck in transition statuses (`starting`, `stopping`, `restarting`) until they reach a stable state, fixing stale UI when operations are triggered via CLI
- Migrate all polling call sites to register in the TanStack Query cache via a shared `pollingQueryKey` for automatic deduplication between mutation-initiated and auto-resume polls
- Use `pollServerUntilStable` (polls until any non-transition status) instead of a fixed target, handling edit cycles where `stopping` leads to `running`

Resolves #1129

https://github.com/user-attachments/assets/7e909051-b83b-460d-9678-7f775efd418f

### DIFF
--- a/renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
+++ b/renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
@@ -116,7 +116,7 @@ export function useMutationRestartServerAtStartup() {
       // Poll until all servers are running (per-server for dedup granularity)
       await Promise.all(
         serverNames.map((serverName) =>
-          fetchPollingQuery(queryClient, serverName, () =>
+          fetchPollingQuery(queryClient, serverName, 'running', () =>
             pollServerStatus(
               () =>
                 queryClient.fetchQuery(
@@ -186,7 +186,7 @@ export function useMutationRestartServer({
     },
     onSuccess: async () => {
       // Poll until server running
-      await fetchPollingQuery(queryClient, name, () =>
+      await fetchPollingQuery(queryClient, name, 'running', () =>
         pollServerStatus(
           () =>
             queryClient.fetchQuery(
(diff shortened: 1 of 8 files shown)


renderer/src/features/mcp-servers/hooks/use-mutation-restart-server.ts
```diff
@@ -112,23 +112,24 @@ export function useMutationRestartServerAtStartup() {
...
+                () =>
+                  queryClient.fetchQuery(
+                    getApiV1BetaWorkloadsByNameStatusOptions({
+                      path: { name: serverName },
+                    })
+                  ),
+                'running'
+              ),
+            staleTime: 0,
+          })
```
peppescg: I guess you can abstract this piece of code that seems shared across the same file but also in the stop one with a different status ?

samuv: we can extrapole the wrapper

samuv:
@claude please do a concise review

PR #1661, opened 2026-02-17
feat(db): add SQLite persistence layer with incremental read migration via feature flags

samuv:
Introduces SQLite (`better-sqlite3`) as a secondary persistence layer alongside `electron-store`,
using a dual-write strategy to enable a safe, incremental migration.
### What's changed
**Core infrastructure**
- SQLite database setup with WAL mode, foreign keys, and a versioned migrator
- AES-256-GCM encryption for sensitive fields (API keys, endpoint URLs) stored as BLOBs
- Startup reconciliation to sync existing `electron-store` data into SQLite on first run
**Schema (9 tables)**
- `settings`, `feature_flags`, `ai_providers`, `selected_model`, `enabled_mcp_tools`,
  `threads`, `thread_messages`, `active_thread`, `shutdown_servers`
**Dual-write integration**
- All writes go to both stores; reads default to `electron-store`
- Integrated across: settings, feature flags, chat providers, threads, shutdown servers
**Incremental read rollout**
- Five per-domain feature flags (`sqlite_reads_*`) to switch reads to SQLite one domain at a time
- `sqlite_reads_*` flags always resolve from `electron-store` to avoid circular reads
**Tests**
- Schema integrity, migrations, encryption round-trips, all writer/reader pairs, and full reconciliation flow
## How to migrate reads
Enable a domain flag to switch reads to SQLite for that domain:
- `sqlite_reads_settings`
- `sqlite_reads_threads`
- `sqlite_reads_chat_settings`
- `sqlite_reads_feature_flags`
- `sqlite_reads_shutdown`

### DIFF
--- a/main/src/db/__tests__/reconcile-from-store.test.ts
+++ b/main/src/db/__tests__/reconcile-from-store.test.ts
@@ -157,7 +157,7 @@ describe('reconcileFromStore', () => {
     expect(tools[0]!.server_name).toBe('server1')
   })
 
-  it('syncs threads and skips unchanged threads based on timestamp', () => {
+  it('syncs all threads from electron-store to SQLite', () => {
     testDb
       .prepare(
         "INSERT INTO threads (id, title, created_at, last_edit_timestamp) VALUES ('t1', 'Old', 1000, 1000)"
--- a/main/src/db/reconcile-from-store.ts
+++ b/main/src/db/reconcile-from-store.ts
@@ -93,14 +93,10 @@ function syncThreads(db: Database.Database): void {
   const threads = threadsStore.get('threads')
   if (!threads || typeof threads !== 'object') return
 
-  // Get all timestamps from SQLite in one query
-  const dbTimestamps = new Map(
-    (
-      db.prepare('SELECT id, last_edit_timestamp FROM threads').all() as {
-        id: string
-        last_edit_timestamp: number
-      }[]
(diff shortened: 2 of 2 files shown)


main/src/db/encryption.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (fs.existsSync(keyPath)) {
+      cachedKey = fs.readFileSync(keyPath)
+      if (cachedKey.length === KEY_LENGTH) return cachedKey
+      log.warn('[DB] Encryption key file has wrong length, regenerating')
+    }
+  } catch {
+    log.warn('[DB] Could not read encryption key, generating new one')
+  }
+
+  cachedKey = crypto.randomBytes(KEY_LENGTH)
```
peppescg, line 30: I think it would be better to encrypt the db key 🤔?
We can leverage directly [safe-storage](https://www.electronjs.org/docs/latest/api/safe-storage ) that supports under the hood the OSs cryptography systems

samuv: yeah I started with that but I don’t want to prompt for the OS keychain password

peppescg: why? the user doesn't have to put the password every time

samuv: Only if they click “Always”

peppescg (approved):

PR #1694, opened 2026-03-05
feat(api): migrate thv API transport from HTTP to UNIX socket / Windows named pipe

samuv:
The `thv` binary previously served the studio API on a localhost TCP port assigned at startup, requiring CSP relaxation, a `127.0.0.1` listener anyone on the box could probe, and a `getToolhivePort()` round-trip in every renderer fetch. This PR moves the managed-thv transport to a UNIX domain socket on macOS/Linux and a Windows named pipe on Windows, with the renderer's hey-api `fetch` forwarded over Electron IPC to the main process — which dials the socket/pipe directly via Node's `http.request({ socketPath })`. The studio is now the only process that can talk to thv; no localhost port is opened for the managed instance.

## Summary

- **Transport** — Managed `thv` serves the API on a per-process UNIX socket (POSIX) or named pipe (Windows). Renderer `fetch` goes through [`ipc-fetch.ts`](renderer/src/common/lib/ipc-fetch.ts) → IPC → main [`unix-socket-fetch.ts`](main/src/unix-socket-fetch.ts) (`http.request({ socketPath })`). No localhost listener for the managed API; CSP [`csp.ts`](main/src/csp.ts) no longer whitelists `thv` in `connect-src`.
- **External `thv`** — `THV_PORT` → **`THV_SOCKET`**; optional **`THV_MCP_PORT`** (validated). `isToolhiveRunning()` is true in this mode so readiness (e.g. secret provider) works. Dev-only [`CustomSocketBanner`](renderer/src/common/components/custom-socket-banner.tsx) when using an external socket.
- **Main-process callers** — `mcp-tools`, `settings-storage`, and builtin `skills` use `createMainProcessApiClient()` instead of `http://localhost:${port}` (fixes `undefined` port / broken playground & skills).
- **Tracing** — Renderer adds `sentry-trace` / `baggage` / `traceparent` via `getTraceData({ propagateTraceparent: true })`. Main wraps the socket request in `Sentry.suppressTracing()` so headers aren’t overwritten.
- **Windows** — Pipes instead of temp-dir AF_UNIX (blocked on matching `thv` release; see linked PR).
- **E2E** — [`app-relaunch.ts`](e2e-tests/helpers/app-relaunch.ts) talks to `thv` over the real socket path.
- **Tests** — New/extended coverage for `ipc-fetch`, `unix-socket-fetch`, CSP, `graceful-exit` `createFetch` path, `ipc-handlers/toolhive`, plus existing `toolhive-manager` socket / `THV_SOCKET` cases.

## How to validate

1. **macOS / Linux** — `pnpm dev`. Confirm app boots, MCP list/registry/playground tool toggle work, no `connect-src` violations.
2. **Windows** — After matching `thv` ships: pipe path from `getToolhiveSocketPath()`, smoke the same flows.
3. **External `thv`** — `THV_SOCKET=… pnpm dev`: banner appears, APIs work, `isToolhiveRunning` true.
4. **Tracing** — With Sentry on studio + `thv`, confirm one trace id links renderer and backend spans.
5. **Tests** — `pnpm vitest run` (or CI) green for the new/updated test files above.

## Out of scope

- thv discovery file / `npipe://` cross-process checks on Windows (follow-up). The studio doesn’t pass `--nonce`, so it doesn’t write the discovery file.
- Bumping bundled `thv.exe` in `utils/constants.ts` until a tagged `thv` release includes the pipe listener.

### DIFF
--- a/e2e-tests/helpers/app-relaunch.ts
+++ b/e2e-tests/helpers/app-relaunch.ts
@@ -1,3 +1,4 @@
+import http from 'node:http'
 import path from 'path'
 import {
   _electron as electron,
@@ -29,7 +30,7 @@ function getExecutablePath(): string {
 export interface LaunchedApp {
   app: ElectronApplication
   window: Page
-  baseUrl: string
+  socketPath: string
   /**
    * Terminate the app without waiting on the renderer's before-quit teardown.
    *
@@ -74,18 +75,20 @@ export async function launchApp(userDataDir: string): Promise<LaunchedApp> {
 
   await window.getByRole('link', { name: /mcp servers/i }).waitFor()
 
-  const port = await window.evaluate(async () => {
+  const socketPath = await window.evaluate(async () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    return (await (globalThis as any).electronAPI.getToolhivePort()) as number
+    return (await (globalThis as any).electronAPI.getToolhiveSocketPath()) as
+      | string
+      | undefined
   })
 
-  if (!port) {
-    throw new Error('Failed to resolve ToolHive port from the launched app')
+  if (!socketPath) {
+    throw new Error(
+      'Failed to resolve ToolHive socket path from the launched app'
+    )
   }
 
-  const baseUrl = `http://127.0.0.1:${port}`
-
-  await waitForThvReady(baseUrl)
+  await waitForThvReady(socketPath)
 
   const close = async () => {
     // Force an immediate exit via Electron's app.exit(), bypassing before-quit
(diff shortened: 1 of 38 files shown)


main/src/toolhive-manager.ts
```diff
@@ -187,11 +227,12 @@ export async function startToolhive(): Promise<void> {
         TOOLHIVE_SKIP_DESKTOP_CHECK: 'true',
       },
     })
+
     log.info(`[startToolhive] Process spawned with PID: ${toolhiveProcess.pid}`)
```
aponcedeleonch: Re: the `spawn(...)` call just above (line 218 onwards): this config doesn't tie thv's lifetime to studio's. `detached: false` on POSIX only controls the process group, it doesn't kill the child when the parent dies, and I don't see a parent-death watcher on the thv side either (no `PR_SET_PDEATHSIG`, no ppid polling).

If studio is SIGKILL'd (force-quit, OOM, debugger detach, hard crash), thv keeps running and keeps holding the discovery file. The next studio launch calls `Discover`, sees the orphan as healthy via `/health`, gets back `StateRunning`, and exits with "another ToolHive server is already running." The user lands on `AlreadyRunningError` until they manually `pkill thv`.

Under the 1:1 invariant this is the only failure mode I'd call a real bug introduced by the new architecture, since it's sticky across launches.

Options:
- thv opens a pipe inherited from studio and exits when the read end closes on parent death (cross-platform, no platform-specific calls).
- Linux: `PR_SET_PDEATHSIG(SIGTERM)`; macOS: kqueue `NOTE_EXIT` on ppid; Windows: studio creates a job object with `KILL_ON_JOB_CLOSE` and assigns the thv child to it.
- Studio-side stopgap: at startup, if `Discover` returns `StateRunning`, check whether the thv's parent PID is alive; if it's an orphan, reap it before bailing out with `ALREADY_RUNNING`.

I'd lean toward the inherited-pipe approach since it's portable and the contract is easy to reason about. Happy to file a follow-up issue if you'd prefer to land this PR first.

e2e-tests/helpers/app-relaunch.ts
```diff
@@ -74,18 +75,20 @@ export async function launchApp(userDataDir: string): Promise<LaunchedApp> {
 
   await window.getByRole('link', { name: /mcp servers/i }).waitFor()
```
aponcedeleonch, line 76: Re: the `env: {}` block in `electron.launch` just above (around line 57): the discovery dir is hardcoded to `xdg.StateHome/toolhive/server` (`pkg/server/discovery/discovery.go:51`). No flag, no env override exposed on the thv side, but `xdg.StateHome` itself honors `XDG_STATE_HOME`.

This helper isolates `userDataDir` per launch but not the discovery state, so any parallel run on the same host collides on the global discovery file:
- dev running studio locally + e2e suite on the same box
- two parallel e2e jobs on a shared CI runner
- retry attempts that overlap

Two reasonable fixes:
- Quick: set `XDG_STATE_HOME` to a per-launch tmp dir in the env block here, and mirror that in `toolhive-manager.ts` when `TOOLHIVE_E2E=true`.
- Cleaner: have thv expose `--discovery-dir=` so studio can pin it explicitly without piggybacking on XDG.

Not blocking, but worth noting in the test plan since the change makes the discovery file load-bearing for startup.

main/src/toolhive-manager.ts
```diff
@@ -349,6 +390,10 @@ export function stopToolhive(options?: { force?: boolean }): void {
     scheduleForceKill(processToKill, pidToKill)
   }
 
+  if (toolhiveSocketPath) {
+    cleanupSocketFile(toolhiveSocketPath)
```
aponcedeleonch, line 414: After `cleanupSocketFile`, `toolhiveSocketPath` is still set to the path we just unlinked. `getToolhiveSocketPath()` keeps returning it, `hasToolhiveConnection()` keeps returning true, and `requireSocketPath()` in `unix-socket-fetch` happily hands the dead path to `http.request`, which errors with `ENOENT`.

Mostly invisible during app shutdown (renderer is going away anyway), but it bites during `restartToolhive` (auto-update's `safeServerShutdown`, registry-auth recovery), where the renderer is still live and will surface `ENOENT` errors during the gap between stop and the next start.

The same issue exists in the `'exit'` handler around line 273: it clears `toolhiveProcess = undefined` but leaves `toolhiveSocketPath` behind, so an unexpected thv crash leaves studio with stale state.

Suggest clearing it in both places:

```ts
// in stopToolhive, after cleanupSocketFile
toolhiveSocketPath = undefined

// in the 'exit' handler
toolhiveProcess = undefined
toolhiveSocketPath = undefined
```

aponcedeleonch (commented):
A few comments on lifecycle coupling between studio and the spawned thv. None block the PR landing once the matching thv is in.

main/src/toolhive-manager.ts
samuv: Agreed this is a real gap for SIGKILL/hard-crash; we’ll track it in a follow-up (parent-death contract: pipe / PDEATHSIG / job object / orphan reaping before ALREADY_RUNNING). Not blocking this transport PR.

e2e-tests/helpers/app-relaunch.ts
samuv: Good call — we’ll defer isolating discovery state (e.g. per-run `XDG_STATE_HOME` in e2e + matching studio when `TOOLHIVE_E2E`, or a future `thv --discovery-dir`) to a follow-up PR; noted for CI/local parallel runs.

main/src/toolhive-manager.ts
samuv: Addressed: we clear `toolhiveSocketPath` / `toolhiveMcpPort` after `cleanupSocketFile` in `stopToolhive`, clear stale managed state on the no-process early return only when `!isUsingCustomSocket()`, and only clear globals in `exit` when `toolhiveProcess === child` so restart doesn’t wipe the new socket. Tests cover managed stop + external `THV_SOCKET` unchanged.

peppescg (approved):
🚢

PR #1720, opened 2026-03-10
refactor(main): extract IPC handlers into dedicated modules

samuv:
moves all ipcMain.handle calls from main.ts into categorized files under main/src/ipc-handlers/ (app, dark-mode, window, toolhive, telemetry, auto-update, dialogs, feature-flags, cli, utils), with chat handlers further split into their own subfolder (providers, streaming, settings, mcp-tools, threads)

main.ts now just wires everything up via initSentry(), enforceSingleInstance(), registerAllEvents() and registerAllHandlers()

### DIFF
--- a/main/src/ipc-handlers/__tests__/index.test.ts
+++ b/main/src/ipc-handlers/__tests__/index.test.ts
@@ -1,0 +1,51 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+const mocks = vi.hoisted(() => ({
+  registerApp: vi.fn(),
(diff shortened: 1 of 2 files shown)


main/src/main.ts
```diff
@@ -1,107 +1,13 @@
-import type { LanguageModelV2Usage } from '@ai-sdk/provider'
-import type { UIMessage } from 'ai'
-import { app, ipcMain, nativeTheme, dialog } from 'electron'
+import { app } from 'electron'
```
peppescg, line 1: love this 🧹

PR #1789, opened 2026-03-23
ci: bump ollama

kantord:
this pr modernizes the ollama version, improves ollama's cache key logic

### DIFF
--- a/e2e-tests/fixtures/electron.ts
+++ b/e2e-tests/fixtures/electron.ts
@@ -18,8 +18,8 @@ export const TEST_SECRET_NAME = 'E2E_TEST_SECRET'
 /** Default timeout for most operations (10 seconds) */
 export const DEFAULT_TIMEOUT = 10_000
 
-/** Long timeout for operations that may take a while, like LLM inference on CI (120 seconds) */
-export const LONG_TIMEOUT = 120_000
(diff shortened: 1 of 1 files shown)


e2e-tests/fixtures/electron.ts
```diff
@@ -18,8 +18,8 @@ export const TEST_SECRET_NAME = 'E2E_TEST_SECRET'
 /** Default timeout for most operations (10 seconds) */
 export const DEFAULT_TIMEOUT = 10_000
 
-/** Long timeout for operations that may take a while, like server installation (60 seconds) */
-export const LONG_TIMEOUT = 60_000
+/** Long timeout for operations that may take a while, like LLM inference on CI (180 seconds) */
+export const LONG_TIMEOUT = 180_000
```
samuv, line 22: do you think is needed even with ollama with docker?

kantord: screenshot-2026-03-23_18-10-23


with this, the test is no longer flaky. I didn't do a detailed benchmark but the conclusion is that overall, with the higher timeout it should be faster most times.

The only downside is that in the case of some legit failures, it will be slower. But in practice, this cost is rarely paid.

PR #1790, opened 2026-03-23
fix(build): resolve Windows auto-update checksum failure

samuv:
Fixes the Windows auto-update failure where Squirrel.Windows throws "Checksummed file size doesn't match" on every update check. Follows up on #1738 which switched to the S3/CloudFront update source.

The root cause: `remoteReleases` in maker-squirrel fetches the CDN's existing RELEASES file and appends the new version's entries, so the RELEASES file accumulates references to all past versions. But the CI promotion step (`aws s3 sync --delete`) correctly removes old nupkg files from `latest/`, creating a mismatch — the RELEASES file references nupkgs that no longer exist (HTTP 403), which causes Squirrel's checksum validation to fail.

Verified on the live CDN — the RELEASES file at `releases.toolhive.dev/stable/latest/win32/x64/RELEASES` listed 5 entries but only the 2 current nupkgs (0.25.0) were actually accessible:

| File | Status |
|---|---|
| `ToolHive-0.23.0-full.nupkg` | 403 |
| `ToolHive-0.24.0-delta.nupkg` | 403 |
| `ToolHive-0.24.0-full.nupkg` | 403 |
| `ToolHive-0.25.0-delta.nupkg` | 200 |
| `ToolHive-0.25.0-full.nupkg` | 200 |

Changes:
- Remove `remoteReleases` and set `noDelta: true` — each build now produces a RELEASES file with only the current version's entry, staying in sync with what's on S3. Trade-off: no more delta updates (~10 MB), only full updates (~184 MB). Acceptable since Squirrel.Windows delta updates are fragile and Squirrel is no longer maintained.
- Add a `postMake` hook that strips the UTF-8 BOM from RELEASES files (defense-in-depth — `electron-winstaller` adds a BOM that can cause edge-case parsing failures).

The CDN's RELEASES file was also manually fixed via `aws s3 cp` + CloudFront invalidation to unblock existing users immediately.

### DIFF
--- a/forge.config.ts
+++ b/forge.config.ts
@@ -11,6 +11,7 @@ import { ensureThv } from './utils/fetch-thv'
 import MakerTarGz from './utils/forge-makers/MakerTarGz'
 import MakerDMGWithArch from './utils/forge-makers/MakerDMGWithArch'
 import { isPrerelease } from './utils/pre-release'
+import { stripBomFromReleasesFiles } from './utils/forge-makers/strip-bom-from-releases'
 import packageJson from './package.json'
 
 function isValidPlatform(platform: string): platform is NodeJS.Platform {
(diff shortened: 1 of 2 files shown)


peppescg (approved):
nice catch

PR #1793, opened 2026-03-24
refactor(db): remove SQLITE_READS_* feature flags and dual-write pattern

samuv:
Remove the `SQLITE_READS_*` feature flags and the electron-store dual-write pattern, making SQLite the single source of truth for all data domains (settings, threads, chat settings, shutdown, and feature flags).

The dual-write strategy was a safe rollout mechanism, every mutation wrote to both electron-store and SQLite, with reads gated by per-domain feature flags. Now that SQLite is proven stable, this removes the indirection.

- Remove all 5 `SQLITE_READS_*` flag keys, keeping only `META_OPTIMIZER`
- Simplify `getFeatureFlag` to always read from SQLite (no more circularity guard or `readFromElectronStore`)
- Remove electron-store reads and dual-writes from settings, threads, chat-settings, and shutdown modules
- Convert `reconcileFromStore` to a one-time migration using a `migration_from_store_complete` sentinel in SQLite — upgrading users get their data migrated on first launch, subsequent starts skip immediately
- Electron-store instances are kept (exported) solely for the one-time migration; they can be deleted in a follow-up after one release cycle
- Update all tests to assert against SQLite readers/writers instead of electron-store mocks

### DIFF
--- a/main/src/db/writers/chat-settings-writer.ts
+++ b/main/src/db/writers/chat-settings-writer.ts
@@ -1,6 +1,5 @@
 import { getDb, isDbWritable } from '../database'
 import { encryptSecret } from '../encryption'
-import log from '../../logger'
 import { withDbSpan } from '../telemetry'
 
 export function writeProvider(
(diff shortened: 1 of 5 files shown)


samuv:
@claude review this change

PR #1859, opened 2026-03-30
feat: add new deep link to initiate install flow directly

kantord:
this adds a new deep link that can be used for initiating the install flow directly, rather than just showing the registry detail page

Here is an example deep link: toolhive-gui://v1/open-registry-server-install?serverName=fetch

Screenshot of it in action (nothing interesting to see in the form itself, this just shows which phase of the installation flow you end up in): 

screenshot-2026-03-30_14-58-02

### DIFF
--- a/common/deep-links.ts
+++ b/common/deep-links.ts
@@ -13,7 +13,6 @@ export type NavigateTarget = {
   to: string
   params?: Record<string, string>
   search?: Record<string, unknown>
-  openInstall?: boolean
 }
 
 /**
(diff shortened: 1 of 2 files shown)


peppescg:
I tested using an html page. It works nice 🚀 
it is expected that clicking on `toolhive-gui://v1/open-registry-server-install?serverName=fetch` it also redirect to the detail page? and it seems if I am already in the detail page, after clicking on the above link I don't see any modal for installing the server

Screenshot 2026-03-30 at 15 14 24

kantord:
> I tested using an html page. It works nice 🚀 it is expected that clicking on `toolhive-gui://v1/open-registry-server-install?serverName=fetch` it also redirect to the detail page? and it seems if I am already in the detail page, after clicking on the above link I don't see any modal for installing the server
> 
> Screenshot 2026-03-30 at 15 14 24

let me see, actually that edge case/bug seems logical, I didn't think about it

PR #1860, opened 2026-03-30
fix(remote-auth): re-saving server config corrupts client_secret on OAuth servers

### LINKED ISSUE #1821: [Bug] Re-saving configuration for servers with remote auth fails
### Bug Description

If I have a remote MCP server that requires an auth flow, e.g github-remote, and I click "Edit configuration" and save without any changes, then my oauth flow fails with wrong credentials.

From what I can tell the client secret is incorrectly saved after the update.
This is a subset of my runconfig before the save:
```
  "remote_auth_config": {
    "client_id": "xxx",
    "client_secret": "OAUTH_CLIENT_SECRET_github-remote_1774535530371142000_c31967b5,target=oauth_secret",
```

And this is my runconfig after the save (note that I didn't change the secret myself):
```
  "remote_auth_config": {
    "client_id": "xxx",
    "client_secret": "OAUTH_CLIENT_SECRET_github-remote_1774536454069004000_11d24298_2,target=OAUTH_CLIENT_SECRET_github-remote_1774536454069004000_11d24298_2",
```

### Steps to Reproduce

1. Create a github-remote MCP server with a valid client ID and secret
2. Complete the OAuth browser flow that's automatically started 
3. Observe that the MCP server works, e.g through playground
4. Click "Edit configuration" on the github-remote MCP server
5. Save without making any changes

### Expected Behavior

I expect the OAuth flow to start and complete successfully. After the OAuth flow completes, I expect my MCP server to be available for me to use.

### Actual Behavior

The GitHub OAuth flow fails, stating I have invalid credentials. The MCP server is in an error state.

### Priority

Medium

### Environment

_No response_

### Additional Context

_No response_

samuv:
Re-saving a remote MCP server configuration without any changes breaks the OAuth
flow because the client_secret reference gets corrupted in the secret store.

**Root cause**: `convertCreateRequestToFormData` used `client_secret.target` (e.g. `"oauth_secret"`) to populate `value.secret` and to check `isFromStore`. Since `target` is a config property name and not a secret store key, the check always returned `false` — causing every re-save to treat an existing secret as new plaintext. A new secret was then created with the wrong value, gained a collision suffix (`_2`), and had both `name` and `target` set to this new
corrupted key.

**Fix**: use `client_secret.name` (the actual store key) for both fields. Same fix applied to `bearer_token`.

### Changes

- Fix `convertCreateRequestToFormData` in `orchestrate-run-remote-server.tsx` to use `.name` instead of `.target` for `client_secret` and `bearer_token`
- Add unit regression test covering the `name ≠ target` case in `orchestrate-run-remote-server.test.tsx`
- Add component integration test that reproduces the exact #1821 scenario (load OAuth server → save without changes → assert no secret recreation and correct `client_secret` reference in the edit payload)
- Add E2E spec that installs a bearer token server, re-saves without changes, and verifies it stays Running

Closes #1821

### DIFF
--- a/renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
+++ b/renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
@@ -484,11 +484,7 @@ describe('DialogFormRemoteMcp', () => {
           name: 'OAUTH_CLIENT_SECRET_GITHUB_REMOTE',
           target: 'oauth_secret',
         },
-        scopes: ['repo'],
-        skip_browser: false,
-        use_pkce: true,
         callback_port: 8080,
-        oauth_params: {},
       },
       group: 'default',
     }))


renderer/src/features/mcp-servers/components/remote-mcp/__tests__/dialog-form-remote-mcp.test.tsx
```diff
@@ -467,6 +467,94 @@ describe('DialogFormRemoteMcp', () => {
     })
   })
 
+  it('re-saving a server with remote OAuth auth does not corrupt client_secret (issue #1821)', async () => {
+    const user = userEvent.setup({ delay: null })
+    const rec = recordRequests()
+
+    mockedGetApiV1BetaWorkloadsByName.override((data) => ({
+      ...data,
+      name: 'github-remote',
```
kantord, line 476: nitpick, but probably it's not necessary to override this many fields in the test data for the tests to work?

samuv: Yeah, I removed the ones that aren’t needed

PR #1890, opened 2026-03-31
feat(newsletter): open newsletter modal from help dropdown

samuv:
- Add `NewsletterModalContext` and `NewsletterModalProvider` to share open state across the tree
- Update `NewsletterModal` to bypass the auto-show gate when `forceOpen` is true, and reset the forced state on dismiss/close
- Add a "Newsletter" item to `HelpDropdown` that calls `openNewsletterModal()` via the context
- Wrap the root component tree with `NewsletterModalProvider` so both `TopNav` and `NewsletterModal` share the same context
- Update `newsletter-modal.test.tsx`, `help-dropdown.test.tsx`, and `render-route.tsx` to include the provider in test wrappers

https://github.com/user-attachments/assets/2902b10d-247f-40f3-bbb0-289cf27863ea

### DIFF
--- a/renderer/src/common/components/newsletter-modal.tsx
+++ b/renderer/src/common/components/newsletter-modal.tsx
@@ -193,21 +193,20 @@ export function NewsletterModal() {
     queryFn: () => window.electronAPI.getNewsletterState(),
   })
 
-  if (isLoading || !newsletterState) return null
+  if (!forceOpen && (isLoading || !newsletterState)) return null
+  if (forceOpen && newsletterState?.subscribed) return null
 
-  if (!forceOpen) {
-    if (closed) return null
-    if (
-      !successMessage &&
-      !shouldShowAfterDismissal(
-        newsletterState.subscribed,
-        newsletterState.dismissedAt,
-        DISMISS_DAYS
-      )
-    ) {
-      return null
-    }
-  }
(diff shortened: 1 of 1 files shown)


renderer/src/common/components/newsletter-modal.tsx
```diff
@@ -185,33 +186,41 @@ export function NewsletterModal() {
...
+    if (
+      !successMessage &&
+      !shouldShowAfterDismissal(
+        newsletterState.subscribed,
+        newsletterState.dismissedAt,
+        DISMISS_DAYS
+      )
+    ) {
+      return null
+    }
```
peppescg: cause it is retuning null for both branch, does it make sense to have const?

```
const shouldHideBecauseClosed = !forceOpen && closed
const shouldHideBecauseDismissed =
  !forceOpen &&
  !successMessage &&
  !shouldShowAfterDismissal(
    newsletterState.subscribed,
    newsletterState.dismissedAt,
    DISMISS_DAYS
  )

if (shouldHideBecauseClosed || shouldHideBecauseDismissed) {
  return null
}
```

kantord: I am for keeping the consts because they describe the meaning of the logic there a little bit.

samuv: sure, it makes sense,  pushed

PR #1891, opened 2026-03-31
fix(registry): show secrets and env vars sections when registry has none pre-defined

### LINKED ISSUE #1876: [Bug] Can't add secrets or env vars to MCP server from registry
### Bug Description

I noticed that if an MCP server in the registry doesn't have any secrets or env vars defined in the registry, the Install Server form doesn't let you add any. I can't pinpoint the version, but I'm pretty certain this is something that used to be possible.

Example server: Playwright

Use case 1: Trying to debug an MCP server with logging/debugging flags that are not captured in their registry entry (like DEBUG=1)

Use case 2: A new version of an MCP server introduces env vars, but the registry entry hasn't been updated to include them

### Steps to Reproduce

1. Try to install Playwright from the registry (or any MCP which has no env vars defined in the registry).
2. Note that the Secrets and Environment Variables sections are missing; there is no opportunity to add them at install time.
3. After running the MCP, it is possible to edit the configuration and add env vars at this point - so the limitation on adding at install-time is confusing.

### Expected Behavior

Users should always be able to specify secrets and/or environment variables when installing an MCP server, even if there aren't any pre-defined in the registry.

### Actual Behavior

_No response_

### Priority

Low

### Environment

_No response_

### Additional Context

_No response_

peppescg:
## Summary

https://github.com/user-attachments/assets/a8eba54f-1873-43bd-a265-7009c3a61104

Screenshot 2026-03-31 at 14 17 47

- When a registry server has no secrets or env vars defined in its metadata, the install form now shows the dynamic "Add secret" / "Add environment variable" components instead of hiding the sections entirely
- Reuses existing `FormFieldsArrayCustomSecrets` and `FormFieldsArrayCustomEnvVars` components (already used in the local MCP form)
- The schema already supports this — `createDynamicNameSchema([])` allows any string name when no pre-defined names exist

Closes #1876

## Test plan
- [ ] Install a registry server with no env vars (e.g. Playwright) — verify "Secrets" and "Environment variables" sections appear with "Add" buttons
- [ ] Add a custom env var (e.g. `DEBUG=1`) and verify it's passed to the server
- [ ] Install a registry server WITH pre-defined env vars — verify the existing behavior is unchanged
- [ ] All 33 existing tests pass

### DIFF
--- a/renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
+++ b/renderer/src/features/registry-servers/components/form-run-from-registry/configuration-tab-content.tsx
@@ -275,7 +275,12 @@ export function ConfigurationTabContent({
       <FormFieldsArrayVolumes<FormSchemaRegistryMcp> form={form} />
 
       <section className="mb-6">
-        <Label className="mb-2">Secrets</Label>
+        <Label
+          className="mb-2"
+          htmlFor={predefinedSecretsCount > 0 ? 'secrets.0.value' : undefined}
+        >
+          Secrets
+        </Label>
         <p className="text-muted-foreground mb-6 text-sm">
           All secrets are encrypted and securely stored by ToolHive.
         </p>
@@ -386,7 +391,12 @@ export function ConfigurationTabContent({
       </section>
 
       <section className="mb-6">
-        <Label className="mb-2">Environment variables</Label>
+        <Label
+          className="mb-2"
+          htmlFor={predefinedEnvVarsCount > 0 ? 'envVars.0.value' : undefined}
+        >
+          Environment variables
+        </Label>
         <p className="text-muted-foreground mb-6 text-sm">
           Environment variables are used to pass configuration settings to the
           server.


peppescg:
@claude succinct review

renderer/src/features/mcp-servers/components/form-fields-array-custom-env-vars.tsx
```diff
@@ -16,7 +16,7 @@ import { DynamicArrayField } from '@/features/registry-servers/components/dynami
 type FormWithEnvVars = {
   envVars: Array<{
     name: string
-    value: string
+    value?: string
```
samuv, line 19: it is optional tho?

peppescg: this component was not used in registry before this pr, and after using it in registry there was a type mismatch, cause the value is optional from the schema validation

renderer/src/common/lib/form-schema-mcp.ts
```diff
@@ -220,21 +220,18 @@ export const addNetworkValidation = (ctx: z.RefinementCtx, data: unknown) => {
...
 // --------
 // REGISTRY SERVER
 
 const createRegistrySecretsSchema = (
-  secretNames: string[],
   secrets: Array<{ name?: string; required?: boolean }>,
   isEmptyValueFn: (value: unknown) => boolean
 ) => {
-  const secretNameSchema = createDynamicNameSchema(secretNames)
+  const secretNameSchema = createDynamicNameSchema()
```
samuv, line 234: This changes the logic for required secrets though, right?

peppescg: Previously `createDynamicNameSchema` received an array of predefined names and created a union of literals, now it accepts custom values from user

samuv: it wound't be better something like:

```
z.union(names.map((name) => z.literal(name))).or(z.string())
```

peppescg: makes sense

peppescg: mm zod is translating it in string cause `"GITHUB_TOKEN" | "API_KEY" | string` is equal to string, I mean it is how ts works. Zod is not able to translate this `  type Name = "GITHUB_TOKEN" | "API_KEY" | (string & {})` it will infer string at runtime 😆 
it seems there isn't a way to do this without a cast where it is used, that I would avoid

samuv: 🤷🏼‍♂️  ok then

PR #1909, opened 2026-04-02
refactor: create app-info file

kantord:


kantord:
@claude  deep review please. make your report very succinct

kantord:
@claude please do another round of review. i made some changes, addressing everythin you pointed out that was actually actionable/worth addressing + made some additional changes

scripts/install-deep-link-handler.ts
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env tsx
```
kantord: note: this looks like a big change but basically it's just a dev helper script for testing deep links on linux and it was re-written from .sh to .ts just so that it can import the same constant. there are no real changes other than that

common/app-info.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * App identity constants — single source of truth for build-configurable
+ * values shared across forge config, deep links, and scripts.
+ */
+
+export const DEEP_LINK_PROTOCOL = 'toolhive-gui'
+export const FLATPAK_APP_ID = 'com.stacklok.ToolHive'
```
samuv, line 7: IDK, I feel this should live in common in any case, so we don’t have specific aliases.

kantord: so you mean, move to common/app-info.ts and the import it as `@common/app-info`?

samuv: yep!

kantord: makes sense. pushed this change, sorry for the delay.

PR #1910, opened 2026-04-02
feat(mcp-optimizer): sunset deprecation banner and hide experimental toggle

samuv:
MCP Optimizer is being removed from the desktop app. This PR starts the sunset by:

- Showing a deprecation banner on the MCP Servers pages for users who have the feature enabled: "MCP Optimizer is leaving the desktop app. Learn about its future here."
- Hiding the experimental toggle in Settings for users who have never enabled it (users who already have it enabled can still toggle it off)
- Adjusting the sidebar top offset and main content height to correctly account for the banner

Screenshot 2026-04-03 at 10 03 34

### DIFF
--- a/main/src/feature-flags/__tests__/flags.test.ts
+++ b/main/src/feature-flags/__tests__/flags.test.ts
@@ -34,10 +34,7 @@ import {
   enableFeatureFlag,
   disableFeatureFlag,
 } from '../flags'
-import {
-  writeFeatureFlag,
-  deleteFeatureFlag,
-} from '../../db/writers/feature-flags-writer'
+import { writeFeatureFlag } from '../../db/writers/feature-flags-writer'
 import { readFeatureFlag } from '../../db/readers/feature-flags-reader'
 
 describe('Feature Flags', () => {
@@ -146,11 +143,14 @@ describe('Feature Flags', () => {
 
   describe('disableFeatureFlag', () => {
     it.each(Object.entries(featureFlagKeys))(
-      'should delete from SQLite for %s when disabled',
+      'should write false to SQLite for %s when disabled',
       (_name, key) => {
         disableFeatureFlag(key)
 
-        expect(deleteFeatureFlag).toHaveBeenCalledWith(`feature_flag_${key}`)
+        expect(writeFeatureFlag).toHaveBeenCalledWith(
+          `feature_flag_${key}`,
+          false
+        )
       }
     )
   })
--- a/main/src/feature-flags/flags.ts
+++ b/main/src/feature-flags/flags.ts
@@ -2,10 +2,7 @@ import Store from 'electron-store'
 import log from '../logger'
 import { featureFlagKeys } from '../../../utils/feature-flags'
 import type { FeatureFlagKey, FeatureFlagOptions } from './types'
-import {
-  writeFeatureFlag,
-  deleteFeatureFlag as deleteFeatureFlagFromDb,
-} from '../db/writers/feature-flags-writer'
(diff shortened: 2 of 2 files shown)


renderer/src/common/components/mcp-optimizer-sunset-banner.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Alert, AlertDescription } from './ui/alert'
+import { useMcpOptimizerBannerVisible } from '@/common/hooks/use-mcp-optimizer-banner-visible'
+
+const BLOG_POST_URL =
+  'https://stacklok.com/blog/mcp-optimizer-is-now-built-into-the-stacklok-platform/?utm_source=toolhive-studio'
+
+export function McpOptimizerSunsetBanner() {
+  const isVisible = useMcpOptimizerBannerVisible()
+
+  if (!isVisible) return null
```
kantord, line 11: due to the temporary nature of this, there is absolutely no need to address this nitpick, but I generally think it's a better idea to control a components visibility from the outside whenever practical. (Otherwise when you read the component that uses it, you will be mislead into thinking it's always visible)

renderer/src/common/hooks/use-mcp-optimizer-banner-visible.ts
```diff
@@ -0,0 +1,13 @@
+import { useMatches } from '@tanstack/react-router'
+import { useFeatureFlag } from './use-feature-flag'
+import { featureFlagKeys } from '../../../../utils/feature-flags'
+import { MCP_OPTIMIZER_BANNER_ROUTES } from '../lib/constants'
+
+export function useMcpOptimizerBannerVisible() {
+  const matches = useMatches()
+  const isMetaOptimizerEnabled = useFeatureFlag(featureFlagKeys.META_OPTIMIZER)
+  const isOnServerPage = matches.some((m) =>
```
kantord, line 9: would it not make more sense to just render is on the server page? then this condition will not be required 🤔 

in any case, I imagine this will be removed in a few weeks so don't worry much about it

renderer/src/features/mcp-servers/components/mcp-servers-sidebar.tsx
```diff
@@ -8,11 +9,15 @@ interface McpServersSidebarProps {
...
+
   return (
     <aside
-      className="border-input text-sidebar-foreground w-sidebar fixed top-16
-        bottom-0 left-0 z-0 flex h-full flex-col items-start gap-3 border-r px-4
-        pt-4 pb-4"
+      className="border-input text-sidebar-foreground w-sidebar fixed bottom-0
+        left-0 z-0 flex h-full flex-col items-start gap-3 border-r px-4 pt-4
+        pb-4"
+      style={{ top: sidebarTop }}
```
peppescg: I would avoid the style prop, here. It could be added directly in className using `cn`

renderer/src/common/components/mcp-optimizer-sunset-banner.tsx
samuv: I didn’t want to put logic in __root because it was temporary

renderer/src/common/hooks/use-mcp-optimizer-banner-visible.ts
samuv: It wouldn’t be possible in the position I chose, because it’s between the page itself and the top nav.

renderer/src/features/mcp-servers/components/mcp-servers-sidebar.tsx
samuv: sure I can move it to the className

PR #1915, opened 2026-04-03
feat(playground): threaded chat with sidebar, LLM titles, star/rename/delete, and scroll persistence

samuv:
This PR turns the Playground from a single stateless chat into a fully-featured threaded chat experience, on par with Claude Desktop and ChatGPT.

## What's new

- **Thread sidebar** — lists all chat threads in a collapsible left-hand panel with a `+ New chat` row at the top. Starred threads are pinned above recents with section headers.
- **Per-thread actions** — each thread has a three-dot menu for Star/Unstar, Rename (also available via double-click), and Delete (with a confirmation dialog).
- **Thread title bar** — compact inline header showing the active thread's title with pencil-edit, star toggle, and delete button; replaces the generic `TitlePage`.
- **LLM-generated titles** — after the first AI response, the thread title is auto-generated by the active model via a new `chat:generate-thread-title` IPC handler. Manual renames set a `titleEditedByUser` flag that prevents auto-overwriting.
- **Per-thread scroll persistence** — `useAutoScroll` stores and restores scroll position per thread for the duration of the session; switching threads and back remembers where you were.
- **Full-bleed playground layout** — `Main` gains `relative` positioning so the sidebar and chat area can fill the viewport correctly without floating padding artefacts.
- **DB migrations** — two new SQLite migrations (`002-thread-title-flag`, `003-thread-starred`) adding `title_edited_by_user` and `starred` columns to the `threads` table.

https://github.com/user-attachments/assets/21af23ad-34d5-4769-a283-76f370f0cc7f

### DIFF
--- a/renderer/src/features/chat/components/__tests__/chat-interface.test.tsx
+++ b/renderer/src/features/chat/components/__tests__/chat-interface.test.tsx
@@ -173,11 +173,13 @@ describe('ChatInterface', () => {
       expect(screen.getByText('Loading chat history...')).toBeInTheDocument()
     })
 
-    it('hides messages area while loading', () => {
+    it('hides messages while loading (loading overlay covers the scroll container)', () => {
       mockStreamingReturn.isPersistentLoading = true
-      setMessages([{ id: 'msg-1', role: 'user', parts: [] }])
+      // Messages are only rendered inside the scroll container when hasMessages is true.
+      // While isPersistentLoading, messages haven't loaded yet so the array is empty.
       renderInterface()
       expect(screen.queryByTestId('message-msg-1')).not.toBeInTheDocument()
+      expect(screen.getByText('Loading chat history...')).toBeInTheDocument()
     })
   })
 
@@ -244,7 +246,6 @@ describe('ChatInterface', () => {
 
   describe('scroll-to-bottom button', () => {
     it('shows scroll-to-bottom button when showScrollToBottom is true', () => {
-      setMessages([{ id: 'msg-1', role: 'user', parts: [] }])
       mockShowScrollToBottom = true
       renderInterface()
       expect(
--- a/renderer/src/features/chat/components/chat-interface.tsx
+++ b/renderer/src/features/chat/components/chat-interface.tsx
@@ -49,7 +49,6 @@ export function ChatInterface({
   const [isSettingsOpen, setIsSettingsOpen] = useState(false)
 
   const { containerRef, showScrollToBottom, scrollToBottom } = useAutoScroll({
-    contentDep: messages,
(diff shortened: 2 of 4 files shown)


peppescg:
https://github.com/user-attachments/assets/07f4dae0-00ef-4e96-846b-bafa507bf14d

- It seems the star option section is not correctly expanded and it is difficult to click on it
- clicking on the second chat star goes below the traffic lights, it should have an higher z-index or force to be visibile on the left

peppescg:
Screenshot 2026-04-07 at 14 57 44

I guess it is not related to this change, but once you scroll up, and the icon appear, if you change the chat, or open a new chat the arrow icon continue to be visibile

samuv:
> * It seems the star option section is not correctly expanded and it is difficult to click on it
> * clicking on the second chat star goes below the traffic lights, it should have an higher z-index or force to be visibile on the left

those I think was alredy fixed


> I guess it is not related to this change, but once you scroll up, and the icon appear, if you change the chat, or open a new chat the arrow icon continue to be visibile

This I just made a refactor of the auto scroll hook, simplify the logic that was a little bit convoluted

renderer/src/features/chat/hooks/use-playground-threads.ts
```diff
@@ -0,0 +1,221 @@
...
+          window.electronAPI.chat.setActiveThreadId(lightweight[0].id)
+        }
+      } catch (err) {
+        log.error('[usePlaygroundThreads] Failed to load threads:', err)
+      } finally {
+        setIsLoading(false)
+      }
+    }
+
+    loadThreads()
```
peppescg, line 52: it would be nice to use react-query for the fetch
```
const threadsQuery = useQuery({
  queryKey: ['threads'],
  queryFn: async () => {
    return await window.electronAPI.chat.getAllThreads()
  },
  select: (allThreads) => {
    const sorted = [...allThreads].sort(
      (a, b) => b.lastEditTimestamp - a.lastEditTimestamp
    )

    return sorted.map(({ id, title, starred, lastEditTimestamp, createdAt }) => ({
      id,
      title,
      starred,
      lastEditTimestamp,
      createdAt,
    }))
  },
})
```

and use the useEffect only for setState
```
const threads = threadsQuery.data ?? []
const isLoading = threadsQuery.isLoading

useEffect(() => {
  if (threads.length > 0 && threads[0]) {
    setActiveThreadId(threads[0].id)
    window.electronAPI.chat.setActiveThreadId(threads[0].id)
  }
}, [threads])
```

samuv: yeah let me iterate in a separated PR

PR #1960, opened 2026-04-10
test: use temporary user data dir for e2e tests

kantord:
With this, end to end tests run in a temporary user data dir, instead of your global one. This improves test isolation for anything that depends on this path (namely, the sqlite db used for chat persistence for instance)


samuv (approved):
looks good

PR #1994, opened 2026-04-14
feat: show registry tab as read-only when permission is disabled

peppescg:
## Summary
**Read-only mode**
Screenshot 2026-04-14 at 18 20 11

- Always show the Registry settings tab, even when the `SETTINGS_REGISTRY_TAB` permission is disabled
- When permission is disabled, form fields are disabled and Save/Reset buttons are hidden (read-only mode)
- Moved permission check from `SettingsTabs` into `RegistryTab` itself, simplifying the tab container

## Test plan
- [x] Existing tests updated: `settings-tabs.test.tsx` verifies tab is visible but read-only
- [x] New test added: `registry-tab.test.tsx` verifies disabled combobox and hidden buttons with permission off
- [x] All 1788 tests pass
- [x] Type-check and lint pass

### DIFF
--- a/renderer/src/common/components/settings/registry/registry-form.tsx
+++ b/renderer/src/common/components/settings/registry/registry-form.tsx
@@ -17,6 +17,7 @@ interface RegistryFormProps {
   hasRegistryError: boolean
   isUnavailableError: boolean
   registryAuthRequiredMessage?: string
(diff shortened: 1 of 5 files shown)


samuv (approved):
🥷

PR #2003, opened 2026-04-15
feat(playground): thread ID in URL with sub-route architecture

samuv:
The Playground route was a single flat page with thread selection managed as React state. This refactors it into a proper sub-route hierarchy so the active thread is always reflected in the URL, enabling deep linking, back-navigation, and a foundation for future sub-sections (evals, prompt templates, etc.).

- Convert `/playground` to a layout route (permission gate + Outlet) and add a `/playground/` index that resolves the last-viewed thread, validating the stored active ID against the real thread list, falling back to most-recent, or creating a new thread —> then redirects to `/playground/chat/$threadId`
- Add `/playground/chat/$threadId` as the chat sub-route; thread ID comes from the URL and navigation (select, create, delete thread) is driven by `useNavigate` rather than `setState`
- Refactor `usePlaygroundThreads` to accept `activeThreadId` as input, persist it to IPC as a side-effect, and return thread IDs from `createThread` / `deleteThread` so the route component can navigate
- Update and extend tests for the new route structure and hook signature

### DIFF
--- a/renderer/src/routes/__tests__/playground.chat.test.tsx
+++ b/renderer/src/routes/__tests__/playground.chat.test.tsx
@@ -1,6 +1,6 @@
 import type { JSX } from 'react'
 import { describe, it, expect, vi, beforeEach } from 'vitest'
-import { screen, waitFor } from '@testing-library/react'
+import { screen, waitFor, fireEvent } from '@testing-library/react'
 import {
   createMemoryHistory,
   createRootRoute,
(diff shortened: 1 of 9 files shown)


samuv:
@claude review this

renderer/src/routes/__tests__/playground.chat.test.tsx
```diff
@@ -0,0 +1,392 @@
...
+        )}
+        {onDelete && (
+          <button data-testid="chat-delete-btn" onClick={onDelete} />
+        )}
+      </div>
+    )
+  },
+}))
+
+vi.mock('@/features/chat/components/playground-sidebar', () => ({
```
kantord, line 64: do we really need to mock all of this? it is so that the tests have no dependence on the DB? 🤔

renderer/src/routes/__tests__/playground.chat.test.tsx
```diff
@@ -0,0 +1,392 @@
...
+    id: 'thread-1',
+    title: 'My thread',
+    starred: false,
+    lastEditTimestamp: 2000,
+    createdAt: 1000,
+    ...overrides,
+  }
+}
+
+function renderChatRoute(
```
kantord, line 111: that is a lot of logic for this, isn't there a simpler way with less mocking? 🤔

renderer/src/routes/__tests__/playground.index.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+  id: 'thread-1',
+  title: 'Test thread',
+  starred: false,
+  lastEditTimestamp: 2000,
+  createdAt: 1000,
+  messages: [],
+  ...overrides,
+})
+
+function renderIndexRoute() {
```
kantord, line 65: again, that is a lot of mocking and boilerplate 🤔

renderer/src/routes/__tests__/playground.index.test.tsx
```diff
@@ -0,0 +1,195 @@
...
+    history: createMemoryHistory({ initialEntries: ['/playground/'] }),
+    defaultNotFoundComponent: () => null,
+  })
+
+  const queryClient = new QueryClient({
+    defaultOptions: { queries: { retry: false } },
+  })
+
+  const utils = render(
+    <NewsletterModalProvider>
```
kantord: these providers seem to repeat across different render util functions

renderer/src/routes/__tests__/playground.index.test.tsx
kantord: basically I'm wondering if it would be possible to just test using a more-or-less real router, just mocking the problematic stuff like DB queries and whatnot 🤔

renderer/src/routes/__tests__/playground.index.test.tsx
samuv: 🤖 Both test files were manually wiring up the full provider stack instead of using the existing renderRoute helper. Fixed in the latest commits: the NewsletterModalProvider / PermissionsProvider / PromptProvider / QueryClientProvider nesting has been replaced with renderRoute(router as unknown as ReturnType<typeof createTestRouter>, ...) in both files, matching the pattern already used in index.groups.test.tsx.

The custom multi-route router construction (createRootRoute + two createRoutes + Router) does need to stay, both tests need a second catch-all route to assert navigation after thread selection/deletion. createTestRouter only supports a single child route so it can't cover that case, but the provider wrapping is now fully delegated to the shared helper.

renderer/src/routes/__tests__/playground.index.test.tsx
samuv: Yeah, but I don’t think we’ll gain much in terms of coverage by avoiding some mocking here.
In these tests, we simplify the components to focus on testing the functions passed as props, since that’s where the logic we care about to test here.

peppescg (approved):
🚢

PR #2027, opened 2026-04-17
feat(skills): render SKILL.md content and sticky detail layout

samuv:
The skill and local build detail pages previously rendered a placeholder (`Skill.md rendering is not yet available`) where the SKILL.md body should live. This PR wires both detail pages to the `/api/v1beta/skills/content` endpoint, renders the markdown with streamdown, and reworks the shared detail layout so the summary + actions stay visible while the Skill.md content scrolls.

- Add `SkillMarkdown` component that fetches the body via `getApiV1BetaSkillsContent`, renders with `Streamdown` (code + mermaid + cjk plugins), and handles loading (skeleton), 404, 503, generic errors, and empty body states
- Add a `stripFrontmatter` prop on `SkillMarkdown` that drops the leading YAML frontmatter block before rendering; used on the local build detail page where `name` / `description` duplicate the header and summary
- Add `getSkillOciRef` helper that derives a content-endpoint ref from a `RegistrySkill`, preferring the first OCI package identifier and falling back to `namespace/name` (git package URLs are intentionally ignored — they point at the catalog repo, not a per-skill ref)
- Render the markdown panel on the registry skill detail page and the local build detail page; local builds pass `build.tag` directly since the `ref` query param accepts OCI refs and local build tags
- Swap the local build detail actions to `Remove` + `Install` to match the card footer order and let the single Install button on the registry detail render at its natural width instead of stretching the column
- Rework `SkillDetailLayout` so the left column (title, badges, Summary, actions) becomes `md:sticky md:top-0` while the Skill.md panel scrolls next to it
- When the original header scrolls out of view (via `IntersectionObserver`), reveal a condensed Back + Title + badges block at the top of the sticky column with a smooth `grid-template-rows` + opacity + margin transition; the transition is suppressed under `motion-reduce`
- Keep `Summary` aligned with `Skill.md` when the condensed header is hidden by collapsing the parent `flex gap-6` with a matching `-mb-6` on the reveal container
- Add MSW fixture and unit tests for `SkillMarkdown` covering success, 404, 503, generic errors, empty body, frontmatter stripping, and the non-stripping default; add unit tests for `getSkillOciRef`

**Not changed in this PR (deliberate, follow-up):**

- Registry skill detail page does not strip frontmatter today. We can opt in once we confirm the frontmatter on registry SKILL.md entries is also redundant with the header/summary.

https://github.com/user-attachments/assets/1aa90072-44b6-4f4f-8a94-5cdb9eca4f41

### DIFF
--- a/renderer/src/common/lib/streamdown-prose.ts
+++ b/renderer/src/common/lib/streamdown-prose.ts
@@ -1,0 +1,9 @@
+/**
+ * Shared Tailwind class list for Streamdown-rendered markdown content.
+ *
+ * Keep in sync across all sites that render prose through Streamdown
+ * (chat messages, SKILL.md viewer, etc.) so tweaks don't drift between
+ * features.
+ */
+export const STREAMDOWN_PROSE_CLASS =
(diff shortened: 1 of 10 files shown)


peppescg (approved):
🚢

PR #2045, opened 2026-04-20
feat(skills): align skills page and cards with design

samuv:
The skills page and installed skill cards had drifted from the current Figma design ([registry tab](https://www.figma.com/design/KuQGsgrNzRCGmSTl520Hwh/Toolhive-Open-Source?node-id=4217-12912), [installed tab](https://www.figma.com/design/KuQGsgrNzRCGmSTl520Hwh/Toolhive-Open-Source?node-id=4217-13004)). This PR aligns the page header, tabs, search/actions placement, and card internals with the mockups.

- Drop the `Skills` page title row and move per-tab actions into the tabs row in `skills-page.tsx`. Registry and Installed tabs show only a search input; Local Builds shows search + the `Build skill` action. The Installed empty state no longer exposes an install action — install is now reached via registry cards and the build flow — and instead points users at the Registry tab. The global `DialogInstallSkill`, `PlusIcon`, and `TitlePage` imports plus the `installOpen` state are removed from `SkillsPage`. The `Tabs` root is given `gap-4` so the 16px Figma spacing between the tabs row and the content is honoured.
- Introduce a reusable `pill` variant for the `Tabs` primitive in `tabs.tsx` using `class-variance-authority`. A small `TabsVariantContext` lets `TabsList` propagate its variant to child `TabsTrigger`s so callers only set `variant="pill"` in one place, with the option to override per-trigger. The default variant is unchanged, so `dialog-form-local-mcp.tsx` and `form-run-from-registry` keep their existing look. The pill list uses `bg-zinc-200 dark:bg-zinc-800` (≈ Figma `#e4e4e4`) because the global `--muted` token (`oklch(0.97 0 0)` in light mode) is too light to show a pill container against the page background.
- Move client badges on the installed skill card to their own row when the skill is project-scoped, so the first row holds only identity/metadata (status + scope + `/path`). User-scoped skills keep clients inline with scope to stay compact on short cards. Cap visible client badges at 3 and render a `+N` outline badge for the remainder; hovering it opens a `Tooltip` listing all hidden clients as a `<ul>`, and the badge exposes `aria-label="<n> more clients"` for accessibility.
- Differentiate the scope badge from client badges with a leading icon (`UserIcon` for `user`, `FolderGit2Icon` for `project`) instead of a variant swap, which barely renders in dark mode where `secondary` is nearly identical to the card background.
- Remove the card description tooltip in `card-skill-base.tsx`. The description still truncates visually via `line-clamp-3`, but the full text now lives on the detail page reached by clicking the card, so the hover tooltip was redundant. Title tooltip is kept as a quick affordance.

- The `+ Create` action on the Registry tab shown in the Figma is deferred — we don't have a UX for creating registry entries yet.
- The "All registries" selector and the grid/list view-switcher icons shown in the Figma are not wired up here.
- The base `--muted` token is not changed globally. Raising its lightness would ripple into skeletons, disabled states, and other consumers; the pill container opts into a dedicated colour instead.

https://github.com/user-attachments/assets/4fa804a5-0fb0-4397-abe2-0b687e3bde76

### DIFF
--- a/renderer/src/common/components/ui/tabs.tsx
+++ b/renderer/src/common/components/ui/tabs.tsx
@@ -1,7 +1,12 @@
 import * as React from 'react'
 import * as TabsPrimitive from '@radix-ui/react-tabs'
+import { cva, type VariantProps } from 'class-variance-authority'
 
 import { cn } from '@/common/lib/utils'
+
+type TabsVariant = 'default' | 'pill'
+
+const TabsVariantContext = React.createContext<TabsVariant>('default')
 
 function Tabs({
   className,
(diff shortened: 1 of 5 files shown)


peppescg (approved):
nice 👍

Issue #2033, opened 2026-04-17
[Task] Add card/table view toggle for MCP Servers and Skills pages


samuv:
## Description

Add a view switcher that lets users toggle between the existing card view and a new table view on both the **MCP Servers** page and the **Skills** page. The control sits in the page toolbar (next to the search input and "Create" button) and persists the user's preference.

The table view surfaces the same data as the cards (name, description, install state, actions) but in a denser, scannable layout that scales better for users with many servers/skills.

Relevant files:
- `renderer/src/features/registry-servers/components/grid-cards-registry.tsx`
- `renderer/src/features/registry-servers/components/card-registry-server.tsx`
- `renderer/src/features/skills/components/grid-cards-registry-skills.tsx`
- `renderer/src/features/skills/components/card-registry-skill.tsx`
- `renderer/src/routes/(registry)/-registry.route.tsx`

## Technical Details

- Add a segmented toggle (grid / list icons) to the registry toolbar, shared between MCP Servers and Skills.
- Introduce a table view that exposes the same row-level actions as the card (Install dropdown, Add to client dropdown, installed-state indicator, kebab menu).
- Persist the selected view (per page or global — to be confirmed during implementation) so the choice survives reloads.
- Keep the existing search, registry filter, and pagination controls working in both views.
- Reuse the existing `Table` primitives from `common/components/ui` where available.

## Acceptance Criteria

- [ ] A card/table toggle is visible in the toolbar on both the MCP Servers and Skills pages
- [ ] Switching to table view renders a table with columns for Server and About, plus the per-row actions shown in the design (Install, Add to client, installed-state icon, kebab menu)
- [ ] Card view continues to work as today
- [ ] Selected view preference persists across reloads
- [ ] Search, registry filter, and pagination work in both views
- [ ] Unit tests cover the toggle behavior and table rendering for both pages
- [ ] Lint, type-check, and existing tests pass

## Out of Scope

- Column sorting, resizing, or column visibility settings
- Changes to the card layout itself
- Changes to other pages (Registries, Clients, etc.)

## Priority

Medium

## Task Type

Feature

PR #2046, opened 2026-04-20
feat: add persisted card/table view toggle for MCP servers and installed skills

samuv:
The installed MCP servers page and the Skills "Installed" tab only rendered cards, and the registry tab Figma hints at a grid/list view switcher. This PR wires up a `ViewToggle` on both pages, adds a table rendering path that surfaces every affordance the card already exposes, and persists the user's choice per-page in the existing SQLite `settings` table so a reload brings them back to the same view.

The Figma frames for the MCP servers table and the skills table were used as a compass for the layout, but the goal was to preserve parity with the current cards — every piece of info, every conditional control (update button, status badge, client overflow tooltip), and every kebab action — not just the columns shown in the mockups.

related task https://github.com/stacklok/toolhive-studio/issues/2033

https://github.com/user-attachments/assets/e6fda555-2c21-4227-a9e6-82707493cd08

### Persistence (`main/` + `preload/`)

- Add a `ui-preferences` IPC namespace backed by the existing SQLite `settings` table. `UI_PREFERENCE_KEYS` enumerates the allowed keys (`ui.viewMode.mcpServers`, `ui.viewMode.skillsInstalled`) and the handlers reject anything that isn't `'card' | 'table'` at the main-process boundary. Unknown keys default to `'card'`.
- Expose `window.electronAPI.uiPreferences.{getViewMode, setViewMode}` from the preload, typed from the main-side source of truth so renderer callers never see `unknown`.
- Add matching stubs to the shared `electronAPI` mock so renderer tests don't need to touch IPC.

### Renderer primitives

- Add `useViewPreference(key)` hook (React Query) that loads the persisted value, defaults to `'card'` while the query is in-flight or errors out, and writes back optimistically so the UI flips without waiting on the DB round-trip.
- Add a `ViewToggle` segmented control (`LayoutGrid` / `List` from `lucide-react`) as two radios in a pill container, with proper aria state and an `onChange` that only fires on actual changes.
- Extend the shadcn `Table` primitive with an optional `containerClassName` prop. Its default wrapper uses `overflow-x-auto`, which produced an internal horizontal scrollbar inside rounded-border wrappers even with `table-layout: fixed`; callers can now merge/override those utilities via `tailwind-merge` without dropping the primitive.

### MCP servers page

- Render `ViewToggle` alongside the search input and branch between `CardMcpServer` grid and a new `TableMcpServers` based on `ui.viewMode.mcpServers`.
- `TableMcpServers` uses `table-layout: fixed` with **Server** `w-auto md:w-[28%]`, **About** `w-auto` (hidden below `md`, Server grows to fill), and fixed pixel widths for **Update** (`w-11`), **Status** (`w-[132px]`), **Type** (`w-10`), and **Actions** (`w-16 pr-3`). Row contents mirror `CardMcpServer`: truncated name + tooltip, truncated description + tooltip, the conditional `UpdateVersionButton` (amber, icon-only when there's registry drift), a start/stop `Switch` with live status label (`min-w-0 + flex-1` so "Unauthenticated" / "Transitioning" truncate cleanly inside 132px), local/remote icon with tooltip, and the full `ServerActionsDropdown` kebab.
- The Actions column is sized to comfortably fit the `size-9` trigger plus its `ml-2` and `pr-3` so the dropdown no longer hugs the rounded border.

### Skills Installed tab

- Render `ViewToggle` next to the search on the Installed tab only; Registry and Local Builds are intentionally unchanged.
- `TableInstalledSkills` mirrors every installed-card affordance: skill name with `reference` subtitle when it differs from `metadata.name`, scope column with `UserIcon` / `FolderGit2Icon`, destination column that shows `SkillClientsBadges` for user-scoped installs and a `project_root` chip for repo-scoped ones (both when applicable), status badge with the same variant map, and the `Uninstall` button that opens `DialogUninstallSkill`.
- Extract `SkillClientsBadges` so the `+N` overflow tooltip behaves identically in cards and rows, and lift `skillStatusVariantMap` into `skill-status.ts` to keep `card-skill.tsx` compatible with `react-refresh/only-export-components`.

### Tests

- Unit tests for `useViewPreference` (default, persisted read, optimistic write), `ViewToggle` (aria + onChange), `TableMcpServers` (header, row rendering, status, update-button visibility, kebab menu), `GridCardsMcpServers` (toggle persistence + default), and `TableInstalledSkills` (columns, client badges, project root, status, uninstall flow).
- Updated `main/src/ipc-handlers/__tests__/index.test.ts` to mock the new `ui-preferences` registrar so it doesn't transitively pull `main/src/logger.ts` (`app.isPackaged`) in a non-Electron test environment.

- The table view on the MCP servers page is per-installed-servers only — the Registry view still renders as cards.
- No sort/filter by column in the tables yet; the existing search input + status filter still drive row visibility.
- The About column hides below `md` (768px). If that breakpoint feels too aggressive for sidebar layouts, it's a one-line swap to `lg`.
- No automatic migration for users who already had a non-persisted mental default — first load in either page lands on the card view until they touch the toggle.

### DIFF
--- a/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
+++ b/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
@@ -14,6 +14,7 @@ import {
   TooltipTrigger,
 } from '@/common/components/ui/tooltip'
 import { Switch } from '@/common/components/ui/switch'
+import { Button } from '@/common/components/ui/button'
 import { cn } from '@/common/lib/utils'
 import { trackEvent } from '@/common/lib/analytics'
 import { ArrowUpCircle, CloudIcon, LaptopIcon } from 'lucide-react'
@@ -53,19 +54,19 @@ function UpdateVersionButton({
   return (
     <Tooltip>
       <TooltipTrigger asChild>
-        <button
+        <Button
+          variant="ghost"
+          size="icon"
+          className="size-8"
           onClick={(e) => {
             e.stopPropagation()
             void promptUpdate('card_button')
           }}
           disabled={disabled || !isReady}
-          className="hover:bg-accent inline-flex size-8 cursor-pointer
-            items-center justify-center rounded-md disabled:pointer-events-none
-            disabled:opacity-50"
           aria-label={`Update to ${drift.registryTag}`}
         >
           <ArrowUpCircle className="size-5 text-amber-500" />
-        </button>
+        </Button>
       </TooltipTrigger>
       <TooltipContent className="max-w-xs">
         Update available: {drift.localTag} → {drift.registryTag}
--- a/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
+++ b/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
@@ -1,0 +1,266 @@
+import type { GithubComStacklokToolhivePkgCoreWorkload as CoreWorkload } from '@common/api/generated/types.gen'
+import type { RegistryEnvVar } from '@common/api/registry-types'
+import {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@/common/components/ui/table'
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/common/components/ui/tooltip'
+import { Switch } from '@/common/components/ui/switch'
+import { cn } from '@/common/lib/utils'
+import { trackEvent } from '@/common/lib/analytics'
+import { ArrowUpCircle, CloudIcon, LaptopIcon } from 'lucide-react'
+import { useIsServerFromRegistry } from '../hooks/use-is-server-from-registry'
+import { useMutationRestartServer } from '../hooks/use-mutation-restart-server'
+import { useMutationStopServerList } from '../hooks/use-mutation-stop-server'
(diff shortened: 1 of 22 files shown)


renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
```diff
@@ -0,0 +1,266 @@
...
+    registryImage,
+    localTag: drift.localTag,
+    registryTag: drift.registryTag,
+    registryEnvVars,
+  })
+
+  return (
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <button
```
peppescg: it should be the button from ui component

renderer/src/features/skills/components/skill-clients-badges.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {visibleClients.map((client) => (
+        <Badge key={client} variant="outline">
+          {client}
+        </Badge>
+      ))}
+      {hiddenClients.length > 0 && (
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <Badge asChild variant="outline">
+              <button
```
peppescg: same above

renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
samuv: yeah pushed

peppescg (approved):

PR #2047, opened 2026-04-20
feat(ui): add card/table view toggle for registry and skills tabs

samuv:
The MCP Registry, Skills Registry, and Skills Local Builds screens only offered a card grid. This PR extends the card/table view toggle already available on MCP Servers and Installed Skills to those three surfaces, and retrofits the two pre-existing tables with a dynamic Name column.

https://github.com/user-attachments/assets/a7528c27-6000-4f2f-93b5-4dcf4c0b0041

- Extend `UI_PREFERENCE_KEYS` with `ui.viewMode.registry`, `ui.viewMode.skillsRegistry`, and `ui.viewMode.skillsBuilds` so each surface persists its own card/table preference via the settings table. `useViewPreference` and the `ViewToggle` primitive are reused without changes.
- Add `TableRegistry` for the MCP Registry page. It renders server and group entries in a single table (Name, About, Type, Stars, repository link), slots the `CardRegistryPromo` at index 6 as a full-width `colSpan` row when the default registry is active, and supports row-level navigation via click and keyboard `Enter` with `stopPropagation` on the inline GitHub link. Wire it into `-registry.route.tsx` next to the existing `GridCardsRegistry`.
- Add `TableRegistrySkills` for the Skills Registry tab. Columns are Skill, Author (namespace, shown on `lg+`), About, and an Install action. Rows with both a namespace and a name navigate to `/skills/$namespace/$skillName`; the Install button generates the correct `defaultReference` for OCI vs Git packages and stops propagation. Wire it into the Registry tab in `skills-page.tsx`.
- Add `TableBuilds` for the Skills Local Builds tab. Columns are Build (title + tag subtitle), Version (shows a `latest` badge when `build.version` is unset), Digest (truncated with tooltip, hidden below `lg`), About, and Remove + Install actions. Runs the same `getApiV1BetaSkillsBuildsOptions` query as `GridCardsBuilds` with local filtering, and exposes an empty state with a `Build skill` CTA. Wire it into the Local Builds tab in `skills-page.tsx`.
- Make the Name/Skill/Build column dynamic across **every** table (MCP Servers, Installed Skills, MCP Registry, Skills Registry, Local Builds). The column now shrinks to the longest content up to a `max-w-[260/280px]` cap instead of a fixed percentage, using `table-auto` layout with a truncating inner span. The flexible column (About or Destination) uses the classic `w-full max-w-0` trick so its content truncates against the remaining space.
- Drop the `overflow-hidden` wrapper that was clipping vertical scroll on the MCP Registry page when the table grew past the viewport. With the auto layout the tables no longer overflow horizontally, so the default `overflow-x-auto` on the shadcn `Table` primitive is sufficient.

- The toolbar on the Registry page does not gain a registry selector or a `+ Create` button shown in the mockups; the toggle replaces only the grid/list affordance.
- The columns on each table mirror the *full* card content (all badges, kebab items, conditional controls) rather than strictly matching the Figma column list, so nothing is lost when switching from cards.

### Tests

- New unit tests for `TableRegistry`, `TableRegistrySkills`, and `TableBuilds` covering column headers, row rendering, promo placement, dialog wiring, `stopPropagation` on inline actions, and click/`Enter` navigation.
- Existing `TableMcpServers` and `TableInstalledSkills` tests continue to pass after the dynamic-column refactor.
- `pnpm run lint`, `pnpm run type-check`, and the full Vitest suite are green.

### DIFF
--- a/main/src/ui-preferences.ts
+++ b/main/src/ui-preferences.ts
@@ -5,6 +5,9 @@ import { writeSetting } from './db/writers/settings-writer'
 export const UI_PREFERENCE_KEYS = [
   'ui.viewMode.mcpServers',
   'ui.viewMode.skillsInstalled',
+  'ui.viewMode.registry',
+  'ui.viewMode.skillsRegistry',
+  'ui.viewMode.skillsBuilds',
 ] as const
 
 export type UiPreferenceKey = (typeof UI_PREFERENCE_KEYS)[number]
--- a/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
+++ b/renderer/src/features/mcp-servers/components/table-mcp-servers.tsx
@@ -135,7 +135,7 @@ function McpServerRow({ mcpServer }: { mcpServer: CoreWorkload }) {
           <TooltipTrigger asChild>
             <span
               className={cn(
-                'block truncate',
+                'block max-w-[260px] truncate',
                 isStopped && 'text-foreground/65'
               )}
             >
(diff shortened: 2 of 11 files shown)


peppescg (approved):
🚢

PR #2063, opened 2026-04-21
feat(skills): explain User vs Project scope in install dialog

samuv:
The install skill dialog exposes a `Scope` select with `User` / `Project` options, but never explains the difference — users have to guess where the skill will land on disk, and the extra `Project root` input only appears after they pick `Project`. This PR adds a hover hint so the choice is self-describing.

Screenshot 2026-04-21 at 18 33 29

- Add a `TooltipInfoIcon` next to the `Scope` label in `DialogInstallSkill` (`renderer/src/features/skills/components/dialog-install-skill.tsx`), reusing the same primitive already used on `form-fields-proxy.tsx`, `form-fields-base.tsx`, `form-fields-auth.tsx`, and the registry `configuration-tab-content.tsx`, so field-level help stays visually consistent across the app.
- Tooltip copy:
  - **User** — installed globally in your home directory and available across all projects.
  - **Project** — installed in the root of a specific project folder and scoped to that project only.
- Wrap the `FormLabel` + icon in a `flex items-center gap-1` container — same pattern used elsewhere — so the `(i)` icon sits inline with the label without disturbing the rest of the form layout.

### DIFF
--- a/renderer/src/common/components/ui/tooltip-info-icon.tsx
+++ b/renderer/src/common/components/ui/tooltip-info-icon.tsx
@@ -1,22 +1,27 @@
-import { TooltipTrigger } from '@radix-ui/react-tooltip'
-import { Tooltip, TooltipContent } from './tooltip'
+import { Tooltip, TooltipContent, TooltipTrigger } from './tooltip'
 import { InfoIcon } from 'lucide-react'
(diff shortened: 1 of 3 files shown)


peppescg (approved):
Makes totally sense

PR #2064, opened 2026-04-21
feat(skills): surface source repository and tighten registry metadata

samuv:
Following [#2045](https://github.com/stacklok/toolhive-studio/pull/2045), this PR brings the skills registry closer to parity with MCP servers: the upstream repo is one click away from every view, the `SKILL.md` preview drops the metadata we already render as badges, and the detail page stops resizing between skills.

The registry response distinguishes between `namespace` (packager, e.g. `io.github.stacklok`) and `repository.url` (upstream source, e.g. `github.com/trailofbits/skills`). Today only `namespace` is visible — it's not obvious the skill was authored elsewhere and packaged downstream. MCP servers already solve this via `repository_url`; this PR lines the two up.

## Changes

- **Card / table / detail** — link `skill.repository.url` with the same GitHub-icon pattern used for MCP servers: ghost icon on the card footer, an icon-only trailing column on the registry table, and an outline `GitHub` button next to `Install` on the detail page. All three render nothing when `repository` is absent, and their click handlers stop propagation so card/row navigation isn't triggered.
- **Table column rename** — `Author` → `Registry`. The column holds the reverse-DNS namespace (the packager/registry), not the author.
- **SKILL.md preview** — enable `stripFrontmatter` on the registry detail. `SkillMarkdown` already supported it; only the build detail was passing it. Preview now starts at the first heading instead of the YAML block duplicating `name` / `description` / `license` / `version`.
- **Detail layout** — the right panel was `flex-1` without `min-w-0`, so wide SKILL.md content (code blocks, tables) pushed the column past its 7/12 allotment and varied between skills. Adding `min-w-0` lets the flex child shrink; overflow now scrolls inside the card. Applies to both registry and build detail pages (shared layout).

## How to validate

1. Registry tab — card footer has a GitHub icon; opens the upstream repo in a new tab, card stays in place.
2. Table view — `Registry` header (previously `Author`); GitHub icon sits between `About` and `Install`.
3. Skill detail (e.g. `io.github.stacklok/agentic-actions-auditor`) — `GitHub` button next to `Install`; SKILL.md starts at the first heading; card width is identical across skills.

### DIFF
--- a/renderer/src/features/skills/components/__tests__/card-registry-skill.test.tsx
+++ b/renderer/src/features/skills/components/__tests__/card-registry-skill.test.tsx
@@ -105,6 +105,32 @@ describe('CardRegistrySkill', () => {
       renderRoute(router)
       expect(screen.getByRole('button', { name: /install/i })).toBeVisible()
     })
+
+    it('does not render a GitHub link when repository is absent', () => {
+      renderRoute(router)
+      expect(
+        screen.queryByRole('link', { name: /open repository on github/i })
+      ).not.toBeInTheDocument()
+    })
+
+    it('renders a GitHub link when repository.url is set', async () => {
+      const skillRouter = createCardTestRouter({
+        ...baseSkill,
+        repository: {
+          type: 'git',
+          url: 'https://github.com/example/skills',
+        },
+      }) as unknown as ReturnType<typeof createTestRouter>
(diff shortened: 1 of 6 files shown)


peppescg (approved):
🚢

PR #2084, opened 2026-04-23
ci: skip macOS code signing on PR validation builds

JAORMX:
## Summary

- The PR validation matrix (`on-pr.yml` → `_build-matrix.yml`) ran macOS code signing on every PR, which fails for forks because `APPLE_*` secrets aren't exposed to forked PR workflows. The `security import` step errored with `SecKeychainItemImport: One or more parameters passed to a function were not valid` ([example failing job](https://github.com/stacklok/toolhive-studio/actions/runs/24796832191/job/72573277517?pr=2080)).
- Signing belongs in release builds (`on-release.yml`) and the opt-in `/build-test` flow (`pr-build-test.yml`, which already blocks forks). A per-PR sanity build doesn't need signed artifacts.
- Drop the `setup-macos-codesign` step from `_build-matrix.yml` and guard `osxSign` in `forge.config.ts` so the build cleanly skips signing when no credentials are present, instead of asking `electron-osx-sign` to auto-detect an identity that doesn't exist on a clean runner.

## Test plan

- [ ] CI passes on this PR (macOS matrix should build without attempting to sign).
- [ ] Confirm next release build (`on-release.yml`) still signs + notarizes — `APPLE_API_KEY` set by `setup-macos-codesign` keeps `osxSign` enabled.
- [ ] Confirm `/build-test` flow still signs macOS — same path as release.
- [ ] Local `pnpm run make` without Apple creds no longer tries to auto-detect an identity.

### DIFF
--- a/.github/workflows/_build-matrix.yml
+++ b/.github/workflows/_build-matrix.yml
@@ -31,17 +31,11 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
-      - name: Setup macOS code signing
-        uses: ./.github/actions/setup-macos-codesign
-        if: runner.os == 'macOS'
-        with:
-          apple-certificate: ${{ secrets.APPLE_CERTIFICATE }}
-          apple-certificate-password: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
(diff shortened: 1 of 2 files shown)


samuv (approved):
🫡

PR #2086, opened 2026-04-23
refactor(mcp-optimizer): remove the feature and migrate legacy users on startup

### LINKED ISSUE #2087: [Chore] Remove MCP Optimizer from the UI
## Goal

Remove the MCP Optimizer feature from the ToolHive Studio UI and ensure that users upgrading from a version where the optimizer was enabled are migrated cleanly on startup, so they don't end up with an orphaned `__mcp-optimizer__` group, a dangling `meta-mcp` workload, or clients registered to a group that no longer has a UI surface.

This follows #1910, which shipped the deprecation banner and prepared users for the sunset.

## Why

The MCP Optimizer was an experimental feature that has been deprecated. Keeping the code, routes, hooks, and UI branches around adds maintenance cost, complicates the MCP servers / groups / clients flows with optimizer-specific branches, and leaves users who had previously enabled it in an inconsistent state after upgrade.

### Remove from the product
- Route + UI surface: `/mcp-optimizer`, the sunset banner (`McpOptimizerSunsetBanner`), the entire `features/meta-mcp` module (group selector, loading dialog, optimizer warnings, and backing hooks).
- All optimizer-specific branches in the MCP servers sidebar, group actions, network isolation tab, local-mcp form, clients button, and the registry / run-custom / run-remote / update / restart / stop / delete orchestration.
- Supporting hooks: `use-create-optimizer-group`, `use-optimize-group-name`, `use-is-optimized-group-name`, `use-mcp-optimizer-banner-visible`, `use-notification-optimizer`, and their tests.
- Config: drop the `toolhive-mcp-optimizer` image from `renovate.json`, drop `MCP_OPTIMIZER_SUNSET_BLOG_URL` from `common/app-info.ts`, trim `constants.ts` down to the identifiers still needed by the cleanup migration.

### Keep for one release
- `META_OPTIMIZER` remains in the feature-flag registry but set to `isDisabled: true`. It never shows up in experimental settings and always returns `false`, but the key is preserved so the cleanup path can observe prior state. To be removed entirely in the next release.
- `MCP_OPTIMIZER_GROUP_NAME` / `META_MCP_SERVER_NAME` constants stay in `constants.ts` because the cleanup hook depends on them. They can be inlined or deleted once the `META_OPTIMIZER` flag is removed.

### Startup migration
- A new `useMcpOptimizerStartupCleanup` hook runs at most once per session from `RootComponent`. It awaits `electronAPI.isToolhiveRunning()` (polling up to 60s) before marking itself as run, then invokes `cleanupMetaOptimizer()`.
- `useCleanupMetaOptimizer` is rewritten so the cleanup is self-contained and callable from outside a component. It fetches groups and the `toolhive-mcp-optimizer` workload at call-time via `queryClient.fetchQuery`, then:
  1. If the meta-mcp workload has `ALLOWED_GROUPS` pointing at a user group that still exists, re-registers the optimizer-group clients to that target group.
  2. Deletes the `__mcp-optimizer__` group with `?with-workloads=true`, taking the meta-mcp workload with it.
- Cleanup is a no-op when the optimizer group doesn't exist, so users who never enabled the feature pay nothing at startup.

The cleanup path is intentionally the same teardown that used to run when the user flipped the experimental toggle off — nothing new is introduced, only relocated so it now fires automatically on first launch post-upgrade.

## Acceptance criteria

- [ ] `/mcp-optimizer` route, the sunset banner, and all optimizer-related UI affordances are gone.
- [ ] On a profile with the optimizer previously enabled, launching the app removes the `__mcp-optimizer__` group, re-registers its clients to whichever group was in `ALLOWED_GROUPS`, and tears down the meta-mcp workload — before any user interaction.
- [ ] On a fresh profile with no optimizer state, the startup hook short-circuits after the daemon-ready wait and never writes to the API.
- [ ] Relaunching the app is a no-op (the session flag short-circuits, and the group no longer exists either way).
- [ ] No main-process changes beyond the feature-flag flip — cleanup runs entirely in the renderer against the public `thv` HTTP API, mirroring the experimental-toggle-off path.
- [ ] Unit tests updated: `use-cleanup-meta-optimizer.test.ts` matches the call-time fetch model, `flags.test.ts` asserts `isDisabled` flags always return `false`, and the obsolete optimizer-specific permission-profile case is removed from `orchestrate-run-local-server.test.tsx`.
- [ ] E2E coverage: `e2e-tests/mcp-optimizer-startup-cleanup.spec.ts` (backed by a reusable `launchApp` + `thvFetch` helper) runs two sessions against the same temporary `userDataDir` and asserts the optimizer group is gone, the client is re-registered to the custom group, and `GET /workloads/toolhive-mcp-optimizer` returns 404.

## References

- PR: #2086
- Previous deprecation work: #1910

samuv:
Follows #1910, which shipped the deprecation banner and prepared users for the sunset. This PR removes the MCP Optimizer feature from the product and adds a one-time, idempotent startup cleanup so that users upgrading from a version where the optimizer was enabled don't end up with an orphaned `__mcp-optimizer__` group, a dangling `meta-mcp` workload, or clients registered to a group that no longer has a UI surface.

The cleanup path is intentionally the same teardown that used to run when the user flipped the experimental toggle off — nothing new is introduced, only relocated so it now fires automatically on first launch post-upgrade.

## What's removed

- **Route + UI surface** — `/mcp-optimizer`, the sunset banner (`McpOptimizerSunsetBanner`), the entire `features/meta-mcp` module (group selector, loading dialog, optimizer warnings, and the hooks that backed them), and all optimizer-specific branches in the MCP servers sidebar, group actions, network isolation tab, local-mcp form, clients button, and registry / run-custom / run-remote / update / restart / stop / delete orchestration.
- **Supporting hooks** — `use-create-optimizer-group`, `use-optimize-group-name`, `use-is-optimized-group-name`, `use-mcp-optimizer-banner-visible`, `use-notification-optimizer`, and the associated tests.
- **Config** — the `toolhive-mcp-optimizer` image is no longer tracked in `renovate.json`, the `MCP_OPTIMIZER_SUNSET_BLOG_URL` constant is dropped from `common/app-info.ts`, and `constants.ts` is trimmed down to the identifiers still needed by the cleanup migration (`MCP_OPTIMIZER_GROUP_NAME`, `META_MCP_SERVER_NAME`, etc.).

## What's kept for one release

- `META_OPTIMIZER` remains in the feature-flag registry but is now `isDisabled: true`. It never shows up in the experimental settings UI and always returns `false`, but the key is preserved so the cleanup path can run and observe prior state. It can be deleted entirely in the next release.

## Startup migration

- `useMcpOptimizerStartupCleanup` is a new hook wired into `RootComponent`. It runs at most once per session, `await`s `electronAPI.isToolhiveRunning()` (polling up to 60s) before marking itself as run, and then invokes `cleanupMetaOptimizer()`. Marking as run *after* the readiness check is important — otherwise a slow daemon start would cause the migration to silently no-op.
- `useCleanupMetaOptimizer` was rewritten so the cleanup is self-contained and callable from outside a component that had rendered the relevant `useQuery`s. It now fetches groups and the `toolhive-mcp-optimizer` workload detail at call-time via `queryClient.fetchQuery`, then:
  1. If the meta-mcp workload has `ALLOWED_GROUPS` pointing at a user group that still exists, it re-registers the optimizer-group clients to that target group.
  2. Deletes the `__mcp-optimizer__` group with `?with-workloads=true`, taking the meta-mcp workload with it.

The cleanup is a no-op when the optimizer group doesn't exist, so users who never enabled the feature pay nothing at startup.

## Tests

- **Unit** — updated `use-cleanup-meta-optimizer.test.ts` to match the call-time fetch model (the hook no longer makes requests on mount), added a `flags.test.ts` case asserting `isDisabled` flags always return `false`, and removed the obsolete optimizer-specific permission-profile case from `orchestrate-run-local-server.test.tsx`.
- **E2E** — new `e2e-tests/mcp-optimizer-startup-cleanup.spec.ts`, backed by a reusable `launchApp` + `thvFetch` helper in `e2e-tests/helpers/app-relaunch.ts`. It runs two sessions against the same temporary `userDataDir`:
  1. Session 1 creates a custom group via the UI, then seeds `__mcp-optimizer__`, registers a `vscode` client against it, and creates a `meta-mcp` workload pointing at a local test MCP server with `ALLOWED_GROUPS` set to the custom group.
  2. Session 2 polls the `thv` HTTP API until the optimizer group is gone, the client is re-registered to the custom group, and `GET /workloads/toolhive-mcp-optimizer` returns 404 — with a best-effort `thv group rm --with-workloads` in `afterAll` as a safety net.

## How to validate

1. On a profile with the optimizer previously enabled (or one seeded with the group + meta-mcp workload via `thv`), launch the app. The `__mcp-optimizer__` group disappears, its clients are re-registered to whichever group was in `ALLOWED_GROUPS`, and the meta-mcp workload is torn down — all before any user interaction.
2. Relaunch the app. The cleanup is a no-op (the session flag short-circuits, and the group no longer exists either way).
3. On a fresh profile with no optimizer state, the startup hook short-circuits after the daemon-ready wait and never writes to the API.
4. The `/mcp-optimizer` route, the sunset banner, and all optimizer-related UI affordances are gone.

## Not changed in this PR (deliberate)

- No main-process changes beyond the feature flag flip — the cleanup runs entirely in the renderer against the public `thv` HTTP API, mirroring the experimental-toggle-off path.
- `MCP_OPTIMIZER_GROUP_NAME` / `META_MCP_SERVER_NAME` constants stay in `constants.ts` for one release because the cleanup hook depends on them. They can be inlined or deleted once the `META_OPTIMIZER` flag is removed.


peppescg (approved):
🔥

PR #2093, opened 2026-04-23
chore: add devcontainer + noVNC setup for parallel/agentic dev servers

kantord:
## What this is

A self-contained dev environment for ToolHive Studio that runs the entire app —
Electron, the bundled `thv` backend, MCP-server containers, secrets, the
assistant — inside a single devcontainer, accessed in the browser via noVNC.

`pnpm devContainer:dev` builds the container, starts the display stack
(Xvfb + fluxbox + x11vnc + websockify/noVNC), runs `pnpm start`, and opens the
host browser at `http://localhost:6080/vnc.html` once everything is ready.
Same path inside GitHub Codespaces — open a Codespace and the preview pane
auto-opens with the running app.

Works on Linux, macOS, and any cloud environment that supports devcontainers
with the `docker-in-docker` feature (Codespaces, GitHub Actions, etc.).

## Why it's useful

**For humans:**
- Run dev servers in isolation without polluting host MCP config, host pnpm
  installs, host X11/GPU, or the host Docker socket.
- Multiple git worktrees can each run their own instance simultaneously
  (per-worktree container labels + named `node_modules` volumes + dynamic
  host port).
- Full cloud development via Codespaces — share a running instance via the
  forwarded URL.

**For agents:**
- The container has `xdotool` + ImageMagick `import`, so agents can take
  screenshots and drive mouse/keyboard against the live Electron window —
  including things Playwright can't reach (file-open dialogs, native menus,
  modals). Confirmed working: an agent can navigate to a registry detail
  page, install a server, and read the logs from screenshots alone.
- Each worktree's container is fully isolated, so multiple agents can run
  in parallel without stepping on each other's MCP config or app state.
- Same setup should host CDP (`--remote-debugging-port`) for DOM-level
  automation, or chrome-devtools-mcp inside the container — not wired up
  yet, but the wiring point is documented.

## Implementation

Three scripts, roles documented in `.claude/skills/devcontainer-dev/SKILL.md`:

- `scripts/devcontainer-dev.sh` (host) — picks a host port, runs
  `devcontainer up`, polls readiness, opens browser.
- `scripts/devcontainer-entrypoint.sh` (container) — starts the display
  stack + DBus + gnome-keyring (needed by ToolHive's secret API), runs
  `pnpm start`.
- `scripts/devcontainer-post-start.sh` (container) — in Codespaces,
  auto-launches the entrypoint on container start.

Container runtime for `thv` is provided by the standard
`docker-in-docker:2` devcontainer feature — no host Docker socket
mount, works identically on Linux/macOS/Codespaces.

The skill file documents the architecture, gotchas (non-stock Linux
kernels, blank-white-page fix, fluxbox WM_CLASS rule, pgrep self-match
trick), and the agent-driving patterns in detail.

## Test plan

Confirmed working:
- [x] Local Linux: `pnpm devContainer:dev` opens browser to a working app
- [x] GitHub Codespaces: preview pane auto-opens, full feature parity
      (secret storage, MCP install, assistant chat)
- [x] Multiple worktrees running simultaneously without port/volume collisions
- [x] Agent screenshot + xdotool flow against the live Electron window

Should work, untested:
- [ ] Local macOS (Docker Desktop / OrbStack)

## Future extensions

- Same approach should work for the enterprise repo / full kind cluster, so
  a cloud agent could work across the whole ecosystem (BE, cloud-ui,
  enterprise desktop).
- CDP / chrome-devtools-mcp wiring for DOM-level agent automation.
- Native-feeling Linux path via `cage` (Wayland kiosk) instead of VNC, if
  the perf overhead becomes a problem locally.


peppescg:
I don't want to block you, so we can merge the pr and we can test it later? cause no issues on UI side it is only CI. wdty?

peppescg (approved):

PR #2101, opened 2026-04-24
feat(playground): agent configuration and Skills Builder agent

samuv:
Extends the Playground with per-agent configuration and ships a new Skills Builder built-in agent that can scaffold and build an MCP skill end-to-end from chat. The new UI surfaces are gated behind a DevTools-only `agents` feature flag so external users see the same chat as before while internal QA can flip it on.

## What's new

- **`agents` feature flag** (DevTools-only, default off): hides the Agents sidebar entry and the chat composer's agent picker. The `ToolLoopAgent` runtime stays active and falls through to the ToolHive Assistant prompt, so chat behaviour with the flag off matches `main`. Toggle with `FeatureFlag.agents.enable()` / `.disable()` in DevTools.
- **Agents section** under `/playground/agents`: list, detail, create, and edit pages. Built-in agents are read-only with a **Duplicate** action; custom agents support name, description, default model, instructions (raw textarea or markdown preview), and an optional built-in tool bundle.
- **Agent runtime**: chat now runs through an ai-sdk `ToolLoopAgent` resolved per thread. Each agent carries its own system prompt, optional default model, and optional built-in tool bundle; user-enabled MCP tools continue to layer on top. With no UI to assign an agent, threads keep `agent_id = null` and resolve to `builtin.toolhive-assistant` (same prompt as before).
- **Agent picker in chat settings** (new thread + sidebar): choose which agent powers the thread alongside the existing provider / MCP pickers. Hidden when the flag is off.
- **Skills Builder built-in agent**: drives a scripted flow (`write_skill_files` → `build_skill`) that writes a skill into a temp workdir and builds it into a local OCI artifact via the ToolHive API. Its system prompt is tuned to keep post-build replies to 1–2 sentences so the install CTA stays visible.
- **"Skill built" card** in chat: dedicated tool-output renderer for `build_skill` showing name / version / tag / digest with **Install**, **View details**, and **Copy name** actions. Install dialog is split into separate Name and Version fields so local builds install cleanly.
- **Reusable `ModelPicker`**: the model dropdown used by chat settings is now shared with the agent form so both UIs behave identically.

## How to validate

**Flag off (default — external user behaviour):**

1. Playground sidebar shows **no** Agents entry. Chat composer toolbar has Model + MCP buttons but **no** agent picker.
2. Send a message in any thread → response streams through the ToolHive Assistant prompt as it did pre-PR.

**Flag on (`FeatureFlag.agents.enable()` in DevTools):**

1. Playground → **Agents** lists ToolHive Assistant + Skills Builder. **Duplicate** a built-in → edit the copy → it appears in the list and can be selected on a new chat.
2. New chat → settings → pick **Skills Builder** → ask it to build a skill. After `build_skill` succeeds, the "Skill built" card appears with **Install** (opens the dialog with Name + Version prefilled), **View details** (navigates to the build detail page), and **Copy name**.
3. Agent detail → **Back** always returns to the agents list.
4. Install a skill without an explicit version → Version field stays empty and defaults to `latest` on submit.

### DIFF
--- a/main/src/chat/agents/builtin-agent-tools/index.ts
+++ b/main/src/chat/agents/builtin-agent-tools/index.ts
@@ -2,7 +2,7 @@ import type { ToolSet } from 'ai'
 import type { BuiltinToolsKey } from '../types'
 import { createSkillsAgentTools, type SkillsAgentToolsHandle } from './skills'
 
-export interface BuiltinToolsHandle {
+interface BuiltinToolsHandle {
   tools: ToolSet
   cleanup: () => Promise<void>
 }
--- a/main/src/chat/agents/builtin-agent-tools/index.ts
+++ b/main/src/chat/agents/builtin-agent-tools/index.ts
@@ -1,0 +1,31 @@
+import type { ToolSet } from 'ai'
+import type { BuiltinToolsKey } from '../types'
+import { createSkillsAgentTools, type SkillsAgentToolsHandle } from './skills'
+
+export interface BuiltinToolsHandle {
+  tools: ToolSet
+  cleanup: () => Promise<void>
+}
+
+const EMPTY_HANDLE: BuiltinToolsHandle = {
+  tools: {},
+  cleanup: async () => {},
+}
+
+export function createBuiltinAgentTools(
+  key: BuiltinToolsKey | null | undefined
+): BuiltinToolsHandle {
+  if (!key) return EMPTY_HANDLE
+
+  switch (key) {
+    case 'skills': {
+      const handle: SkillsAgentToolsHandle = createSkillsAgentTools()
+      return handle
+    }
+    case 'planner':
+      // Planner is instructions-only for now; reserved for future extension.
+      return EMPTY_HANDLE
+    default:
+      return EMPTY_HANDLE
+  }
+}
--- a/main/src/chat/agents/builtin-agent-tools/index.ts
+++ b/main/src/chat/agents/builtin-agent-tools/index.ts
@@ -22,9 +22,6 @@ export function createBuiltinAgentTools(
       const handle: SkillsAgentToolsHandle = createSkillsAgentTools()
       return handle
     }
-    case 'planner':
-      // Planner is instructions-only for now; reserved for future extension.
-      return EMPTY_HANDLE
     default:
       return EMPTY_HANDLE
   }
--- a/main/src/chat/agents/builtin-agent-tools/skills.ts
+++ b/main/src/chat/agents/builtin-agent-tools/skills.ts
@@ -207,7 +207,7 @@ export function createSkillsAgentTools(): SkillsAgentToolsHandle {
           }
 
           const effectiveBuild: LocalBuild = build ?? {
-            tag: tag ?? apiReference,
+            tag: apiReference,
             ...(tag ? { version: tag } : {}),
           }
 
--- a/main/src/chat/agents/builtin-agent-tools/skills.ts
+++ b/main/src/chat/agents/builtin-agent-tools/skills.ts
@@ -1,0 +1,253 @@
+import path from 'node:path'
+import fs from 'node:fs/promises'
+import { app } from 'electron'
+import { nanoid } from 'nanoid'
+import { z } from 'zod'
+import { tool, type ToolSet } from 'ai'
+import log from '../../../logger'
+import { createClient } from '@common/api/generated/client'
+import {
+  getApiV1BetaSkillsBuilds,
+  postApiV1BetaSkillsBuild,
+} from '@common/api/generated/sdk.gen'
+import type { GithubComStacklokToolhivePkgSkillsLocalBuild as LocalBuild } from '@common/api/generated/types.gen'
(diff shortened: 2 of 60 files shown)


main/src/chat/agents/builtin-agent-tools/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!key) return EMPTY_HANDLE
+
+  switch (key) {
+    case 'skills': {
+      const handle: SkillsAgentToolsHandle = createSkillsAgentTools()
+      return handle
+    }
+    default:
+      return EMPTY_HANDLE
+  }
```
peppescg, line 27: a switch with one case 😆

peppescg:
Nice PR! One small suggestion:

**Sidebar wiring duplication across agent routes**

The `usePlaygroundThreads(null)` + handler wiring + `<PlaygroundSidebar>` block is duplicated almost identically in three route files:
- `playground.agents.tsx`
- `playground.agents_.$agentId.tsx`
- `playground.agents_.$agentId_.edit.tsx`

Since `playground.agents.tsx` is already a child of the `/playground` layout, it could itself become a **layout route** for all `/playground/agents/*` paths — own the sidebar once and render an `<Outlet />` for the child pages:

```tsx
// playground.agents.tsx — layout route
function PlaygroundAgentsLayout() {
  const navigate = useNavigate()
  const { threads, hasThreads, createThread, deleteThread, renameThread, toggleStarThread } =
    usePlaygroundThreads(null)

  // ... shared handlers ...

  return (
    <div className="absolute inset-0 flex">
      {hasThreads && (
        <PlaygroundSidebar
          threads={threads}
          activeThreadId={null}
          onSelectThread={handleSelectThread}
          onCreateThread={...}
          onDeleteThread={...}
          onRenameThread={renameThread}
          onToggleStar={toggleStarThread}
        />
      )}
      <div className="flex min-w-0 flex-1 flex-col overflow-y-auto">
        <Outlet />
      </div>
    </div>
  )
}
```

Then the child routes (`$agentId.tsx`, `$agentId_.edit.tsx`, `new.tsx`) drop the sidebar entirely and only render their page-specific content. Saves ~60 lines of duplication and makes future agent sub-pages trivial to add.

Not blocking — with 3 routes the copy-paste is manageable.

main/src/chat/agents/builtin-agent-tools/index.ts
samuv: I mean, at the beginning I added more agents, but they weren’t really useful. I need to think about it more,  that’s why it’s currently a single item in a switch.

samuv:
> Nice PR! One small suggestion:
> 
> **Sidebar wiring duplication across agent routes**
> 
> The `usePlaygroundThreads(null)` + handler wiring + `<PlaygroundSidebar>` block is duplicated almost identically in three route files:
> 
> * `playground.agents.tsx`
> * `playground.agents_.$agentId.tsx`
> * `playground.agents_.$agentId_.edit.tsx`
> 
> Since `playground.agents.tsx` is already a child of the `/playground` layout, it could itself become a **layout route** for all `/playground/agents/*` paths — own the sidebar once and render an `<Outlet />` for the child pages:
> 
> ```tsx
> // playground.agents.tsx — layout route
> function PlaygroundAgentsLayout() {
>   const navigate = useNavigate()
>   const { threads, hasThreads, createThread, deleteThread, renameThread, toggleStarThread } =
>     usePlaygroundThreads(null)
> 
>   // ... shared handlers ...
> 
>   return (
>     <div className="absolute inset-0 flex">
>       {hasThreads && (
>         <PlaygroundSidebar
>           threads={threads}
>           activeThreadId={null}
>           onSelectThread={handleSelectThread}
>           onCreateThread={...}
>           onDeleteThread={...}
>           onRenameThread={renameThread}
>           onToggleStar={toggleStarThread}
>         />
>       )}
>       <div className="flex min-w-0 flex-1 flex-col overflow-y-auto">
>         <Outlet />
>       </div>
>     </div>
>   )
> }
> ```
> 
> Then the child routes (`$agentId.tsx`, `$agentId_.edit.tsx`, `new.tsx`) drop the sidebar entirely and only render their page-specific content. Saves ~60 lines of duplication and makes future agent sub-pages trivial to add.
> 
> Not blocking — with 3 routes the copy-paste is manageable.

yeah fair, I would handle it in another PR if it possible this one is too big

main/src/chat/agents/builtin-agent-tools/skills.ts
```diff
@@ -0,0 +1,253 @@
...
+  }
+  return written
+}
+
+export interface SkillsAgentToolsHandle {
+  tools: ToolSet
+  cleanup: () => Promise<void>
+}
+
+export function createSkillsAgentTools(): SkillsAgentToolsHandle {
```
peppescg, line 62: I would split the props that are big or more logic in a function with a comment for better readability

peppescg (approved):
Nice job love it 🚀 
no blockers only suggestions up to you if you want to handle them in case in a separate pr

PR #2102, opened 2026-04-24
feat(skills): registry card polish and persisted pagination state

samuv:
Follow-up polish to the Skills → Registry tab on top of #2097 / #2098.

https://github.com/user-attachments/assets/100c6ae8-57ba-4270-8635-9037ebd2d646

Summary

- Normalized repo label replaces the io.github.stacklok subheading on registry cards and the namespace badge on the skill detail page. GitHub URLs render as org/repo, other hosts as host/path. The namespace stays on props — only hidden from the UI. Table view is untouched.
- Pagination bar pinned to the viewport bottom so it stays anchored with 3 cards or 30. Overflowing grids still scroll as before.
- Last-page button (>|) added to the shared Pagination component, next to the existing next/first/prev controls.
- Items-per-page is remembered across sessions via a new ui.pageSize.skillsRegistry preference, wired through the same uiPreferences IPC path used for view-mode. URL ?limit= still wins over the stored value.

### DIFF
--- a/main/src/ipc-handlers/ui-preferences.ts
+++ b/main/src/ipc-handlers/ui-preferences.ts
@@ -1,8 +1,12 @@
 import { ipcMain } from 'electron'
 import {
+  getPageSizePreference,
   getViewModePreference,
+  setPageSizePreference,
   setViewModePreference,
+  UI_PAGE_SIZE_PREFERENCE_KEYS,
   UI_PREFERENCE_KEYS,
+  type UiPageSizeKey,
   type UiPreferenceKey,
   type ViewMode,
 } from '../ui-preferences'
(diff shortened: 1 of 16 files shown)


main/src/ipc-handlers/ui-preferences.ts
```diff
@@ -34,4 +46,18 @@ export function register() {
...
+    (_event, key: unknown): number | undefined =>
+      isUiPageSizeKey(key) ? getPageSizePreference(key) : undefined
+  )
+
+  ipcMain.handle(
+    'ui-preferences:set-page-size',
+    (_event, key: unknown, value: unknown): void => {
+      if (!isUiPageSizeKey(key) || !isValidPageSize(value)) return
+      setPageSizePreference(key, value)
+    }
```
peppescg, line 61: not a blocker only my 2c, it seems overkill saving the page size in DB, it is only an option of the endpoint

peppescg (approved):
🚀

main/src/ipc-handlers/ui-preferences.ts
samuv: We have the settings table that has all kinds of ui preference in there now, I wanted to be consistence and have it in one place only

PR #2104, opened 2026-04-25
fix: update Google playground model list

thisisryanswift:
## Summary
- Add Gemma 4 model IDs to the Google provider list in the playground
- Remove deprecated/shutdown Google model IDs from the playground picker

## Validation
- Validated model IDs against Gemini API docs
- Ran `pnpm run type-check` with Node 24.14.1

### DIFF
--- a/main/src/chat/constants.ts
+++ b/main/src/chat/constants.ts
@@ -110,7 +110,6 @@ export const CHAT_PROVIDER_INFO: ChatProviderInfo[] = [
       'gemini-3.1-flash-lite-preview',
 
       // Gemini 3
-      'gemini-3-pro-preview',
       'gemini-3-pro-image-preview',
       'gemini-3-flash-preview',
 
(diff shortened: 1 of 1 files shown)


samuv:
Thanks for the update @thisisryanswift! Quick question before we merge: have you actually run a chat against `gemma-4-31b-it` and `gemma-4-26b-a4b-it` end-to-end?

I checked `@ai-sdk/google@3.0.64` (the latest published version, also what we have locally) and these IDs aren't in the `GoogleGenerativeAIModelId` union yet — only Gemma 3 variants are listed (`gemma-3-1b-it`, `gemma-3-4b-it`, `gemma-3-12b-it`, `gemma-3-27b-it`, `gemma-3n-e2b-it`, `gemma-3n-e4b-it`). The union ends with `(string & {})` so it'll typecheck fine, but that's just the catch-all, it doesn't tell us the SDK has actually been updated to route Gemma 4 correctly.

Could you confirm you've smoke-tested both models from the playground? Just want to make sure they actually respond and aren't silently 404'ing against the Generative Language API endpoint the SDK is pointed at. I'll also try to verify on my side once I get access to a Google API key again.

thisisryanswift:
@samuv They're working for me locally! ai-sdk is probably out of date since they dropped in the last couple of weeks? 

image

image

samuv (approved):
Amazing, if they are working let's ship it then 🚢

Issue #663, opened 2025-07-25
[Bug] When the server deletion finishes it focuses the MCP servers tab


eleftherias:
### Bug Description

When I delete an MCP server and then switch to the **Registry** tab, the app takes me back to the **MCP Servers** tab when the deletion finishes.

### Steps to Reproduce

Start an MCP Server, for example fetch.
Stop the MCP server.
Quickly navigate to the Registry tab.

### Expected Behavior

I expect to remain on the Registry tab.

### Actual Behavior

After a few seconds I'm taken back to the MCP Servers tab.

### Priority

Medium

### Environment

macOS 
ToolHive Version 0.1.5

PR #2120, opened 2026-04-27
chore: add visual bugfix agent

kantord:
## TL;DR

Proves the project's devcontainer (Xvfb + fluxbox + Electron + thv) can boot inside a GitHub Actions runner and be driven end-to-end via `docker exec` — no AI agent involved here, just the substrate. Adds two helper scripts and corrects the `devcontainer-dev` skill so future agents can use this safely. Steady-state cost is ~4 min/run with the caching landed here, with a clear path to sub-minute via follow-up #2136.

## What lands in this PR

1. **`.github/workflows/experiment-devcontainer-proof.yml`** — push-triggered, branch-pinned proof workflow. Builds the devcontainer, boots the app, captures screenshots, drives one xdotool keystroke, runs `pnpm run type-check` inside the container, uploads screenshots as an artifact.
2. **`scripts/devcontainer-screenshot.sh [host-path]`** — auto-finds the container, runs `import` inside, streams the PNG out via `docker exec cat` (works around the tmpfs/`docker cp` bug — see below). Prints absolute host path on stdout.
3. **`scripts/devcontainer-steal.sh <container-id> <container-path> <host-path>`** — generic file extractor for any tmpfs path. Use for log files, generated artifacts, anything else that lives in `/tmp`. The screenshot helper calls this internally.
4. **`.claude/skills/devcontainer-dev/SKILL.md`** (and `.codex/`, `.cursor/` copies kept in sync) — replaces the broken `docker cp` example with the new helpers; documents the tmpfs gotcha; adds an "agent-grade readiness" note.
5. **`actions/cache` for the named `node_modules` volume** in the proof workflow — drops `pnpm install` from ~100s to ~9s on warm.

## Proof

CI run: <https://github.com/stacklok/toolhive-studio/actions/runs/25062176682>

Final-state timings (with all fixes applied, attempt #2 = warm cache):

| Stage | Time |
|---|---|
| Devcontainer build & start | 167s |
| Wait for app readiness (rendered UI) | 25s |
| Screenshots (×2) + xdotool driving | ~4s |
| Type-check inside container | 22s |
| **Total wall time** | **4m24s** |

Both screenshots in the artifact are valid 1920×1200 PNGs of the rendered ToolHive Studio UI.

## Bugs surfaced and fixed

- **`docker cp` is silently broken for `/tmp` paths.** `/tmp` inside the devcontainer is `tmpfs`; `docker cp` only traverses overlay layers, so it returns "Could not find the file" even when the file demonstrably exists. Confirmed in CI and locally. Fix: stream via `docker exec cat` (encapsulated in `devcontainer-steal.sh`).
- **Process readiness ≠ UI rendered.** The original readiness gate only checked process presence — the first run captured a blank screen because Electron was alive but mid-first-paint. Fix: also gate on `xdotool search --class ToolHive` matching the main window, plus a 2s settling sleep.
- **`cacheFrom: type=gha` was silently failing** with `unknown cache importer: gha`. Fix: add `docker/setup-buildx-action@v3` so BuildKit's gha backend is registered.
- **Layer cache import alone isn't enough** because the `node_modules` named Docker volume doesn't survive across CI jobs, so `postCreateCommand` reruns `pnpm install` from scratch. Fix: cache the volume contents via `actions/cache`, populate before `devcontainers/ci` runs, dump back on cache miss.
- **`devcontainers/ci@v0.3` doesn't actually export the gha cache** (only imports). Discovered while measuring the warm rerun — the apt install layer takes a fresh ~35s every run regardless. Not fixed in this PR; tracked in #2136.

## Cost compared to the production bug-fix agent

> Rough estimates from a sample of two production runs. Real distribution is wider — agent thinking time varies a lot per bug.

The current production agent runs without a devcontainer and takes ~18 min per successful fix, dominated by Phase 1 Opus + Phase 2 Sonnet plus two full test-suite runs ([#24982779175](https://github.com/stacklok/toolhive-studio/actions/runs/24982779175), [#24781689310](https://github.com/stacklok/toolhive-studio/actions/runs/24781689310)).

| | Production agent today | + this PR's substrate | + #2136 (prebuilt image) |
|---|---|---|---|
| Per-fix wall time | ~18 min | ~22 min | ~19 min |
| Overhead vs current | — | +22% | +5% |

Phase 1 (5–9 min) dominates either way — that's the agent's actual thinking, unrelated to substrate. Live-app access mostly augments Phase 1: bugs that currently fall through to Phase 2b ("direct fix without TDD" → lower-quality PRs) might succeed at writing a failing test if the agent could observe the running app. Whether the +22% (or +5% post-#2136) is worth it depends on how much Phase 1's success rate improves on UI/routing/IPC bugs. **This PR doesn't answer that question** — it only proves the substrate. The actual ROI test is to run the integrated agent on a known-tricky UI bug (#663 is the natural candidate) and compare outcomes.

## Follow-ups

- **#2136 — Publish a prebuilt devcontainer image to GHCR.** Replaces the per-run apt + base-image build with a single `docker pull`. Combined with this PR's `node_modules` cache, gets steady-state below 1 minute.
- **#2129 — Enable Chrome DevTools Protocol in the devcontainer.** Lets agents drive the app via accessibility tree / DOM (Playwright CLI) instead of pixel-based `xdotool` + screenshots. Reduces token spend on the agent side.
- **Visual bug-fix agent integration** — separate parallel workflow (`_visual-bug-fix-agent.yml` + `visual-bug-fix-on-label.yml`) that wires this substrate into a label-gated copy of the production agent. Coming as a follow-up PR; doesn't modify the production agent.

## How to fire a run

```
git commit --allow-empty -m "trigger experiment" && git push
```

The workflow is `push:`-triggered and pinned to this branch. Use `[skip ci]` to push without firing.

---

_Experimental — not for routine production CI. Intended to be merged so future agent work has a working substrate to build on._


kantord:
@claude deep review please

PR #2122, opened 2026-04-27
fix(skills): split version from OCI reference when opening install dialog from registry

samuv:
Installing a skill from the **Skills → Registry** tab put the entire OCI ref into the **Name or reference** field of the install dialog and left the dedicated **Version** field empty. Submitting the form then sent something like `{ name: "ghcr.io/stacklok/dockyard/skills/agents-md:0.1.0" }` to the API, even though the request body has a separate `version` parameter.

## Root cause

`getSkillInstallReference` in [`lib/skill-reference.ts`](renderer/src/features/skills/lib/skill-reference.ts) returned a single string that was wired straight into the dialog's `defaultReference` prop. The split between `name` and `version` only existed at submit time — by then the user had been shown (and could have edited) a single conjoined string. The dialog's Version field was prefilled with `''` regardless of what the registry advertised.

## Changes

- **[`lib/skill-reference.ts`](renderer/src/features/skills/lib/skill-reference.ts)** — replace `getSkillInstallReference(skill): string` with `getSkillInstallDefaults(skill): { reference, version? }`. The new helper splits an OCI identifier on `:` (tag) or `@` (digest), or — when the identifier is bare — uses the package's separate `ref` field as the version. Non-OCI fallback (`namespace/name`) is unchanged and returns `version: undefined`.
- **[`dialog-install-skill.tsx`](renderer/src/features/skills/components/dialog-install-skill.tsx)** — add an optional `defaultVersion` prop that prefills the existing **Version** form field.
- **Registry call sites** — [`card-registry-skill.tsx`](renderer/src/features/skills/components/card-registry-skill.tsx), [`table-registry-skills.tsx`](renderer/src/features/skills/components/table-registry-skills.tsx), and [`skill-detail-page.tsx`](renderer/src/features/skills/components/skill-detail-page.tsx) now pass `defaultReference` + `defaultVersion` from the new helper.
- **Local builds ([`card-build.tsx`](renderer/src/features/skills/components/card-build.tsx))** are intentionally untouched — they hand the user-authored `tag` directly to the dialog, since local-build tags are not OCI refs.

## Verification

1. Skills → Registry → click **Install** on a skill whose registry entry has a tagged OCI identifier (e.g. `ghcr.io/stacklok/dockyard/skills/agents-md:0.1.0`). **Name or reference** shows `ghcr.io/stacklok/dockyard/skills/agents-md`, **Version** shows `0.1.0`.
2. Submit as-is — `POST /api/v1beta/skills` body is `{ name: "ghcr.io/.../agents-md", version: "0.1.0", scope: "user" }` instead of the previous conjoined `name`.
3. Skills with no OCI package (git-only) still prefill `namespace/name` and leave Version empty.
4. The skill detail page **Install** button and the registry card view both behave the same as the table.
5. Local-build cards (Skills → Builds tab) are unaffected.
6. `pnpm test:nonInteractive` and `pnpm type-check` pass.


samuv:
@copilot resolve the merge conflicts in this pull request

peppescg (approved):

PR #2137, opened 2026-04-28
feat(playground): persist streaming responses and survive route changes

samuv:
Adds end-to-end persistence for in-flight chat streams in the Playground so the LLM call keeps running in the main process when the user navigates away, and the conversation is saved to SQLite as tokens arrive — not just at `finish`. Also lands a couple of UX touch-ups in the sidebar and the token-usage line.

https://github.com/user-attachments/assets/5e23c767-a68e-4001-9739-626f6a799c5a

## What's new

- **Main-process active streams registry.** New `ActiveStreamsRegistry` keyed by `chatId` keeps every running stream alive in the main process. Late subscribers receive a synthesized replay built from per-block state, then attach to live IPC events for the tail, multiple `WebContents` can attach, and the `UIMessage[]` snapshot is throttled-persisted (~250ms) plus on tool-call boundaries and stream `finish` / `error`. An `AbortController` is wired through to the agent so cancel actually stops the underlying LLM call.
- **`useChat({ resume: true })` reconnect.** The renderer's `ElectronIPCChatTransport` now implements `reconnectToStream`: it asks the main registry for the active stream id, replays buffered chunks, and pipes live IPC events into the AI SDK stream. Switching threads or reloading the route resumes the in-flight assistant message instead of dropping it. The composer's stop button now also calls `chat:stream:cancel`, which aborts the main-process stream, flushes the persisted snapshot, and notifies all subscribers — no more orphaned LLM calls.
- **Streaming indicator in the sidebar.** A new `useStreamingThreads` hook subscribes to `chat:stream:state` broadcasts and exposes the set of thread IDs currently streaming. The Playground sidebar renders a small `Loader2` next to those titles so it's clear which thread is still working in the background, even if the user has navigated to another thread.
- **Sidebar UX polish.** Thread date is now pushed flush-right and the three-dot menu only appears **on hover** (the active thread no longer permanently hides its date). Truncated titles get a `pr-8` gutter on hover so the ellipsis can't overlap the menu icon.
- **Live token-usage UX.** Token usage is accumulated from `finish-step` parts (multi-step tool loops grow the count step-by-step) and merged into the assistant message metadata as it streams. While usage is still unavailable, the chat shows a subtle pulsing `…` placeholder with a tooltip instead of the misleading `0 → 0 = 0` line.

## How to validate

1. Start a long response in any chat → navigate away (e.g. to Skills) and back. The assistant message keeps growing, no chunks are lost, and the response continues from where it was.
2. While streaming, reload the renderer (`Cmd+R`) — the partial assistant message is already in SQLite from the throttled snapshots, so the chat restores without losing the in-progress turn.
3. Open multiple threads streaming in parallel → the sidebar shows a spinner next to each active title; spinners disappear on `finish` / `error`.
4. Press the stop button mid-stream → the running indicator clears, the partial message stays persisted, and no further chunks arrive (verifiable in main-process logs).
5. Multi-step tool-loop chat (Skills Builder etc.): token usage grows after each step's `finish-step`, then settles to the final number on `finish`.
6. Single-step chat: token usage shows a pulsing `…` until the model reports usage, then snaps to the final breakdown.
7. Sidebar: the date sits flush-right on every thread (including the active one); hovering swaps it for the three-dot menu without overlapping the title's truncation ellipsis.


main/src/chat/__tests__/active-streams.test.ts
```diff
@@ -0,0 +1,291 @@
...
+}))
+
+vi.mock('../../logger', () => ({
+  default: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}))
+
+vi.mock('electron', () => ({
+  app: { once: vi.fn() },
+}))
+
```
kantord, line 36: the electron app thing I think is either already mocked globally, or should be 🤔 

or this is to be able to spy on the "once" function? (I guess even in that case the global mock should be usable)

main/src/chat/__tests__/active-streams.test.ts
```diff
@@ -0,0 +1,291 @@
...
+function asWebContents(sender: FakeSender): WebContents {
+  return sender as unknown as WebContents
+}
+
+/** Drain pending microtasks so the for-await loop in runManagedStream can
+ * pick up enqueued chunks. */
+async function flushMicrotasks(times = 20) {
+  for (let i = 0; i < times; i++) {
+    await Promise.resolve()
+  }
```
kantord, line 69: 🤔  it is a little bit unclear what this does or how it works, because on the face of it seems almost a no-op.


this is doing something to node's internal thread scheduling or sth like that?

renderer/src/features/chat/transport/electron-ipc-chat-transport.ts
```diff
@@ -100,6 +110,142 @@ export class ElectronIPCChatTransport implements ChatTransport<ChatUIMessage> {
...
+            isClosed = true
+            window.electronAPI.removeListener?.(
+              'chat:stream:chunk',
+              handleChunk
+            )
+            window.electronAPI.removeListener?.('chat:stream:end', handleEnd)
+            window.electronAPI.removeListener?.(
+              'chat:stream:error',
+              handleError
+            )
```
JAORMX: So, this is a real bug and it took me a minute to spot it.

In `preload/src/api/events.ts:23-30`, `on()` registers a *wrapped* listener: `const wrapped = (_, ...args) => listener(...args)`. The closure-captured `wrapped` is what `ipcRenderer.on` actually holds.

But here we're calling `removeListener('chat:stream:chunk', handleChunk)` with the *unwrapped* `handleChunk`. Electron's `ipcRenderer.removeListener` does reference equality... and `wrapped !== handleChunk`, so this never matches. The listener is never removed.

Net effect: every `attachToActiveStream` call (i.e. every send, every reconnect) accumulates 3 IPC listeners that live forever. After a few dozen turns or page reloads you'll start seeing `MaxListenersExceededWarning`, and stale listeners keep firing on subsequent streams (the `matchesStream` guard saves us from correctness bleed, but the memory keeps growing).

The fix is to use the unsubscribe handle that `events.on` already returns... `useStreamingThreads` does this correctly:

```ts
const unsub = window.electronAPI.on?.('chat:stream:chunk', handleChunk)
// ...
cleanup = () => { unsub?.(); /* ... */ }
```

Note that `EventsAPI.removeListener` is broken-by-design for this same reason... probably worth deleting it from the API entirely so nothing else falls into this trap.

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+    bufferedChunks: [],
+    subscribers: new Set(),
+    runningMessage: undefined,
+    persistTimer: null,
+    pendingWrite: false,
+    abortController,
+    status: 'streaming',
+    toolUiMetadata: initialToolUiMetadata ?? null,
+  }
+  streams.set(chatId, stream)
```
JAORMX, line 277: What happens if the user fires two `chat:stream` calls for the same `chatId` in quick succession? Double-click on send, retry-while-streaming, anything where the user gets impatient.

Today: `streams.set(chatId, stream)` overwrites the previous entry silently. The first stream is still running its `for await` loop, but its registry handle is gone. When the *first* stream's iterator eventually finishes, line 270's `streams.delete(chatId)` deletes the **second** stream's entry. Late subscribers can't reconnect to it anymore even though chunks are still being broadcast. Oops!

Worse: the first stream's `abortController` is never aborted, so we're paying the LLM provider for tokens nobody will ever see. And since both streams broadcast on the same channel, subscribers of the second stream may briefly see chunks intended for the first (the renderer's `matchesStream` filters by `streamId` so this is benign in practice, but it's still ugly).

Options... in order of how aggressive you want to be:

1. **Reject:** check `streams.has(chatId)` at the top of `runManagedStream` (or in the `chat:stream` IPC handler) and return an error to the renderer. Simplest, makes the contract explicit.
2. **Replace with abort:** call `oldStream.abortController.abort()` and `await` its teardown before starting the new one.

I'd lean towards option 1. Surfaces the problem instead of hiding it, and the renderer is already in a position to coordinate (it has `getActiveStreamId`).

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+      `[ACTIVE_STREAMS] Assembler iteration failed for ${chatId}:`,
+      error
+    )
+  })
+
+  try {
+    for await (const chunk of chunkBranch as unknown as AsyncIterable<ChatUIMessageChunk>) {
+      // Buffer is bounded by stream lifetime — released in `finally`
+      // below. Swap to snapshot + tail-buffer if streams ever go huge.
+      stream.bufferedChunks.push(chunk)
```
JAORMX: The TODO comment above is honest, but I think this needs a cap before it ships.

`bufferedChunks` grows for the entire lifetime of the stream so late subscribers can replay everything from the start. For a quick chat turn that's fine. For a long agent loop with chunky tool outputs (think: a Skills Builder run that goes for 10 minutes and dumps multi-megabyte MCP tool responses), this is a memory leak the user has no visibility into.

Good news... we already maintain `runningMessage` via the assembler! That's the assembled state of the in-flight message. Late subscribers don't actually need every delta from t=0, they need:

1. the current assembled `runningMessage` (snapshot)
2. plus any chunks that arrived after that snapshot was last updated (tail)

So `subscribeToStream` could return `{ snapshot: runningMessage, tailChunks: lastN }` and the renderer replays just the tail on top of the snapshot. The renderer already does message-id reconciliation via the AI SDK, so it should drop in cleanly.

Or, simpler floor: cap `bufferedChunks` at e.g. the last 500 entries with a circular buffer + a "missed-chunks" flag. Force the renderer to fall back to the SQLite hydration path if the buffer wrapped.

Either way... unbounded buffer in a long-running registry isn't somewhere we want to be.

main/src/ipc-handlers/chat/streaming.ts
```diff
@@ -1,10 +1,52 @@
...
+}
 
 export function register() {
+  ensurePurgeListener()
+
   ipcMain.handle('chat:stream', async (event, request: ChatRequest) => {
     const streamId = `stream-${Date.now()}-${Math.random().toString(36).substring(2, 9)}`
     handleChatStreamRealtime(request, streamId, event.sender)
     return { streamId }
   })
```
JAORMX, line 53: Note that `handleChatStreamRealtime` is invoked here without `await` and not wrapped in a try/catch. The renderer immediately gets back `{ streamId }` and starts attaching listeners.

If `handleChatStreamRealtime` throws **synchronously before** the registry entry is created (provider not configured, model not found, missing API key validation server-side, settings query failure, etc.), nothing ever broadcasts `chat:stream:error` because there's no `subscribers` set yet, and nothing rejects the IPC call because it already returned. The renderer hangs waiting for chunks that never come, until its local AbortController eventually trips on user navigation or unmount.

Two fixes that work together...

1. In the handler: wrap the call in try/catch and surface the error via the IPC reply (e.g. `return { error: '...' }` or just `throw` so `ipcMain.handle` rejects the renderer's promise).
2. In `handleChatStreamRealtime`: create the registry entry as the *first* thing it does, before any throwable code, so even on creation failure the error path can broadcast.

The current shape essentially says "it'll work as long as nothing fails before we get into `runManagedStream`"... which is exactly the kind of assumption that bites under flaky network or misconfigured providers.

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+try {
+  app.once('before-quit', () => {
+    isShuttingDown = true
+    for (const stream of streams.values()) {
+      flushPersist(stream)
+    }
+  })
+} catch {
+  // `app` may be unavailable in unit-test contexts; that's fine.
+}
```
JAORMX, line 119: Two related things on shutdown / error hygiene that I want to flag together...

**App quit (this hunk):** we set `isShuttingDown = true` and flush each stream's persist, which is great for SQLite durability. But `stream.abortController.abort()` is never called. The upstream LLM HTTP socket is left to the OS to reap when the process dies. On a graceful `app.quit()` flow this is mostly fine because the process actually exits, but provider-side you can keep getting billed until the connection is fully torn down.

**Upstream error (line 258 catch):** same shape. We catch, flush, broadcast, but never abort. If the error comes from `result.toUIMessageStream` rather than the underlying `fetch` (think: schema validation throws partway through assembling), the model fetch may keep running until the provider's read timeout.

Defensive fix in both spots:

```ts
try { stream.abortController.abort() } catch {}
```

before broadcasting / cleaning up. AbortController is idempotent, so even if upstream already aborted us, it's a no-op.

One more thing on this hunk: `app.once('before-quit', ...)` only fires once. If quit gets cancelled (user dismisses a save-prompt dialog, or another `before-quit` listener calls `event.preventDefault()`), a subsequent quit attempt will skip the flush entirely. `app.on(...)` is safer here.

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+      log.warn(
+        `[ACTIVE_STREAMS] Snapshot write failed for ${stream.chatId}: ${result.error}`
+      )
+    }
+  } catch (error) {
+    log.error(
+      `[ACTIVE_STREAMS] Snapshot write threw for ${stream.chatId}:`,
+      error
+    )
+  }
```
JAORMX, line 205: The try/catch swallows persist errors and logs to the main-process console, which is the right safety call (don't kill the stream because SQLite is locked for a moment). But it leaves the user completely in the dark when persistence is *actually* broken.

The scenario I'm worried about: disk full, or SQLite locked because a backup is running. Stream renders fine in the UI, user thinks everything is saved... reload the renderer and the partial is gone. Silent data loss is the worst kind.

Worth surfacing this somehow. One option: track a `persistFailed` flag on the stream and broadcast a `chat:stream:persist-error` event the first time it flips, so the renderer can toast something like "Your chat may not be saved." We don't need to spam (one event per stream is plenty), but a fully silent failure mode here feels wrong.

JAORMX (commented):
Did a connection-resilience pass on this, focused on what happens when streams get reset, restarted, or the renderer reloads mid-flight. The core architecture is solid... multi-subscriber broadcast, throttled persistence, replay buffer for late attachers, abort wiring through to the agent, all there. Most happy paths work as advertised, and the test coverage on the registry is genuinely good.

That said, I found a handful of things I'd want addressed before this lands. The two that worry me most:

1. **IPC listener leak in the renderer transport.** `removeListener` is called with the unwrapped handler, but `events.on` registers a wrapped one. Reference equality fails, the listener never detaches, and we accumulate three listeners per send/reconnect for the lifetime of the renderer. Inline below.
2. **Same-`chatId` concurrent streams silently clobber the registry.** No guard on `streams.set`. The first stream's `finally` ends up deleting the *second* stream's entry, late subscribers can't reconnect, and the first stream's abort controller never fires.

The others are smaller but worth thinking about: unbounded replay buffer, fire-and-forget IPC handler that hides synchronous errors, no abort on upstream error or app quit (so the LLM HTTP socket dangles and we keep paying for tokens), and a fully silent persist-failure path.

I also noticed a smaller race I didn't comment inline on: between `getActiveStreamId` returning a streamId and `subscribeToStream` actually subscribing, the stream can finish and be deleted. The late call returns `null`, which is technically clean (no hang because SQLite hydration covers it), but a tiny "completed-recently" cache would let the renderer get a deterministic `end` event instead of inferring it. Not blocking, just nice-to-have.

Nothing here is fundamentally broken... the bones are good and the persistence-on-finish path is durable. Just a few sharp edges around stream lifecycle that bite specifically when networks misbehave or users get impatient. Inline comments below for each.

Happy to pair on any of these if it speeds things up!

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+    stream.persistTimer = null
+    flushPersist(stream)
+  }, PERSIST_THROTTLE_MS)
+}
+
+function isPersistenceBoundary(chunk: ChatUIMessageChunk): boolean {
+  switch (chunk.type) {
+    case 'tool-input-start':
+    case 'tool-input-available':
+    case 'tool-output-available':
```
kantord: nitpick, but this could be a 


```ts
// global 
const PERSISTANCE_BOUNDARIES = ['foo', 'bar', ...]

// function definition

return PERSISTANCE_BOUNDARIES.includes(chunk.type)
```

main/src/chat/active-streams.ts
```diff
@@ -0,0 +1,370 @@
...
+    streamId,
+    originalMessages,
+    uiMessageStream,
+    abortController,
+    initialSender,
+    initialToolUiMetadata,
+    onComplete,
+  } = options
+
+  const stream: ActiveStream = {
```
kantord, line 259: these are nicpicks but i guess this function is pretty big, it could be in a separate file and you could have subfunctions like 

```ts
const stream = buildActiveStream(options)
```

potentially the length and cyclomatic complexity for this function would be reduced and it would read in a clear and compressed way, almost as a living documentation of the workflow/process it implements

kantord: an already existing example of this pattern that works really well is the broadcast function. it packages/shuffles complexity in a very beautiful way to make this function more readable

main/src/chat/active-streams.ts
kantord: I guess there is already a throttle mechanism in `schedulePersist` for instance. perhaps a generalized throttle/debounce/batching mechanism would solve all problems of this nature?

main/src/chat/__tests__/active-streams.test.ts
samuv: The test file launches under `vitest:electron`, so the *real* `app` singleton is reachable; without a local mock, the module-level `app.on('before-quit', ...)` in `active-streams.ts` attaches to the real app and persists across test files (we hit exactly that kind of cross-file leak earlier in the cycle).

The mock you referring to is `windows.electronAPI` for the rerender I think

main/src/chat/__tests__/active-streams.test.ts
samuv: Fair, the previous comment was too terse. Each `await Promise.resolve()` yields one microtask hop on the JS task queue (no thread/scheduling magic, Node still runs everything on the main thread).

Updated the comment tho

main/src/chat/active-streams.ts
samuv: Done, `runManagedStream` is now a 12-line orchestration function that reads as the workflow itself:

```ts
const stream = buildActiveStream(options)
const [chunkBranch, assemblerBranch] = options.uiMessageStream.tee()
const assemblerTask = startAssembler(stream, assemblerBranch)
try {
  await pumpChunks(stream, chunkBranch)
  await assemblerTask
  finalizeSuccess(stream)
} catch (error) {
  finalizeError(stream, error)
} finally {
  await teardownStream(stream, options.onComplete)
}
```

Phases: `buildActiveStream` (duplicate guard + registry insert + initial-sender wiring + `streaming` broadcast), `startAssembler` (drain the assembler branch into `runningMessage`), `pumpChunks` (broadcast + replay-buffer cap + throttled persist), `finalizeSuccess`, `finalizeError` (also aborts the upstream), `teardownStream`. Kept everything in the same file since the helpers only make sense in the registry's lifecycle, but happy to split into a folder if you'd rather.

main/src/chat/active-streams.ts
samuv: Done. Used a module-level `ReadonlySet<ChatUIMessageChunk['type']>` so the lookup is O(1) and the boundaries are tied back to the chunk-type union (typo would surface at compile time):

```ts
const PERSISTENCE_BOUNDARIES: ReadonlySet<ChatUIMessageChunk['type']> = new Set([
  'tool-input-start', 'tool-input-available',
  'tool-output-available', 'tool-output-error',
  'finish', 'finish-step', 'error',
])

function isPersistenceBoundary(chunk: ChatUIMessageChunk): boolean {
  return PERSISTENCE_BOUNDARIES.has(chunk.type)
}
```

renderer/src/features/chat/transport/electron-ipc-chat-transport.ts
samuv: > `removeListener` is called with the unwrapped handler, but `events.on` registers a wrapped one … listener is never removed.

Confirmed and fixed. `events.on` now hands back its own unsubscribe function and the transport keeps those handles instead of relying on reference equality:

```ts
const offChunk = window.electronAPI.on?.('chat:stream:chunk', handleChunk)
const offEnd   = window.electronAPI.on?.('chat:stream:end',   handleEnd)
const offError = window.electronAPI.on?.('chat:stream:error', handleError)

cleanup = () => {
  offChunk?.(); offEnd?.(); offError?.()
  // ... abort-listener detach
}
```

Updated `electron-ipc-chat-transport.test.ts` to assert the unsubscribe handles fire on both `end` and consumer-side `cancel()`. Global mocks (`vitest.setup.ts`, `electronAPI.ts`) now make `on` return a no-op unsubscribe so any test using the real contract works out of the box.

On killing `EventsAPI.removeListener`: agreed that it's a footgun. I'd rather do that as a follow-up so this PR stays focused — there are a couple of unrelated callers (test fixtures + `useStreamingThreads`) and I'd like to inspect them properly. Filing an issue.

main/src/chat/active-streams.ts
samuv: What happens if the user fires two `chat:stream` calls for the same `chatId` in quick succession? … `streams.set(chatId, stream)` overwrites the previous entry silently … I'd lean towards option 1.

Went with option 1 as you suggested. `runManagedStream` now rejects at the top of `buildActiveStream`:

```ts
if (streams.has(chatId)) {
  throw new Error(`Stream already active for chat ${chatId}`)
}
```

The IPC handler (next comment) catches the rejection and broadcasts `chat:stream:error` to the originating sender, so the renderer's AI SDK stream errors out cleanly. Original stream is unaffected.

Test added: `runManagedStream` throws and the original stream's `streamId` is still the active one after the second attempt.

main/src/ipc-handlers/chat/streaming.ts
samuv: > `handleChatStreamRealtime` is invoked here without `await` and not wrapped in a try/catch.

Wrapped. We keep the fire-and-forget shape (the renderer attaches via IPC events and never awaits the handler), but failures that throw *before* the registry entry exists now reach the originating sender as `chat:stream:error`:

```ts
void handleChatStreamRealtime(request, streamId, event.sender).catch((error) => {
  const message = error instanceof Error ? error.message : 'Unknown error'
  log.error(`[CHAT_STREAM_IPC] Setup failed for ${request.chatId}:`, error)
  if (!event.sender.isDestroyed()) {
    event.sender.send('chat:stream:error', { streamId, chatId: request.chatId, error: message })
  }
})
```

The duplicate-stream rejection from the previous comment flows through the same path, so a double-click sees a real error instead of a hang.

main/src/chat/active-streams.ts
samuv: > The TODO comment above is honest, but I think this needs a cap before it ships … unbounded buffer in a long-running registry isn't somewhere we want to be.

Went your "right answer" route directly — `bufferedChunks` / `bufferOverflowed` / `MAX_BUFFER_CHUNKS` are gone. Tried a 5000-chunk cap first, but it fixed the wrong bug: once tripped, `subscribeToStream` returned `null`, the renderer fell back to SQLite, and live updates stopped reaching the message until the *next* completion — exactly the failure mode we're trying to avoid for long Skills Builder runs. There was also a latent duplicate-text bug: `setMessages(snapshot)` seeds the AI SDK's `state.message` with the in-flight assistant from SQLite, then replaying every original chunk pushes a *second* `TextUIPart` because each `text-start { id }` adds a fresh part on top.

The replacement is the snapshot-from-`runningMessage` design you proposed. `pumpChunks` taps each chunk into per-block structural state on the stream:

```ts
interface ActiveStream {
  messageId: string | undefined
  blockOrder: Array<
    | { kind: 'text'; id: string }
    | { kind: 'reasoning'; id: string }
    | { kind: 'tool'; toolCallId: string }
  >
  textParts: Map<string, { text: string; done: boolean; providerMetadata?: ... }>
  reasoningParts: Map<string, { text: string; done: boolean; providerMetadata?: ... }>
  toolParts: Map<string, { toolName, dynamic, partialInputText, input, output, errorText, state, ... }>
}
```

`subscribeToStream` always succeeds on a live stream and returns synthesized `replayChunks` built from that state via `buildReplayChunks(stream)`: `start { messageId }`, then per `blockOrder` entry one `*-start` + (if non-empty) one **consolidated** `*-delta` carrying the full accumulated text + (if `done`) `*-end`; tools emit `tool-input-start` + (if still streaming) one consolidated `tool-input-delta` + the appropriate `tool-input-available` / `tool-output-available` / `tool-output-error`. This is correct because `processUIMessageStream` only cares that `activeTextParts[id]` exists when a delta arrives — it doesn't validate granularity. So thousands of original deltas collapse to one synthesized delta per block, and replay size is bounded by message structure, not elapsed chunks.

The duplicate-message bug is fixed in the renderer: `useChatStreaming`'s hydrate effect peeks `getActiveStreamId(threadId)` first; when a stream is live, it strips the trailing assistant from the SQLite snapshot before `setMessages`, and the synthesized `start { messageId }` rebuilds it from scratch so the live tail extends a single message. There's a race guard for the `getActiveStreamId` → `subscribeToStream` window: if the stream finishes in between, we re-poke the registry and invalidate the thread query so the now-finalized snapshot reapplies cleanly.

Out of scope for this PR (filed as follow-ups): replay correctness for `start-step` / `finish-step` / token-usage metadata (cosmetic — token count snaps back on the next live `finish-step`) and for `data-*` / `source-*` / `file` parts (not used by current agents).

Tests rewritten:

- Deleted the old buffer-overflow test (no longer applicable).
- New `subscribeToStream` replay assertions for: empty stream → `[]`; single streaming text part → `start` + `text-start` + consolidated `text-delta`; completed text part → same plus `text-end`; mixed text + tool with output → ordered correctly per `blockOrder` with the partial deltas dropped once `state` advances; tool with partial input still streaming → `tool-input-start` + consolidated `tool-input-delta(accumulated)`.
- New `use-chat-streaming` hydration tests: active-stream branch trims the trailing assistant from `setMessages`; no-active-stream branch passes the snapshot through unchanged.

While I was in there I also renamed the IPC payload from `bufferedChunks` to `replayChunks` since "buffered" no longer described what shipped over the wire — touched `preload/src/api/chat.ts`, `electron-ipc-chat-transport.ts`, and the matching transport tests.

main/src/chat/active-streams.ts
samuv: > `stream.abortController.abort()` is never called … `app.once('before-quit', ...)` only fires once.

Both fixed:

```ts
// before-quit: now `app.on` so cancelled-then-retried quits still flush.
app.on('before-quit', () => {
  isShuttingDown = true
  for (const stream of streams.values()) {
    flushPersist(stream)
    try { stream.abortController.abort() } catch {}
  }
})
```

```ts
// Upstream error: abort before flushing/broadcasting.
} catch (error) {
  stream.status = 'error'
  try { stream.abortController.abort() } catch {}
  flushPersist(stream)
  // ... broadcast chat:stream:error, broadcastState('error'), log
}
```

Test added: registers two streams, fires the captured `before-quit` listener, asserts both abort signals trip.

main/src/chat/active-streams.ts
samuv: > Worth surfacing this somehow … track a `persistFailed` flag on the stream and broadcast a `chat:stream:persist-error` event the first time it flips.

Done exactly as proposed. New `persistFailed` flag, new `reportPersistFailure` helper, single broadcast per stream:

```ts
function reportPersistFailure(stream: ActiveStream, error: string) {
  if (stream.persistFailed) return
  stream.persistFailed = true
  broadcast(stream, 'chat:stream:persist-error', {
    streamId: stream.streamId, chatId: stream.chatId, error,
  })
}
```

Renderer side, `useChatStreaming` listens (scoped to the current thread) and shows a single `toast.warning('Your chat may not be saved — saving to disk failed.')`. No spam, no silent loss.

Test added: stub `updateThreadMessages` to return `{ success: false, error: 'disk full' }`, drive the throttle window twice, assert exactly one `chat:stream:persist-error` was broadcast with the correct payload.

samuv:
> Did a connection-resilience pass on this … core architecture is solid … a handful of things I'd want addressed before this lands.

Thanks for the thorough pass, it caught real bugs. All five inline items addressed below, plus the race you mentioned in the summary. PR description updated with a "review feedback" section, validation steps included.

> a smaller race I didn't comment inline on: between `getActiveStreamId` returning a streamId and `subscribeToStream` actually subscribing, the stream can finish and be deleted … a tiny "completed-recently" cache would let the renderer get a deterministic `end` event instead of inferring it. Not blocking, just nice-to-have.

Acknowledged. Leaving it as-is for this PR, the SQLite hydration path covers correctness (the renderer ends up with the final assembled message either way) and the `null` from `subscribeToStream` already cleanly tells the AI SDK stream to fall back. Filing an issue for the completed-recently cache so we can pick it up next time we touch the registry.

kantord (approved):

PR #2184, opened 2026-05-04
feat(playground): installed-skills picker and unified Skill Engineer agent

samuv:
Adds an app-level **Skills** picker to the Playground composer toolbar and a single **Skill Engineer** built-in agent that both authors new skills and audits the ones the user has installed. The picker writes to a small SQLite allow-list which the agent reads at stream time, so only the skills the user explicitly enables are surfaced to the model.

https://github.com/user-attachments/assets/a4af1fd7-a982-43f6-925d-cfa0c7d066f6

## Summary

- **`enabled_skills` table** — new migration `006-enabled-skills.ts`. Pure existence flag (presence of a row = enabled), with reader/writer helpers in `chat-settings-reader.ts` / `chat-settings-writer.ts` and full coverage in `writers-readers.test.ts` and `schema-and-migrator.test.ts`.
- **IPC + preload toggle** — `chat:get-enabled-skills` and `chat:set-enabled-skill` handlers in `main/src/ipc-handlers/chat/skills.ts`, wired into `electronAPI.chat.getEnabledSkills` / `setEnabledSkill`. The getter accepts an optional `installedNames` hint that triggers a server-side `pruneEnabledSkillsTo`, so stale rows are wiped automatically as soon as the renderer hook learns about an uninstall. Matching helpers (`getEnabledSkills`, `setSkillEnabled`, `pruneEnabledSkillsTo`) in `settings-storage.ts`.
- **Skill Engineer built-in agent** — the existing `skills` bundle now ships **6 tools** instead of 2: authoring keeps `write_skill_files` / `build_skill`, inspection adds `list_skills` / `load_skill` / `read_skill_file` / `list_skill_tree`. Resolves the on-disk install across user-home and project-root variants (per-client subdir map for `claude-code`, `cursor`, `codex`, `cline`, `roo-code`, `continue`, `gemini-cli`, `kiro`, `amp-*`) and reads `SKILL.md` plus referenced files (256 KB cap, recursive listing capped at 1000 entries). The `SKILL.md` filename is checked case-strictly via `readdir` so miscased variants (`Skill.md` / `skill.md`) fail the audit on every platform; `load_skill` surfaces a `filesTruncated` flag whenever the bundled file listing is capped, and the audit prompt instructs the agent to fall back to `list_skill_tree` for the relevant subtrees in that case. Agent is renamed to **Skill Engineer**, the system prompt is rewritten with separate authoring + audit flows and a shared PASS / PARTIAL / FAIL rubric.
- **Progressive disclosure of enabled skills** — `createBuiltinAgentTools` is now async and can return an `instructionsSuffix` that `streaming.ts` appends to the agent's instructions for that turn. The `skills` bundle uses it to inject a live list of enabled installed skills (name, version, description, install variants) following the Vercel [\"Add Skills to Your Agent\"](https://ai-sdk.dev/cookbook/guides/agent-skills) progressive-disclosure pattern.
- **Skill picker in the chat composer** — new `SkillSelector` dropdown in `chat-input-prompt.tsx`, backed by `useAvailableSkills` which joins the `getApiV1BetaSkills` payload with the local `enabled_skills` allow-list, dedups by `metadata.name`, and surfaces install variants (user-home vs. one or more project roots) as compact badges with tooltips for the full project path. Includes a live count badge, a \"Clear enabled skills\" action that wipes the raw allow-list (so it can clear stale rows even when the underlying skill is gone), and an `aria-label` on the compact trigger. The trigger is **disabled with an explanatory tooltip** for any agent that doesn't bind the `skills` bundle (e.g. the default ToolHive Assistant), so the affordance stays discoverable but never produces a silent no-op.
- **Toolbar room** — agent / MCP server / model selectors switch from truncated text labels to compact icon + count + chevron triggers wrapped in tooltips, so the new selector fits the same toolbar row at common viewport widths. The MCP picker trigger gains an `aria-label` to match.

## How to validate

1. Open Playground → any chat. The composer toolbar now shows an extra **Skills** dropdown (live count badge) next to the existing Agent / MCP Servers / Model selectors.
2. Install a skill from the Skills page (or use one already installed). Open the new dropdown — it lists the skill with badges for each install site (`User` / `<projectName>`). Toggle it on, the count badge bumps to `1`.
3. Send a message in the same thread on the **Skill Engineer** agent. The system prompt now includes an \"Available installed skills\" section listing the enabled skill (verifiable via main-process logs or by asking the agent to call `list_skills`).
4. Ask *\"audit my `<skill-name>` skill\"*. The agent calls `list_skills`, then `load_skill`, then `read_skill_file` for each referenced bundled file, and produces a PASS / PARTIAL / FAIL verdict with a short bullet list of findings.
5. Switch back to the Toolhive Assistant agent — the picker is now greyed out with a tooltip pointing back at Skill Engineer, confirming the bundle scoping.
6. Uninstall a skill → its name is automatically pruned from `enabled_skills` next time the picker resyncs.

## Out of scope

- No download / extract of OCI artifacts: inspection still reads from the on-disk install location ToolHive materializes for the active client.
- No project-scope **install** flow change — only surfaces project installs in the picker.
- The allow-list is keyed by skill **name only** today, so a skill installed in multiple locations (user-scope and one or more project roots) collapses to a single toggle. Letting the user pick a specific variant — or scoping the audit to one of them — is a follow-up.


renderer/src/features/chat/components/skill-selector.tsx
```diff
@@ -0,0 +1,266 @@
...
+import { cn } from '@/common/lib/utils'
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/common/components/ui/tooltip'
+import { trackEvent } from '@/common/lib/analytics'
+import { useAvailableSkills } from '../hooks/use-available-skills'
+import { useAgents, useThreadAgentId } from '../../agents/hooks/use-agents'
+import { DEFAULT_AGENT_ID } from '../../../../../main/src/chat/agents/types'
```
peppescg: shouldn't be in a shared types 🤔 ?

samuv: damn yes

samuv: moved to common

peppescg (approved):

PR #2187, opened 2026-05-04
feat(deep-links): skill registry detail and one-click install

### LINKED ISSUE #2186: [Story] Deep linking for skills (registry detail + one-click install)
### Story Title

As a user, I want to open a specific skill registry page or trigger a skill install via a clickable link.

## Context

Deep linking already works for MCP servers via two intents defined in `common/deep-links.ts`:

- `toolhive-gui://v1/open-registry-server-detail?serverName=<name>` — opens the registry detail page
- `toolhive-gui://v1/open-registry-server-install?serverName=<name>` — opens the registry detail page **and** auto-opens the install dialog with the server preselected

Skills are now a first-class entity in ToolHive Studio (registry browsing, install/uninstall, builds), but they have no equivalent deep-link entry points. Docs, error messages, blog posts, and external tools should be able to drop users straight into the right skill in Studio — same as we do for MCP servers.

## Goal

Extend the existing deep-link system (`common/deep-links.ts`) with two new intents so skills get the same deep-link surface as MCP servers:

1. **`open-registry-skill-detail`** → navigates to the skill detail page
2. **`open-registry-skill-install`** → navigates to the skill detail page **and** opens the install dialog with the reference prefilled

Skills are addressed by **`namespace` + `skillName`** (not a single name), so both intents take both params.

## Examples

```
toolhive-gui://v1/open-registry-skill-detail?namespace=stacklok&skillName=skill-creator
toolhive-gui://v1/open-registry-skill-install?namespace=stacklok&skillName=skill-creator
```

### Acceptance Criteria

- [ ] New deep-link intent `open-registry-skill-detail` defined in `common/deep-links.ts` with params `{ namespace, skillName }` (both validated via `safeIdentifier`), navigating to `/skills/$namespace/$skillName`.
- [ ] New deep-link intent `open-registry-skill-install` defined in `common/deep-links.ts` with the same params, navigating to `/skills/$namespace/$skillName` with `search: { install: true }`.
- [ ] Both intents added to `allDeepLinks` and to the `deepLinkSchema` discriminated union.
- [ ] Skill detail route (`renderer/src/routes/skills_.$namespace.$skillName.tsx`) extended with `validateSearch` to read `install` (boolean), mirroring how `registry_.$name.tsx` does it today.
- [ ] `SkillDetailPage` reads the `install` search param and, when `true`, opens `DialogInstallSkill` on mount with `defaultReference` (and any default version derived from the skill metadata) prefilled — same UX as the MCP `?install=true` flow.
- [ ] Invalid params (missing namespace or skillName, or characters that fail `safeIdentifier`) fall back to the existing `show-not-found` flow — no app crash.
- [ ] Manual verification works on macOS, Windows, and Linux:
  ``` ./node_modules/.bin/electron . "toolhive-gui://v1/open-registry-skill-detail?namespace=stacklok&skillName=skill-creator"
  ./node_modules/.bin/electron . "toolhive-gui://v1/open-registry-skill-install?namespace=stacklok&skillName=skill-creator" ```
- [ ] Unit tests cover parsing of the two new intents (valid + invalid params) alongside the existing deep-link tests.
- [ ] `.claude/skills/deep-links/SKILL.md` updated to mention the new skill intents in the URL schema examples.

### Target Release

TBD

### Priority

Medium

### Additional Context

- Existing MCP deep-link implementation: `common/deep-links.ts`
- Existing auto-install pattern referenced for parity: `renderer/src/routes/(registry)/registry_.$name.tsx` — see `validateSearch`, the `useSearch({ from: ... })` call, and the `useState(install ?? false)` initial value for the install dialog.
- Install dialog already supports prefill via props: `DialogInstallSkill` accepts `defaultReference` and `defaultVersion`, so the renderer-side wiring is purely about flipping the dialog open and passing the right defaults.
- Original deep-link story for context: #956

#### Open question (please weigh in before implementation starts)

Should `open-registry-skill-install` also accept an optional `?version=<v>` query param to prefill the version field in the install dialog? `DialogInstallSkill` already supports `defaultVersion`, so it would be a small addition. The MCP equivalent does **not** take a version today, so leaving it out keeps strict parity; adding it would make the link more useful for "install exactly this pinned version" links from docs.

samuv:
Extends the existing `toolhive-gui://` deep-link surface so skills get the same one-click entry points as MCP servers — opening a registry skill detail page, or jumping straight into the install dialog with the reference (and an optional version) prefilled. Closes #2186.

https://github.com/user-attachments/assets/c6672b30-942f-4083-a50c-4ebb6f1d95b1

## Summary

- **Two new v1 intents** in [`common/deep-links.ts`](common/deep-links.ts), mirroring the existing MCP pair:
  - `toolhive-gui://v1/open-registry-skill-detail?namespace=<ns>&skillName=<n>`
  - `toolhive-gui://v1/open-registry-skill-install?namespace=<ns>&skillName=<n>[&version=<v>]`
  - Both params validated by the existing `safeIdentifier` regex. The optional `version` reuses the same regex (tag-style only — OCI digests `sha256:…` are intentionally rejected because of the colon; users wanting digest pinning can paste it into the dialog directly).
- **Route + page wiring** ([`skills_.$namespace.$skillName.tsx`](renderer/src/routes/skills_.$namespace.$skillName.tsx), [`skill-detail-page.tsx`](renderer/src/features/skills/components/skill-detail-page.tsx)). `validateSearch` extracts `install` and `version`; `SkillDetailPage` now accepts `initialInstall` / `initialVersion` and opens [`DialogInstallSkill`](renderer/src/features/skills/components/dialog-install-skill.tsx) on mount when triggered. The dialog is keyed by `namespace/name#openCount` so it remounts on every (re)open — `DialogInstallSkill` reads `defaultReference` / `defaultVersion` only inside `useForm`'s `defaultValues`, which would otherwise cache the first deep-link's values. The manual Install button resets any deep-link version override and bumps the key so clicks after a deep link still show the metadata-derived version.
- **Same-page reopen** ([`renderer.tsx`](renderer/src/renderer.tsx)). When a deep-link target carries skill params + `search.install`, the IPC listener now dispatches a new `toolhive:open-install-skill-modal` CustomEvent (with `version` forwarded from `search.version`). The skill detail page listens and reopens the dialog. Kept as a separate event from `toolhive:open-install-modal` to avoid overloading the detail shape — MCP carries `serverName`, skills carry `namespace + skillName + version`.
- **Tests**:
  - [`parse.test.ts`](main/src/deep-links/__tests__/parse.test.ts) — 13 new cases covering both intents (valid params, dotted namespaces like `io.github.stacklok`, missing/invalid params, the documented digest-style rejection, extra-params stripping).
  - [`skill-detail-page.test.tsx`](renderer/src/features/skills/components/__tests__/skill-detail-page.test.tsx) — new file. Covers closed-by-default, mount-time install prefill (metadata version), `initialVersion` override, same-page reopen via the custom event with a different version, and the mismatched-skill ignore case.
- **Docs sync** in `.claude/`, `.codex/`, and `.cursor/skills/deep-links/SKILL.md` — URL schema examples now include both new skill intents and document the tag-only limitation.

## How to validate

The fastest path: open the included tester (or your own browser bookmark) with these three URLs while ToolHive Studio is registered as the protocol handler.

```bash
./node_modules/.bin/electron . "toolhive-gui://v1/open-registry-skill-detail?namespace=io.github.stacklok&skillName=agents-md"
./node_modules/.bin/electron . "toolhive-gui://v1/open-registry-skill-install?namespace=io.github.stacklok&skillName=agents-md"
./node_modules/.bin/electron . "toolhive-gui://v1/open-registry-skill-install?namespace=io.github.stacklok&skillName=agents-md&version=0.0.99"

### DIFF
--- a/renderer/src/features/skills/components/skill-detail-page.tsx
+++ b/renderer/src/features/skills/components/skill-detail-page.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { Button } from '@/common/components/ui/button'
 import {
   TagIcon,
@@ -17,10 +17,28 @@ import { trackEvent } from '@/common/lib/analytics'
 
 interface SkillDetailPageProps {
   skill: RegistrySkill
+  /** When true, opens the install dialog on mount (e.g. from a deep link). */
+  initialInstall?: boolean
+  /**
+   * Overrides the metadata-derived default version when prefilling the
+   * install dialog. Sourced from the deep-link `?version=<v>` param.
+   */
+  initialVersion?: string
 }
 
-export function SkillDetailPage({ skill }: SkillDetailPageProps) {
-  const [installOpen, setInstallOpen] = useState(false)
+export function SkillDetailPage({
+  skill,
+  initialInstall,
+  initialVersion,
+}: SkillDetailPageProps) {
+  const [installOpen, setInstallOpen] = useState(initialInstall ?? false)
+  const [overrideVersion, setOverrideVersion] = useState<string | undefined>(
+    initialVersion
+  )
+  // Bumped every time the dialog is (re)opened from a deep link so the
+  // dialog remounts and `useForm`'s defaultValues pick up the latest
+  // reference/version (defaultValues are only read once on mount).
+  const [openCount, setOpenCount] = useState(initialInstall ? 1 : 0)
 
   const name = skill.name ?? 'Unknown skill'
   const namespace = skill.namespace
@@ -30,6 +48,26 @@ export function SkillDetailPage({ skill }: SkillDetailPageProps) {
   const repoLabel = getDisplayRepoLabel(skill.repository?.url)
   const installDefaults = getSkillInstallDefaults(skill)
   const ociRef = getSkillOciRef(skill)
+
+  useEffect(() => {
+    const handler = (e: Event) => {
+      const detail = (
+        e as CustomEvent<{
+          namespace: string
+          skillName: string
+          version?: string
+        }>
+      ).detail
+      if (detail.namespace === namespace && detail.skillName === skill.name) {
+        setOverrideVersion(detail.version)
+        setInstallOpen(true)
+        setOpenCount((c) => c + 1)
+      }
+    }
+    window.addEventListener('toolhive:open-install-skill-modal', handler)
+    return () =>
+      window.removeEventListener('toolhive:open-install-skill-modal', handler)
+  }, [namespace, skill.name])
 
   const hasBadges = !!(version || repoLabel || license)
 
@@ -83,6 +121,8 @@ export function SkillDetailPage({ skill }: SkillDetailPageProps) {
                   name,
                   namespace: namespace ?? '',
                 })
+                setOverrideVersion(undefined)
+                setOpenCount((c) => c + 1)
                 setInstallOpen(true)
               }}
             >
@@ -135,10 +175,11 @@ export function SkillDetailPage({ skill }: SkillDetailPageProps) {
       />
 
       <DialogInstallSkill
+        key={`${namespace ?? ''}/${skill.name ?? ''}#${openCount}`}
         open={installOpen}
         onOpenChange={setInstallOpen}
         defaultReference={installDefaults.reference}
-        defaultVersion={installDefaults.version}
+        defaultVersion={overrideVersion ?? installDefaults.version}
       />
     </>
   )
--- a/renderer/src/features/skills/components/skill-detail-page.tsx
+++ b/renderer/src/features/skills/components/skill-detail-page.tsx
@@ -17,6 +17,16 @@ import { trackEvent } from '@/common/lib/analytics'
 
 interface SkillDetailPageProps {
   skill: RegistrySkill
+  /**
+   * Route-param namespace. Always present on this page, unlike
+   * `skill.namespace` which is optional on the registry response.
+   */
+  namespace: string
+  /**
+   * Route-param skill name. Always present on this page, unlike
+   * `skill.name` which is optional on the registry response.
+   */
+  skillName: string
   /** When true, opens the install dialog on mount (e.g. from a deep link). */
   initialInstall?: boolean
   /**
@@ -28,6 +38,8 @@ interface SkillDetailPageProps {
 
 export function SkillDetailPage({
   skill,
+  namespace,
+  skillName,
   initialInstall,
   initialVersion,
 }: SkillDetailPageProps) {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: renderer/src/features/skills/components/skill-detail-page.tsx [hunk 1]
```diff
@@ -17,10 +17,28 @@ import { trackEvent } from '@/common/lib/analytics'
...
+  initialVersion,
+}: SkillDetailPageProps) {
+  const [installOpen, setInstallOpen] = useState(initialInstall ?? false)
+  const [overrideVersion, setOverrideVersion] = useState<string | undefined>(
+    initialVersion
+  )
+  // Bumped every time the dialog is (re)opened from a deep link so the
+  // dialog remounts and `useForm`'s defaultValues pick up the latest
+  // reference/version (defaultValues are only read once on mount).
+  const [openCount, setOpenCount] = useState(initialInstall ? 1 : 0)
```
kantord: I wonder if there is a better way to do this. we can look at it tomorrow because I vaguely remember that experienced a similar problem with MCP installation but I don't remember the exact details 😸

peppescg: if the issue is `defaultValues` in `useForm({ defaultValues: { reference, version } })`  I guess if I remember correctly it is possible to use values, that should update automatically and remove the useState 🤔

renderer/src/renderer.tsx
```diff
@@ -59,12 +59,28 @@ if (!window.electronAPI || !window.electronAPI.getToolhivePort) {
...
+        window.dispatchEvent(
+          new CustomEvent('toolhive:open-install-skill-modal', {
+            detail: {
+              namespace: target.params.namespace,
+              skillName: target.params.skillName,
+              version,
+            },
+          })
+        )
+      }
```
peppescg: nit: to avoid nested if/else if

```
 if (!target.search?.install || !target.params) return

 if (target.params.name) {
      window.dispatchEvent(
        new CustomEvent('toolhive:open-install-modal', {
          detail: { serverName: target.params.name },
        })
      )
      return
  }

 if (!target.params.namespace || !target.params.skillName) return

  window.dispatchEvent(
      new CustomEvent('toolhive:open-install-skill-modal', {
        detail: {
          namespace: target.params.namespace,
          skillName: target.params.skillName,
          version: typeof target.search.version === 'string' ? target.search.version : undefined,
        },
      })
    )
```

peppescg (dismissed):
only a nit and a weird useState handle, I am missing the issue but most of the time the defaultValues caused sync issues

samuv:
@kantord @peppescg — addressed in [e3ac1ae](https://github.com/stacklok/toolhive-studio/pull/2187/commits/e3ac1aee53384434e2f62d3658a3ee8d4af1b19d)
Switched DialogInstallSkill to useForm({ values, resetOptions: { keepDirtyValues: true } }) per @peppescg's suggestion. Drops the openCount state and the dialog key entirely; the deep-link version propagates via the values prop and user edits to scope/project_root/clients survive the sync.

renderer/src/renderer.tsx
samuv: yeah, done

kantord (approved):

PR #2207, opened 2026-05-07
fix(playground): defer empty chat thread persistence until first message

samuv:
Fresh `/playground` visits and `+ New chat` clicks used to immediately write an empty thread row to SQLite via `createChatThread()`, even when the user never ended up sending a message. Result: every session leaked one (or more) empty threads into the DB, indistinguishable from real conversations in the sidebar and on reload. This PR moves thread id generation into the renderer and defers the DB write until the user actually submits the first message.

## What changes

- **Renderer-side draft ids.** New [`renderer/src/features/chat/lib/thread-id.ts`](renderer/src/features/chat/lib/thread-id.ts) produces ids with the same `thread_<ts>_<rand>` shape the main process uses, so a draft is visually indistinguishable from a persisted id in the URL and sidebar.
- **No-IPC index resolver.** [`playground.index.tsx`](renderer/src/routes/playground.index.tsx)'s `beforeLoad` keeps the existing "redirect to most-recent / active thread" logic for the case where DB threads exist, but when there are none it now redirects to a locally-generated draft id instead of calling `createChatThread()`.
- **Pending drafts in the sidebar.** [`usePlaygroundThreads`](renderer/src/features/chat/hooks/use-playground-threads.ts) tracks drafts in renderer state with a `pending: true` flag so the sidebar still renders them as normal items. `createThread()` reuses an existing empty draft instead of piling up new ones on repeated `+ New chat` clicks, and `deleteThread()` short-circuits the IPC for drafts (which would otherwise return `Thread not found` and block the navigation away).
- **Promote-on-send.** [`useChatStreaming.validatedSendMessage`](renderer/src/features/chat/hooks/use-chat-streaming.ts) awaits `ensureThreadExists(currentThreadId)` before invoking `sendMessage`, so the row is in place by the time `runManagedStream` starts persisting snapshots via `updateThreadMessages`. Promotion failure short-circuits the send with a thrown error.
- **Main-process plumbing.** [`createThread`](main/src/chat/threads-storage.ts) accepts an optional `explicitId`, and [`ensureThreadExists`](main/src/chat/thread-integration.ts) passes the caller-supplied id through to it on the missing-row branch — previously it would have generated a fresh id and broken the URL/`useChat` instance the renderer is already bound to. New [`main/src/chat/__tests__/thread-integration.test.ts`](main/src/chat/__tests__/thread-integration.test.ts) covers the existing / promoted / generated / failure branches.

## How to validate

1. Fresh app state (or `clearAllThreads`) → open `/playground`. The chat surface renders, the sidebar shows a single pending draft, and **no row appears in `chat-threads` SQLite** until you send a message.
2. Click `+ New chat` repeatedly with the current thread still empty → the same draft id is reused, no extra entries pile up in the sidebar, no IPC `chat:create-chat-thread` is fired.
3. Send the first message in a draft → `ensureThreadExists(<draft-id>)` is called once before the stream starts, the row is written with the **same id** the URL is bound to, snapshot persistence works as before, auto-titling kicks in on completion.
4. Click `+ New chat` again after sending → a new draft is created (the old one is no longer empty); deleting the new draft removes it locally without the "Thread not found" error path.
5. Reload the app while sitting on an unsent draft → the index resolver picks the most recent persisted thread (or generates a fresh draft), the stale draft is gone — no orphaned empty rows.
6. `pnpm run type-check`, lint on changed files, and `pnpm run test:nonInteractive run renderer/src/features/chat main/src/chat renderer/src/routes` (600 chat + 219 route tests) all pass locally.

## Out of scope

- The legacy auto-create branch in [`useThreadManagement`](renderer/src/features/chat/hooks/use-thread-management.ts) (lines 46–53) is unreachable from current routes (they always pass `externalThreadId`) and is left untouched here.
- Drafts intentionally do not survive an app reload. `setActiveThreadId(draftId)` is left as-is — main stores the id but the index resolver validates it against `getAllThreads()` on the next visit and falls back if missing.

### DIFF
--- a/main/src/chat/__tests__/thread-integration.test.ts
+++ b/main/src/chat/__tests__/thread-integration.test.ts
@@ -1,0 +1,84 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+const mockCreateThread = vi.hoisted(() => vi.fn())
+const mockGetThread = vi.hoisted(() => vi.fn())
+const mockSetActiveThreadId = vi.hoisted(() => vi.fn())
+
+vi.mock('../threads-storage', () => ({
+  createThread: mockCreateThread,
+  getThread: mockGetThread,
+  setActiveThreadId: mockSetActiveThreadId,
+}))
+
+import { ensureThreadExists } from '../thread-integration'
+
+beforeEach(() => {
+  vi.clearAllMocks()
+})
+
+describe('ensureThreadExists', () => {
+  it('returns the existing thread id and sets it active when the row already exists', () => {
+    mockGetThread.mockReturnValue({
+      id: 'existing-thread',
+      messages: [],
+      lastEditTimestamp: 0,
+      createdAt: 0,
+    })
+
+    const result = ensureThreadExists('existing-thread', 'Some title')
+
+    expect(result).toEqual({
+      success: true,
+      threadId: 'existing-thread',
+      isNew: false,
+    })
+    expect(mockSetActiveThreadId).toHaveBeenCalledWith('existing-thread')
+    expect(mockCreateThread).not.toHaveBeenCalled()
+  })
+
+  it('promotes a draft id by creating the row with that exact id', () => {
+    mockGetThread.mockReturnValue(null)
+    mockCreateThread.mockReturnValue({
+      success: true,
+      threadId: 'draft-id',
+    })
+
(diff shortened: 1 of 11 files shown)


samuv:
@claude make a review, please not be verbose and stay on point

peppescg (approved):

PR #2217, opened 2026-05-11
fix(main): prevent orphaned thv serve when parent exits before SIGKILL timer

reyortiz3:
## Summary

- `stopToolhive()` nulled the module-level `toolhiveProcess` reference immediately after sending SIGTERM and scheduling SIGKILL via `setTimeout`. When Electron's parent process exited before that 2s timer fired, the synchronous `process.on('exit')` handler at `main/src/app-events/process-exit.ts:7` called `stopToolhive({ force: true })` but it short-circuited on `!toolhiveProcess` — leaving the spawned `thv serve` alive. The next launch then collided on the port, `thv` exited with code 1, and the renderer was pointed at a dead port (all API calls fail with `TypeError: Failed to fetch`).
- Hold `toolhiveProcess` alive until the child's `'exit'` event fires. Use a new `isStopping` flag so `isToolhiveRunning()` still reports `false` once a stop is in flight, preserving the existing contract.
- Guard the child's `'exit'` handler so a late-firing exit on a stale process can't clobber globals already pointing at a fresh restart.

## Test plan

- [x] `pnpm run test:nonInteractive` — full suite passes (2245/2245), including the new regression test under `Bug: orphaned thv after parent exits before SIGKILL timer fires`
- [x] `pnpm run lint`
- [x] `pnpm run type-check`
- [ ] Manual: Cmd+Q the packaged app while `thv serve` is running, then `pgrep -fl 'thv serve'` to confirm no orphan; relaunch and confirm the renderer reaches the backend (no "Failed to fetch")

### DIFF
--- a/main/src/tests/toolhive-manager.test.ts
+++ b/main/src/tests/toolhive-manager.test.ts
@@ -112,6 +112,8 @@ const mockGetQuittingState = vi.mocked(getQuittingState)
 class MockProcess extends EventEmitter {
   pid = 12345
   killed = false
+  exitCode: number | null = null
+  signalCode: NodeJS.Signals | null = null
   stderr = new EventEmitter()
   kill() {
     // Don't automatically set killed - let tests control this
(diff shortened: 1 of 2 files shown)


samuv:
Hey @reyortiz3, thanks for addressing this. FYI, we’ve moved from TCP to sockets, which is a major change on our side. Could you double-check if the fix still applies in the same way and rebase? Or let me know if you need help and I can take a look myself

reyortiz3:
@claude can you verify if all the comments have been addressed?

samuv (approved):
thank you @reyortiz3

PR #2235, opened 2026-05-12
feat(playground): copy button on chat messages

samuv:
Adds a small Copy affordance to every chat message in the playground. Hover-revealed icon button at the bottom of each row, copying the message's text content (excluding tool outputs and reasoning blocks, which already have their own copy paths). Part 1 of three planned UX enhancements — Edit/resend and Queue-while-streaming follow in separate PRs.

Screenshot 2026-05-12 at 16 17 11

## What changes

- **New shared hook** [`renderer/src/common/hooks/use-copy-to-clipboard.ts`](renderer/src/common/hooks/use-copy-to-clipboard.ts) — `useCopyToClipboard()` returns `{ copy(text, { successMessage?, errorMessage? }) }`. Wraps `navigator.clipboard.writeText` in `try/catch` and toasts success/failure via sonner. Replaces the two inline implementations that have been growing in the chat feature.
- **New message-actions slot** [`renderer/src/features/chat/components/chat-message/message-actions.tsx`](renderer/src/features/chat/components/chat-message/message-actions.tsx) — small flex row, hover-revealed (`opacity-0` → `group-hover:opacity-100`, plus `focus-within` so keyboard users can discover it). For this PR only the Copy button is rendered; the `onEdit` prop is reserved for the follow-up PR and intentionally not wired up yet.
- **Copy-text extraction** [`renderer/src/features/chat/lib/message-copy-text.ts`](renderer/src/features/chat/lib/message-copy-text.ts) — `getMessageCopyText(message)` walks `message.parts`, keeps only `text` parts, joins with `\n\n`. Excludes tool outputs and reasoning (internal traces aren't what the user is reaching for when they hit Copy on the visible answer).
- **Row wire-up** in [`user-message.tsx`](renderer/src/features/chat/components/chat-message/user-message.tsx) and [`assistant-message.tsx`](renderer/src/features/chat/components/chat-message/assistant-message.tsx) — added `group` to the row, rendered `<MessageActions copyText={…} />` only when `copyText` is non-empty. Placement: bottom-right on user rows (same side as the right-aligned bubble), bottom-left on assistant rows alongside the timestamp/model footer.
- **Refactor** [`tool-output-content.tsx`](renderer/src/features/chat/components/chat-message/tool-output-content.tsx) — drops its inline `handleCopy` (try/catch + sonner) in favor of the new hook. Behavior preserved.

## How to validate

1. Open a playground thread with at least one user message and one assistant response.
2. Hover over a user message → Copy button fades in at the bottom-right. Click → toast `Copied to clipboard`, paste somewhere external shows the message text.
3. Hover over an assistant message that contains text + a tool call → Copy button fades in at the bottom-left. Copying yields only the assistant's text response, not the tool input/output (those have their own per-block Copy inside `tool-output-content`).
4. Tab through the chat with the keyboard → focusing the Copy button keeps it visible (via `focus-within`) so it's discoverable without a mouse.
5. Triggering a tool that produces a large output → confirm the per-tool-output Copy button still works after the refactor (it now routes through the same shared hook).
6. `pnpm run test:nonInteractive run renderer/src/common/hooks renderer/src/features/chat` — 465/465 locally, including 8 new tests (4 for the hook, 4 for `<MessageActions>`).
7. `pnpm run type-check` and `npx eslint` on the touched files — clean.

## Out of scope

- The Edit button in `<MessageActions>` is intentionally not rendered — that's PR 2 (`feat: edit/resend user message`). The `onEdit` prop already exists in the component's type so PR 2 only has to wire the button body and call site.
- Queue-while-streaming is PR 3.
- [`skill-build-result-card.tsx`](renderer/src/features/chat/components/tool-results/skill-build-result-card.tsx) was left untouched. Its inline copy fires a `trackEvent('Skills: copy reference', …)` on the success branch only; faithfully preserving that through the new hook would require broadening the hook's API (returning a success boolean), which felt like scope creep for this PR. Easy follow-up if we decide we want it.

### DIFF
--- a/renderer/src/common/hooks/__tests__/use-copy-to-clipboard.test.ts
+++ b/renderer/src/common/hooks/__tests__/use-copy-to-clipboard.test.ts
@@ -1,0 +1,77 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { renderHook, act } from '@testing-library/react'
+import { useCopyToClipboard } from '../use-copy-to-clipboard'
+
+const mockToastSuccess = vi.fn()
+const mockToastError = vi.fn()
+vi.mock('sonner', () => ({
+  toast: {
+    success: (...args: unknown[]) => mockToastSuccess(...args),
+    error: (...args: unknown[]) => mockToastError(...args),
+  },
+}))
+
+const writeText = vi.fn()
+Object.defineProperty(navigator, 'clipboard', {
+  value: { writeText },
+  writable: true,
+  configurable: true,
+})
+
+describe('useCopyToClipboard', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    writeText.mockResolvedValue(undefined)
+  })
+
+  it('writes text to the clipboard and shows the default success toast', async () => {
+    const { result } = renderHook(() => useCopyToClipboard())
+
+    await act(async () => {
+      await result.current.copy('hi')
+    })
+
+    expect(writeText).toHaveBeenCalledWith('hi')
+    expect(mockToastSuccess).toHaveBeenCalledWith('Copied to clipboard')
+    expect(mockToastError).not.toHaveBeenCalled()
+  })
+
+  it('shows the default error toast when clipboard write fails', async () => {
+    writeText.mockRejectedValueOnce(new Error('denied'))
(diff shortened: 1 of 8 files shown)


peppescg (approved):
nice 🙌

PR #2241, opened 2026-05-13
feat(playground): edit/resend a previous user message

samuv:
Adds an Edit affordance to a user's own past messages in the playground. Two distinct paths depending on whether the assistant is currently streaming a response to that message:

- **Idle**: Edit prefills the composer; submit sends as a **new message** at the end of the thread. History is unchanged.
- **Streaming the last user message**: Edit prefills the composer with a "rewind" submit button. Submit **cancels the in-flight stream, drops the partial assistant response and the original user message, and sends the edited text as a fresh turn** — effectively a "retry with edits" against the same conversation context.

Edit on older user messages during streaming is unchanged for now (submit is still blocked by the existing stop button) — PR 3 (queue-while-streaming) will round that out separately.

Part 2 of three planned chat UX enhancements. Copy shipped in #2235; Queue-while-streaming follows.

https://github.com/user-attachments/assets/d421590e-cd45-4c56-ad33-3607b305a13b

### Composer plumbing

- **New** [`chat-composer-context.tsx`](renderer/src/features/chat/components/chat-composer-context.tsx) — `useChatComposer()` exposes `{ setDraftText, focusComposer, editingMessageId, beginEdit, clearEdit }`. Returns `null` outside a provider so message rows degrade gracefully.
- **[`chat-interface.tsx`](renderer/src/features/chat/components/chat-interface.tsx)** owns `editingMessageId` state and provides the context, wrapping the message list + composer. The state is paired with a thread-id snapshot so navigating between threads resets the editing state during render (avoiding the `react-hooks/set-state-in-effect` rule).
- **[`chat-input-prompt.tsx`](renderer/src/features/chat/components/chat-input-prompt.tsx)** registers a `{ setText, focusTextarea }` handle into a parent-owned ref on mount. Focus moves selection to the end via `setSelectionRange(text.length, text.length)`. The composer auto-calls `clearEdit()` when the draft text becomes empty or a normal submit succeeds.

### Edit button on user messages

- **[`message-actions.tsx`](renderer/src/features/chat/components/chat-message/message-actions.tsx)** — when `onEdit` is supplied, a `Pencil` icon button renders to the left of the existing Copy button. PR 1 reserved the `onEdit` prop for exactly this.
- **[`user-message.tsx`](renderer/src/features/chat/components/chat-message/user-message.tsx)** wires `onEdit` to `composer.beginEdit(message.id, text)`. Assistant messages don't get Edit.

### Rewind-and-retry path

- **[`use-chat-streaming.ts`](renderer/src/features/chat/hooks/use-chat-streaming.ts)** exposes `lastUserMessageId` (derived) and a new `rewindAndResend({ text, files, editingMessageId })` method. The method awaits `cancelStream` (main-side) → `stop()` (renderer-side) → `clearError` → `setMessages(slice)` → `updateThreadMessages(trimmed)` (DB sync) → `validatedSendMessage({ text, files })`. Defensive fallback to a normal send if the editing id is somehow missing from `messages`.
- **DB sync via existing IPC**: re-used `window.electronAPI.chat.updateThreadMessages(threadId, messages)` (already exposed for thread-level edits). No new IPC handler needed. The trim is persisted immediately after cancel+stop complete, so a mid-flight navigation can't resurrect the partial assistant.
- **Submit button decision** in `ChatInputPrompt`: when `editingMessageId === lastUserMessageId` AND status is streaming/submitted, the button switches to a `RefreshCw` icon (`aria-label="Resend edited message"`) and routes to `rewindAndResend` instead of `stop` or `send`. Otherwise the existing send/stop behavior is unchanged.
- **Edit-active chip** rendered inside the composer body above the textarea when rewind mode is active: *"Editing last message — submit to rewind and retry"* with a cancel button. Subtle styling (`text-muted-foreground`, small badge). Cancel clears both `editingMessageId` and the composer draft.

### Why a context-with-ref instead of lifting state up?

`ChatInputPrompt` is rendered in two parent containers (empty-state overlay vs bottom toolbar), each keyed `key={threadId ?? 'no-thread'}` so they remount per thread. Lifting `useThreadDraft` up to `ChatInterface` would have required rekeying that wrapper too. The ref-registration pattern keeps draft state where it already lives and presents the same `{ setDraftText, focusComposer, beginEdit, clearEdit }` API the consumer expects.

### Stop semantics

`useChat.stop()` returns a Promise that resolves when the AI SDK abort lands. On cancellation the SDK does NOT fire `onFinish` (that's success-only), but `runManagedStream`'s `finalizeError` path in `main/src/chat/streaming.ts` can still trigger one final `flushPersist` — which is why `rewindAndResend` issues `updateThreadMessages(trimmed)` AFTER both cancel+stop have completed. Even if a stale partial snapshot races us, the next stream's `onFinish` overwrites with `[...trimmed, newUser, newAssistant]`.

### Idle edit path
1. Open a thread with at least one prior user message. Hover over a user message → both Copy and Edit buttons appear.
2. Click Edit → composer text is replaced, textarea is focused, cursor at end. Original message remains in the history.
3. Tweak text and submit → edited version appears as a **new** message at the bottom; history is untouched.
4. Edit a message that contains text + an attachment → only the text prefills; the attachment is not re-uploaded.

### Streaming rewind path
5. Send a slow-streaming message in a thread. While the assistant is still streaming, hover over the user message and click Edit → composer prefills, a chip appears above the textarea reading *"Editing last message — submit to rewind and retry"*, and the submit button shows a refresh icon.
6. Modify the text and click submit → the in-flight assistant response disappears, the original user message disappears, and the edited text appears as a fresh user message with a new stream starting underneath it. Token count for the original turn does not survive.
7. Cancel the edit while streaming: click the chip's cancel button → composer clears, the chip disappears, the assistant continues streaming the original turn normally.
8. Edit an **older** user message during streaming → composer prefills as usual but the submit button stays as Stop. Submit just stops the stream (no rewind). (PR 3 will queue the message in this case.)

### Edge cases
9. Edit while streaming, then navigate to another thread before submitting → editing state clears (the chip is gone in thread B, gone again when you return to A — `editingMessageId` resets on thread switch). Stream continues in the background.
10. Edit, then manually clear the composer to empty → the chip disappears and the rewind mode resets to normal (next submit during streaming will just stop, not rewind).

### Quality
11. `pnpm run test:nonInteractive` — 2358/2358 pass (376 chat-feature tests, +16 net new).
12. `pnpm run type-check` — clean.
13. ESLint on touched files — clean.

## Out of scope

- **Editing assistant messages** — explicitly user-only per the original ask.
- **Editing older user messages during streaming** — submit stays blocked; PR 3 will handle the queue case.
- **Carrying forward original-message attachments through a rewind** — only the composer's current attachments are sent on rewind.
- **Inline "this is a draft of message N"** — cleaner UX but requires composer-state changes far beyond a prefill; deferred.
- **Queue-while-streaming** — PR 3 in the series.

### DIFF
--- a/renderer/src/features/chat/components/chat-input-prompt.tsx
+++ b/renderer/src/features/chat/components/chat-input-prompt.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useRef, type RefObject } from 'react'
 import type { ChatStatus, FileUIPart } from 'ai'
 import log from 'electron-log/renderer'
+import { RefreshCw, X } from 'lucide-react'
 import {
   PromptInput,
   PromptInputActionAddAttachments,
@@ -17,6 +18,7 @@ import {
   usePromptInputAttachments,
   type PromptInputMessage,
 } from '@/common/components/ai-elements/prompt-input'
+import { Button } from '@/common/components/ui/button'
 import { trackEvent } from '@/common/lib/analytics'
 import {
   Tooltip,
@@ -74,6 +76,17 @@ interface ChatInputProps {
     text: string
     files?: FileUIPart[]
   }) => Promise<void>
+  /**
+   * Cancel the active stream, drop the partial assistant response, drop the
+   * original user message being edited, and send the new text as a fresh
+   * message. Only invoked when `editingMessageId === lastUserMessageId` and
+   * the assistant is currently streaming.
+   */
+  onRewindAndResend?: (args: {
+    text: string
+    files?: FileUIPart[]
+    editingMessageId: string
+  }) => Promise<unknown>
   onStopGeneration: () => void
   onSettingsOpen: (isOpen: boolean) => void
   handleProviderChange: (providerId: string) => void
@@ -86,6 +99,12 @@ interface ChatInputProps {
    * outside the React subtree (e.g. clicking Edit on a past user message).
    */
   composerHandleRef?: RefObject<ChatComposerHandle | null>
+  /** Id of the user message currently being edited, or `null` when idle. */
+  editingMessageId?: string | null
+  /** Id of the most recent user message in the thread, or `null` if none. */
+  lastUserMessageId?: string | null
+  /** Exit edit mode (called from the chip's cancel button and on auto-clear). */
+  onClearEdit?: () => void
 }
 
 function InputWithAttachments({
@@ -101,10 +120,22 @@ function InputWithAttachments({
   hasProviderAndModel,
   hasMessages,
   threadId,
-}: Omit<ChatInputProps, 'onSendMessage' | 'composerHandleRef'> & {
+  isEditingStreaming,
+  onCancelEdit,
+}: Omit<
+  ChatInputProps,
+  | 'onSendMessage'
+  | 'onRewindAndResend'
+  | 'composerHandleRef'
+  | 'editingMessageId'
+  | 'lastUserMessageId'
+  | 'onClearEdit'
+> & {
   text: string
   setText: (text: string) => void
   textareaRef: RefObject<HTMLTextAreaElement | null>
+  isEditingStreaming: boolean
+  onCancelEdit: () => void
 }) {
   const attachments = usePromptInputAttachments()
   const prevTextRef = useRef(text)
@@ -135,6 +166,10 @@ function InputWithAttachments({
 
   const handleSubmit = () => {
     trackEvent(`Playground: submit`, { 'playground.status': status })
+    // In "editing the streaming message" mode the form's onSubmit owns the
+    // rewind-and-resend flow — don't fire the stop handler here, the parent
+    // will cancel as part of that path.
+    if (isEditingStreaming) return
     const isStoppable = ['streaming', 'error', 'submitted'].includes(status)
     if (isStoppable) {
       onStopGeneration()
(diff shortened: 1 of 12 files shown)


renderer/src/features/chat/components/chat-input-prompt.tsx
```diff
@@ -60,17 +76,41 @@ interface ChatInputProps {
...
+   * Cancel the active stream, drop the partial assistant response, drop the
+   * original user message being edited, and send the new text as a fresh
+   * message. Only invoked when `editingMessageId === lastUserMessageId` and
+   * the assistant is currently streaming.
+   */
+  onRewindAndResend?: (args: {
+    text: string
+    files?: FileUIPart[]
+    editingMessageId: string
+  }) => Promise<unknown>
```
peppescg, line 89: the unknown here is expected?

renderer/src/features/chat/components/chat-input-prompt.tsx
```diff
@@ -182,14 +276,69 @@ export function ChatInputPrompt({
...
+
+  // Expose imperative controls so siblings outside this subtree (e.g. the
+  // message list's "Edit message" button) can pre-fill and focus the
+  // composer. The ref is populated on mount and cleared on unmount so the
+  // context can tell whether a composer is currently mounted.
+  useEffect(() => {
+    if (!composerHandleRef) return
+    composerHandleRef.current = {
+      setText,
+      focusTextarea: () => {
```
peppescg, line 317: I would move this into a useEditMsg hook for readebility

peppescg (approved):
nice job. No blockers only nit

samuv:
> nice job. No blockers only nit

nice, I will tackle them in the next PR

renderer/src/features/chat/components/chat-input-prompt.tsx
samuv: no ti should be void

PR #2253, opened 2026-05-14
feat(playground): per-thread model, MCP, and skills selection

samuv:
## Summary

- Each playground thread now persists its own model/provider, enabled MCP tools, and enabled skills (agent was already per-thread). Switching threads no longer reshuffles the toolbar.
- Globals stay as the "default for new threads" source. Every per-thread mutation dual-writes the global so the next new thread inherits the last-used selection — UX is unchanged on the first pick.
- A one-time migration (`007-per-thread-settings.ts`) snapshots current globals into every existing thread so no thread regresses on upgrade.

## What changed

- **Schema** — adds nullable `selected_provider`, `selected_model`, `enabled_mcp_tools` (JSON), `enabled_skills` (JSON) columns to `threads`; mirrors the `agent_id` precedent from migration 005. Snapshot happens in the same transaction.
- **Main process** — new `thread-settings-storage.ts` + IPC module; readers/writers extended with focused per-thread helpers; `writeThread` preserves the four new columns when callers don't supply them; streaming path now resolves MCP tools and skills per-thread via `request.chatId`.
- **Preload** — adds a `threadSettings` sub-object mirroring the `agents` shape; `getMcpServerTools` accepts an optional `threadId`.
- **Renderer hooks** — `useChatSettings`, `useAvailableServers`, `useAvailableSkills` accept an optional `threadId`, nest query keys under `['chat','thread',threadId,…]`, and dual-write to the global "last used" default. Mutation does synchronous `setQueryData` before invalidating so a Send immediately after a model click sees the new selection without racing the refetch.
- **Renderer transport** — `ElectronIPCChatTransport` reads the per-thread `selectedModel` cache first, falls back to global. Without this the global "last used" would override the active thread's selection.
- **UI** — `mcp-server-selector`, `mcp-tools-modal`, `skill-selector` thread `threadId` through and dual-write per-thread + global. `chat-input-prompt` passes `threadId` to `McpServerSelector`.
- **Tests** — `skill-selector.test.tsx` mocks updated for the new `threadSettings` IPC surface.

## Behaviour

- **New thread defaults**: inherit last-used selection (via the dual-write into globals).
- **Existing threads on upgrade**: keep working — migration snapshots current globals into every row.
- **Global pickers** (provider settings dialog) stay single-write to globals so they don't retroactively mutate the active thread.

## Test plan

- [x] `pnpm run type-check` clean
- [x] `pnpm run test:nonInteractive` clean for changed files (`use-chat-settings.test.ts`, `skill-selector.test.tsx`, `use-available-skills.test.ts`, `threads-storage.test.ts`)
- [x] **Fresh DB**: pick model X / MCP server A / skill Y → confirm both the global tables and the thread row have matching values
- [x] **Migration**: open a pre-feature DB with existing threads → confirm `SELECT selected_provider, selected_model, enabled_mcp_tools, enabled_skills FROM threads` is non-NULL on every row
- [x] **Per-thread isolation**: change model/MCP/skill in thread A → start new thread (inherits A) → change in thread B → switch back to A and confirm A still shows A's selection
- [x] **Restart persistence**: quit, relaunch, each thread keeps its selection
- [ ] **Draft thread**: open Playground with no thread active, toggle a skill — the row materialises and persists across navigation
- [ ] **Stream correctness**: send a message in thread A and thread B with different selections, confirm responses come from the correct model

### DIFF
--- a/main/src/chat/mcp-tools.ts
+++ b/main/src/chat/mcp-tools.ts
@@ -17,6 +17,7 @@ import { createMainProcessApiClient } from '../unix-socket-fetch'
 import log from '../logger'
 import type { AvailableServer } from './types'
 import { getEnabledMcpTools } from './settings-storage'
+import { getThreadEnabledMcpTools } from './thread-settings-storage'
 import {
   type McpToolDefinition,
   buildRawTransport,
@@ -278,7 +279,8 @@ export async function getToolhiveMcpInfo(
 
 // Get MCP server tools information
 export async function getMcpServerTools(
-  serverName: string
+  serverName: string,
+  threadId?: string
 ): Promise<AvailableServer | null> {
   if (!serverName) {
     log.error('getMcpServerTools: serverName is not passed')
@@ -287,8 +289,11 @@ export async function getMcpServerTools(
   const workloads = await fetchWorkloads()
   const workload = workloads.find((w) => w.name === serverName)
 
-  // Get enabled tools for this server
-  const enabledTools = await getEnabledMcpTools()
+  // Get enabled tools for this server (per-thread when threadId is given,
+  // otherwise fall back to the global defaults used by settings UIs).
+  const enabledTools = threadId
+    ? getThreadEnabledMcpTools(threadId)
+    : await getEnabledMcpTools()
   const enabledToolNames = enabledTools[serverName] || []
 
   if (!workload && serverName === TOOLHIVE_MCP_SERVER_NAME) {
@@ -337,7 +342,7 @@ export async function getMcpServerTools(
 }
 
 // Create MCP tools for AI SDK
-export async function createMcpTools(): Promise<{
+export async function createMcpTools(threadId?: string): Promise<{
   tools: ToolSet
   clients: Awaited<ReturnType<typeof createMCPClient>>[]
   enabledTools: Record<string, string[]>
@@ -393,7 +398,9 @@ export async function createMcpTools(): Promise<{
   try {
     const [workloads, resolvedEnabledTools] = await Promise.all([
       fetchWorkloads(),
-      getEnabledMcpTools(),
+      threadId
+        ? Promise.resolve(getThreadEnabledMcpTools(threadId))
+        : getEnabledMcpTools(),
     ])
     enabledTools = resolvedEnabledTools
     // Flag is set here, AFTER the two required calls resolved, so any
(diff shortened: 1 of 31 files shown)


main/src/chat/mcp-tools.ts
```diff
@@ -287,8 +289,11 @@ export async function getMcpServerTools(
   const workloads = await fetchWorkloads()
   const workload = workloads.find((w) => w.name === serverName)
 
-  // Get enabled tools for this server
-  const enabledTools = await getEnabledMcpTools()
+  // Get enabled tools for this server (per-thread when threadId is given,
+  // otherwise fall back to the global defaults used by settings UIs).
+  const enabledTools = threadId
+    ? getThreadEnabledMcpTools(threadId)
```
peppescg, line 295: this is not needed to be awaited like the below one?

samuv: The asymmetry is intentional but I get how it looks suspicious.

- `getThreadEnabledMcpTools(threadId)` is **synchronous**: it just reads a single per-thread row from SQLite and returns `Record<string, string[]>` directly.
- `getEnabledMcpTools()` is **async**: it hits the ToolHive workloads API to filter out tools belonging to stopped servers before returning, so it has to be awaited.

The ternary works because both branches resolve to the same shape (`Record<string, string[]>`) — `await` on a non-promise is a no-op, so awaiting the sync branch would also be valid but redundant. I went with the no-`await` form to make the sync/async difference explicit at the call site.

renderer/src/features/chat/components/mcp-server-selector.tsx
```diff
@@ -50,11 +88,8 @@ export function McpServerSelector() {
       })
       // Disable all tools for this server
       try {
-        await window.electronAPI.chat.saveEnabledMcpTools(serverName, [])
-        // Invalidate query to refetch latest state
-        queryClient.invalidateQueries({
-          queryKey: ['chat', 'enabledMcpServers'],
-        })
+        await saveServerTools(serverName, [])
+        invalidateEnabledTools()
```
peppescg: shouldn't be ?
```
const response = await saveServerTools(serverName, [])
if (response.success) invalidateEnabledTools()
```

samuv: You're right, this is a real bug — same class as the dual-write issues from the earlier round, I missed it in the disable branch. The disable path was invalidating the cache unconditionally and swallowing failures, so the UI would render the server as disabled while the DB still held the old enabled tools.

Fixed in dd235783: the disable branch now mirrors the enable branch — check `response.success`, only invalidate on success, toast on failure.

peppescg (approved):

PR #2263, opened 2026-05-15
feat: add color scheme customization

kantord:
This PR allows for design tokens to be overridden on the user level. 

What is included:
- user-level override using JSON on all operating systems
- an example theme for debugging that the theme system works
- technical documentation on the feature
- a new theme that matches adjusted brand identities

What is only partially included:
- Adjustments to the tokens and work to simplify tokens and unify them with other codebases
-things related to the marketing call to action feature: still considering the best approach for those, but **in practice theme overriding won't be needed for those**, they just need to look good with the default theme. Best course of action might be to ignore the issue.

What is not included:
- system-level theme override on Mac: https://github.com/stacklok/toolhive-studio/issues/2267
- documentation site updates on this feature (would make more sense after better unification across code bases)
- making the new theme with new brand identity default
- autoreload/live reload for theme changes. A single theme handles both dark and light mode, so live reloading/autoreloading is not needed for any known use case today. A theme override change requires closing and opening the app
- any GUI for theme picking etc, not clear product requirement for those

### "stacklok theme" (NOT made default yet)

after this pr is merged, we need to decide if we want to make this the default theme

https://github.com/user-attachments/assets/61751b29-6c95-4ba8-80cf-7ee7c1d2942f

### "demo theme" for testing

Not meant to be particularly pretty, intentionally very different from the default theme to make it easier to spot problems/"stress" test override capability.

https://github.com/user-attachments/assets/64a870c5-b417-4755-b24b-ef6c9cbbde0a


kantord:
@claude deep review please

peppescg (approved):

Issue #2296, opened 2026-05-29
chore(settings/branding): parameterize ToolHive literals via app.getName() exposed in preload


peppescg:
## Summary

Many user-visible strings in Studio hard-code the literal `"ToolHive"` instead of reading the runtime product name. This issue tracks refactoring those literals to read from a single source of truth: Electron's built-in `app.getName()` (i.e. `productName` from `package.json`), exposed to the renderer via the preload bridge.

The refactor:

- Expose `app.getName()` **synchronously** via a one-shot preload IPC at load time (`get-app-display-name`)
- Surface the value as `window.electronAPI.appDisplayName` (a plain `string`, available immediately — no async/Suspense)
- Replace every user-visible `"ToolHive"` literal with the template form `${appDisplayName}`

This is a **parameterization-only** refactor. **OSS UI is unchanged**: `productName` in OSS stays `"ToolHive"`, so `app.getName()` returns `"ToolHive"` and the rendered text is identical to today. Downstream rebrands (including our enterprise distro shipping as `"Stacklok Desktop"`) get the rebrand for free without touching a single line of overlay code.

Context: internal Stacklok Slack thread on 2026-05-29 — the previously proposed direct string-replace (`"ToolHive"` → `"Stacklok Desktop"` in OSS) was reverted in favor of this parameterized approach so OSS branding doesn't change.

## Out of scope

- **PR #2295** — log-path IPC change. It is the **canonical precedent** for the pattern proposed here (one-shot preload IPC surfacing a main-process value as a synchronous renderer property). It can land in parallel.
- **Window title / `productName` / installer artifact names** — already driven by `forge.config.ts` and `package.json`.
- **Non-string identifiers** (`APP_IDENTIFIER`, `DEEP_LINK_PROTOCOL`, etc.).
- **`thv` CLI / OSS `stacklok/toolhive` references** — see "Keep as-is" section below.

## Infrastructure (new bits required)

A small one-time bootstrap is needed before any literals can be replaced. The new IPC `get-app-display-name` returns `app.getName()`, which Electron resolves from the `productName` field of `package.json`. The IPC handler is the **single source of truth** the renderer consumes — and it's wired directly to `package.json` `productName`. Forks / branded builds that override `productName` (e.g. our enterprise distro setting it to `"Stacklok Desktop"`) get the rebrand for free — no overlay code changes needed.

1. **Main-process IPC handler** — `main/src/ipc-handlers/app.ts`:

   ```ts
[...]

PR #2300, opened 2026-05-29
feat: remove hardcoded app names

kantord:
see #2296

Notes:

- `APP_NAME` needed to be changed, because it was used for various different things, such as referring to `thv` but also for package configuration. This means that I could not leave this const as it is. I change it's usages that were referring to the desktop app to use `APP_DISPLAY_NAME` (derived from package.json), then I basically renamed it to `THV_DISPLAY_NAME` to avoid confusion. This way the existing correct references to Toolhive (not Toolhive Studio) were kept unchanged, but with a different variable name
- some usages of `APP_NAME` were referring to the desktop app, and explicitly diverging form the `APP_DISPLAY_NAME` format; in such ways that these usages cannot be derived from package.json without a deeper reworking and audit of the entire build process. It's not the right choice to do that here, so for those usages I was forced to create a "legacy const" named `DEPRECATED_APP_NAME_FOR_BUILD` so this issue will not spread.
- In some cases a usage was slightly ambiguous or potentially confusing to the user, but also did not add any informational value. In some such cases I opted for simply removing the brand name, for instance: https://github.com/stacklok/toolhive-studio/pull/2300/changes#diff-78aca5d68c50f753ae7f2b0d719ff08d8a2bde69433349fa0080426147720644

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -64,6 +64,7 @@ export default defineConfig([
     ignores: [
       '**/__tests__/**',
       '**/*.test.{ts,tsx}',
+      '**/mocks/**',
       'common/app-info.ts',
     ],
     rules: {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -53,4 +53,38 @@ export default defineConfig([
       globals: globals.node,
     },
   },
+  // EXPERIMENT (issue #2296): ban hardcoded brand literals; enumerate sites.
+  {
+    files: [
+      'renderer/**/*.{ts,tsx}',
+      'main/**/*.{ts,tsx}',
+      'common/**/*.{ts,tsx}',
+      'preload/**/*.{ts,tsx}',
+    ],
+    ignores: [
+      '**/__tests__/**',
+      '**/*.test.{ts,tsx}',
+      'common/app-info.ts',
+    ],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        {
+          selector: 'Literal[value=/ToolHive/]',
+          message:
+            "Hardcoded 'ToolHive' brand string — import from @common/app-info instead.",
+        },
+        {
+          selector: 'JSXText[value=/ToolHive/]',
+          message:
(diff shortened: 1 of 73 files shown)


eslint.config.mjs
```diff
@@ -53,4 +53,38 @@ export default defineConfig([
...
+    ],
+    ignores: [
+      '**/__tests__/**',
+      '**/*.test.{ts,tsx}',
+      'common/app-info.ts',
+    ],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        {
```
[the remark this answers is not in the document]
kantord: we will fix them all on this PR, or otherwise address them

kantord:
@claude i know this pr is not yet fully done, but can you please do a deep review of it for me?

make sure your summary is succint and you double validate every fiding

main/src/system-tray.ts
```diff
@@ -135,8 +134,8 @@ const handleStartOnLogin = async (toolHiveIsRunning: boolean) => {
 
 const createStatusMenuItem = (toolHiveIsRunning: boolean) => ({
   label: toolHiveIsRunning
-    ? `🟢 ${APP_NAME} is running`
-    : `🔴 ${APP_NAME} is stopped`,
+    ? `🟢 ${app.getName()} is running`
+    : `🔴 ${app.getName()} is stopped`,
```
kantord, line 138: note: this is controversial: the logic is strictly about toolhive. So we might want to change this back to ToolHive. But it looks a bit off because nothign is explaining what ToolHive even is.

kantord:
@claude another deep review please

JAORMX (approved):

renderer/src/common/components/error/connection-refused-error.tsx
```diff
@@ -53,13 +54,16 @@ export function ConnectionRefusedError() {
     },
     onSuccess: (result) => {
       if (result.success) {
+        // eslint-disable-next-line no-restricted-syntax -- TODO: decide on branding in logs
```
peppescg, line 57: why we cannot change this from app-info const?

PR #2304, opened 2026-06-04
fix(startup): eliminate cold-start window flash by awaiting socket readiness

peppescg:
## Summary

Three coupled changes that together remove the dark-window flash users see on cold launch, between the app icon being clicked and the first usable UI frame.

### 1. `startToolhive()` actually awaits a real socket connection
`main/src/toolhive-manager.ts`

The current implementation has a missing \`return\` on its \`Sentry.withScope\` wrapper, and even after fixing that it only awaits \`spawn()\` (synchronous). So the renderer could fire its first \`getHealth\` while thv was still binding the socket — producing an ENOENT in main logs and a retry-driven flash on cold start.

The new \`waitForSocketReady()\` polls a real \`net.connect()\` probe (cross-platform: UNIX sockets + Windows named pipes — \`existsSync\` is unreliable for the latter and even on Unix the file can appear before the server has called \`Listen()\`). 15s timeout with 100ms probe interval, graceful resolve-anyway fallback so the renderer's existing retry logic still applies if thv is genuinely slow to start.

### 2. \`BrowserWindow\` opens with a theme-matched background
\`main/src/main-window.ts\`

Adds \`backgroundColor\` to the BrowserWindow constructor, sampling \`nativeTheme.shouldUseDarkColors\` and using values that mirror the Tailwind theme \`--background\` tokens (\`#f9faf9\` light / \`#2e2e2e\` dark, which approximates \`oklch(0.1867 0 0)\` in sRGB). Without this the OS shows a default white/black window for the brief gap between window creation and the first paint of \`index.html\`.

### 3. Static splash painted from the first parse of \`index.html\`
\`renderer/index.html\`

The Vite bundle in dev (and the chunked production bundle) takes ~1s to load and evaluate. During that window, only the bare BrowserWindow background was visible. Now the HTML body contains an inline splash card (\"Starting ToolHive\" + spinner) styled with CSS variables that mirror the Tailwind \`--background\` / \`--foreground\` / \`--border\` tokens for both light and dark via \`prefers-color-scheme\`. \`createRoot().render()\` replaces the splash contents on first commit, so React takes over transparently.

## Why these three together

Each change closes one of three sequential gaps on cold start:

| Gap | Without fix | With fix |
|---|---|---|
| Window opened → \`index.html\` parsed (~50ms) | Default OS bg (white/black) | Theme-matched bg color |
| \`index.html\` parsed → React mounts (~1s) | Bare body color | Static splash card |
| React mounts → first route renders | Existing app skeleton | Existing app skeleton |
| First route → thv socket ready | ENOENT + retry flash | thv already ready (gap moved into \`startToolhive\` await) |

## Test plan

- [ ] \`pnpm test:nonInteractive\` — passes (217 files, 2450 tests)
- [ ] \`pnpm type-check\` — passes
- [ ] \`pnpm lint\` — passes
- [ ] \`pnpm format\` — clean
- [ ] Manual cold-start on macOS: dock bounces, window appears with splash visible, transitions seamlessly to app
- [ ] Manual cold-start on Linux: same behavior
- [ ] Manual cold-start on Windows: same behavior (named-pipe probe path)
- [ ] Verify \`pnpm e2e\` still passes (build awaits socket; downstream UI gates should now find thv ready immediately)

### DIFF
--- a/main/src/main-window.ts
+++ b/main/src/main-window.ts
@@ -1,9 +1,15 @@
-import { app, BrowserWindow, shell } from 'electron'
+import { app, BrowserWindow, nativeTheme, shell } from 'electron'
 import path from 'node:path'
 import log from './logger'
 import { hideWindow } from './dock-utils'
 import { getQuittingState, getTray } from './app-state'
 import { pollWindowReady } from './util'
+
+// Mirror the Tailwind theme `:root` / `.dark` --background values so that
+// the brief window between BrowserWindow creation and the first paint of
+// index.html doesn't flash white-on-dark (or vice versa).
+const BG_LIGHT = '#f9faf9'
+const BG_DARK = '#2e2e2e'
 
 // Forge environment variables
 declare const MAIN_WINDOW_VITE_DEV_SERVER_URL: string | undefined
@@ -160,6 +166,7 @@ export async function createMainWindow(
       height: options.height || DEFAULT_WINDOW_HEIGHT,
       show: options.show ?? !shouldStartHidden,
       autoHideMenuBar: true,
+      backgroundColor: nativeTheme.shouldUseDarkColors ? BG_DARK : BG_LIGHT,
       webPreferences: {
         contextIsolation: true,
         preload: path.join(__dirname, 'preload.js'),
--- a/main/src/tests/toolhive-manager.test.ts
+++ b/main/src/tests/toolhive-manager.test.ts
@@ -14,6 +14,8 @@ import {
   stopToolhive,
   restartToolhive,
   binPath,
+  probeSocket,
+  waitForSocketReady,
 } from '../toolhive-manager'
 import { updateTrayStatus } from '../system-tray'
 import log from '../logger'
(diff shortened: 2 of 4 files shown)


peppescg:
Closing in favor of an enterprise-overlay-only fix that doesn't require modifying upstream window lifecycle. The enterprise overlay handles the cold-start flash via:

1. Pushing `--hidden` to argv so `BrowserWindow` is constructed with `show: false`
2. Overriding `window.show()` until the renderer signals it has committed its first React render (via a new `enterprise:renderer-ready` IPC)
3. Keeping the React `LoadingScreen` mounted continuously until the first route resolves (via `router.subscribe('onResolved')`)

See https://github.com/stacklok/stacklok-enterprise-platform/pull/1407 for the enterprise-side implementation.

If maintainers later decide they want similar improvements upstream (the `startToolhive()` socket-await change in particular is independently useful — it eliminates the latent ENOENT race even on OSS), happy to re-open or open a focused follow-up PR for just that piece.

PR #2332, opened 2026-06-25
fix: default network isolation to enabled for new workloads

### LINKED ISSUE #2326: Align workload network isolation with backend default (enable by default)
## Summary

The ToolHive backend now defaults **network isolation ON** for locally-run MCP servers, for both `thv run` and the REST API. When an API client omits the `network_isolation` field on `POST /api/v1beta/workloads`, the backend now treats it as enabled (egress routed through a proxy, Docker gateway / `host.docker.internal` blocked, general internet still allowed under the default profile).

Studio currently **overrides this default** by explicitly sending `network_isolation: false` in nearly every workload-create payload. As a result, servers launched from Studio continue to run with no egress filtering, and Studio users do not benefit from the new backend hardening. This issue tracks aligning Studio with the backend default.

## Current behavior

Studio sets `network_isolation` explicitly (defaulting to `false`) in all three create paths:

- Local/custom servers: `renderer/src/features/mcp-servers/lib/orchestrate-run-local-server.tsx:91` (form default `false` at `renderer/src/features/mcp-servers/components/dialog-form-local-mcp.tsx:78`)
- Registry servers: `renderer/src/features/registry-servers/lib/orchestrate-run-registry-server.tsx:51` (form default `false` at `renderer/src/features/registry-servers/components/form-run-from-registry/index.tsx:136`)
- MCP Optimizer: hardcoded `network_isolation: false` at `renderer/src/features/meta-mcp/hooks/use-create-optimizer-workload.ts:130`
- Update path: `renderer/src/features/mcp-servers/lib/orchestrate-run-local-server.tsx:281`

Because an explicit `false` overrides the backend nil-default, the new default never applies to Studio-managed workloads.

The isolation UI already exists (the "Network Isolation" tab with an enable switch plus Allowed Hosts / Allowed Ports), and registry entries already carry `permissions.network.outbound.allow_host` that the form pre-fills. Only the default value of the switch and the wiring of those allow-lists need to change.

## Proposed changes

1. **Flip the default to enabled.** Default the `networkIsolation` form value to `true` for local and registry create flows, and stop hardcoding `false` where it is sent unconditionally. When the user has not customized anything, omit `network_isolation` (or send `true`) so workloads inherit the hardened default.

2. **Wire registry allow-lists through by default.** When launching a registry server with isolation on, populate `permission_profile.network.outbound.allow_host` / `allow_port` from the registry entry's declared permissions so the server keeps the outbound access it needs out of the box, rather than only when the user manually opens the tab.

3. **Handle custom/local servers gracefully.** Custom servers have no allow-list source. Decide the UX: keep `insecure_allow_all` (isolation on, but outbound permitted, which still blocks the Docker gateway) as the default for custom servers, or surface a clear prompt to configure allowed hosts. Avoid silently dropping all egress (the existing tab already warns about this).

4. **Reconcile the MCP Optimizer with isolation.** The Optimizer needs to reach the local ToolHive API. Today it works only because it pairs host access with `network_isolation: false` (`use-create-optimizer-workload.ts:115-138`): on Linux it uses `network.mode: 'host'` + `TOOLHIVE_HOST=127.0.0.1`, and on Docker Desktop it relies on `host.docker.internal`. Note that with isolation enabled, the Docker gateway is blocked and there is currently **no REST-exposed opt-out** for it (`--allow-docker-gateway` is CLI-only on the backend). So the Optimizer must either continue to run with isolation off, or move to `network.mode: 'host'` on all platforms. This needs an explicit decision and likely a coordinated backend change if we want isolation-on + host access via the API.

## Notes / dependencies

- The host to per-server proxy ingress path is unchanged by isolation, so the chat agent still reaches each server at `127.0.0.1:<port>`. No change is needed to endpoint discovery.
- Under the default profile, general outbound internet still works once isolation is on. The visible behavior change for most servers is only that the Docker gateway / host-local services become unreachable.
- Each isolated workload spins up additional helper containers (DNS + egress proxy, plus an ingress proxy for HTTP transports), so there is a resource/startup-time cost worth validating in the desktop experience.
- Potential backend follow-up: expose an `allow_docker_gateway` equivalent on the `POST /api/v1beta/workloads` request so clients can keep isolation on while permitting host access (needed for a clean Optimizer story). File separately on the `stacklok/toolhive` repo if we go that route.

## Acceptance criteria

- New local and registry servers launched from Studio default to network isolation enabled.
- Registry servers retain required outbound access automatically via their declared allow-lists.
- The MCP Optimizer continues to function with a documented, deliberate networking configuration.
- Users can still disable isolation or customize allowed hosts/ports through the existing UI.

peppescg:
## Summary
Screenshot 2026-06-25 at 16 15 44

Aligns Studio with the ToolHive backend default: network isolation is now **enabled by default** for all new MCP server workloads.

- Default `networkIsolation` form value flipped from `false` → `true` in both the local-server and registry-server create dialogs
- `insecure_allow_all` is now computed dynamically: `true` when isolation is on but no allow-list is configured (prevents silently blocking all egress for custom servers), `false` when an explicit allow-list is provided
- Applied consistently across all three code paths: create-local, update-local, create-registry

Closes #2326

## Background

The backend now defaults `network_isolation: true` when the field is omitted from `POST /api/v1beta/workloads`. Studio was overriding this by always sending `network_isolation: false`, so users never benefited from the hardened default.

The `insecure_allow_all: true` case (isolation on, all outbound permitted except Docker gateway) maps directly to the backend's built-in `"network"` permission profile. Custom/local servers that have no declared allow-list get this profile by default, preserving general internet connectivity while still blocking the Docker gateway.

Registry servers retain their declared outbound permissions automatically — the form already pre-fills `allowedHosts`/`allowedPorts` from the registry entry's `permissions.network.outbound` metadata, so flipping the default is sufficient.

The MCP Optimizer referenced in the issue (`use-create-optimizer-workload.ts`) no longer exists in the codebase — that feature was previously removed. No action needed there.

## Test plan

- [x] `prepareCreateWorkloadData` with isolation on + empty allow-lists → `insecure_allow_all: true`
- [x] `prepareCreateWorkloadData` with isolation on + populated allow-lists → `insecure_allow_all: false`
- [x] `prepareUpdateLocalWorkloadData` with isolation on + empty allow-lists → `insecure_allow_all: true`
- [x] Local server form defaults `networkIsolation` to `true`
- [x] Registry server form defaults `networkIsolation` to `true`
- [x] Full test suite: 2454 tests pass

### DIFF
--- a/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
+++ b/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
@@ -329,6 +329,52 @@ describe('DialogFormLocalMcp', () => {
     })
   })
 
+  it('defaults networkIsolation to true for a new server', async () => {
+    const mockInstallServerMutation = vi.fn()
+    mockUseRunCustomServer.mockReturnValue({
+      installServerMutation: mockInstallServerMutation,
+      isErrorSecrets: false,
+      isPendingSecrets: false,
+    })
+
+    renderWithProviders(
+      <Wrapper>
+        <DialogFormLocalMcp isOpen closeDialog={vi.fn()} groupName="default" />
+      </Wrapper>
+    )
+
+    await waitFor(() => {
+      expect(screen.getByRole('dialog')).toBeVisible()
+    })
+
+    // Fill minimal required fields and submit
+    await userEvent.type(
+      screen.getByRole('textbox', { name: /name/i }),
+      'test-server'
+    )
+    await userEvent.click(screen.getByLabelText('Transport'))
+    await userEvent.click(screen.getByRole('option', { name: 'stdio' }))
+    await userEvent.type(
+      screen.getByRole('textbox', { name: 'Docker image' }),
+      'ghcr.io/test/server'
+    )
+
+    await userEvent.click(
+      screen.getByRole('button', { name: 'Install server' })
+    )
+
+    await waitFor(() => {
+      expect(mockInstallServerMutation).toHaveBeenCalledWith(
+        {
+          data: expect.objectContaining({
+            networkIsolation: true,
+          }),
+        },
+        expect.any(Object)
+      )
+    })
+  })
+
   it('submits package manager form with all fields', async () => {
     const mockInstallServerMutation = vi.fn()
     mockUseRunCustomServer.mockReturnValue({
--- a/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
+++ b/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
@@ -1039,7 +1039,7 @@ describe('DialogFormLocalMcp', () => {
       await waitFor(() => {
         expect(
           screen.getByText(
-            'This configuration blocks all outbound network traffic from the MCP server.'
+            'All outbound traffic is permitted. Add specific hosts or ports below to restrict network access.'
           )
         ).toBeInTheDocument()
       })
@@ -1052,7 +1052,7 @@ describe('DialogFormLocalMcp', () => {
       await waitFor(() => {
         expect(
           screen.queryByText(
-            'This configuration blocks all outbound network traffic from the MCP server.'
+            'All outbound traffic is permitted. Add specific hosts or ports below to restrict network access.'
           )
         ).not.toBeInTheDocument()
       })
--- a/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
+++ b/renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
@@ -806,7 +806,6 @@ describe('DialogFormLocalMcp', () => {
       const networkTab = screen.getByRole('tab', { name: /network isolation/i })
       await userEvent.click(networkTab)
 
-      // Network isolation is enabled by default; no need to toggle the switch
       const addHostBtn = screen.getByRole('button', { name: /add a host/i })
       await userEvent.click(addHostBtn)
       const hostInput = screen.getByLabelText('Host 1')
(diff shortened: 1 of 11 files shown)


renderer/src/features/mcp-servers/components/local-mcp/__tests__/dialog-form-local-mcp.test.tsx
```diff
@@ -760,11 +806,7 @@ describe('DialogFormLocalMcp', () => {
       const networkTab = screen.getByRole('tab', { name: /network isolation/i })
       await userEvent.click(networkTab)
 
-      const switchLabel = screen.getByLabelText(
-        'Enable outbound network filtering'
-      )
-      await userEvent.click(switchLabel)
-
+      // Network isolation is enabled by default; no need to toggle the switch
```
kantord: an agent or human reading this in the future won't remember that this click was ever there, so they won't have any questions about it

instead of mentioning it in every test case, it may be better to communicate this behavior through a test case name, like 

```
 it('enables network isolation for all MCPs by default'
```

renderer/src/features/mcp-servers/lib/orchestrate-run-local-server.tsx
```diff
@@ -278,11 +289,10 @@ export function prepareUpdateLocalWorkloadData(
...
             outbound: {
-              allow_host: data.allowedHosts?.map((host) => host.value),
-              allow_port: data.allowedPorts?.map((port) =>
-                parseInt(port.value, 10)
-              ),
-              insecure_allow_all: false,
+              allow_host: filteredHosts,
+              allow_port: filteredPorts,
+              insecure_allow_all:
+                filteredHosts.length === 0 && filteredPorts.length === 0,
```
kantord, line 295: so in other words

```
const allowsAllNetworkAccess = hasNoAllowedPorts & hasNoAllowedHosts
```

this is confusing because it is equivalent to "if no ports or hosts are allowed, then all ports and hosts are allowed". 

So what I'm missing is an explicit 

```
const isNetworkIsolationEnabled =  .... 
```

that is derived from an actual form state, which we should have, because seems like we have an actual form field for this.

I'm not totally sure if there is an uncovered situation here (network isolation despite those fields being empty), but in any case if that invariant exists, I would probably enforce it as a validation, considering the explicit form field to be the authoritative arbiter of truth

eleftherias (changes requested):
Blocking a merge until we do further testing on the backend

eleftherias:
@kd Should we close this PR since you have already done this work in other PRs?

kantord:
partly folded into & superseded by: https://github.com/stacklok/toolhive-studio/pull/2469

PR #2335, opened 2026-06-26
fix: discover proxied stdio tools with raw MCP transport

### LINKED ISSUE #2333: [Bug] Playground fails to detect tools for stdio-based MCP servers
### Bug Description

In the Playground, enabling most MCP servers fails with a "Failed to detect tools for <server>" toast. The main-process log shows repeated `Error invoking remote method 'chat:get-mcp-server-tools': reply was never sent`. Fetch works, but context7, github, playwright, and most other servers do not.

The root cause is in tool discovery. The Playground connects to each server through `@ai-sdk/mcp`'s `createMCPClient` using the `{ type: 'http', url }` transport config (`createTransport` in `main/src/utils/mcp-tools.ts`). ToolHive's stdio→streamable-http proxy returns JSON-RPC responses as SSE frames with a `data:` line but **no `event:` line**. Per the WHATWG SSE spec this dispatches as a `message` event, but `@ai-sdk/mcp`'s HTTP transport only forwards frames where `event === "message"`, and its `EventSourceParserStream` reports `event: undefined` for event-less frames. The `initialize` response is dropped, `createMCPClient()` never resolves, the IPC handler hangs, and Electron reports `reply was never sent`.

**Why Fetch is the exception:** Fetch and toolhive-doc-mcp are native streamable-http servers whose SSE frames include an explicit `event: message` line, so the AI SDK parses them. The failing servers are stdio servers fronted by ToolHive's proxy, which emits event-less frames.

**Suspected regression:** `eccd4338` switched streamable-http from a raw `StreamableHTTPClientTransport` instance to the AI SDK `{ type: 'http' }` config form (to avoid a `protocolVersion` getter-only crash). That swap exposed the AI SDK's event-less-SSE handling.

### Steps to Reproduce

1. Run any stdio-based MCP server (e.g. context7, github, playwright) in ToolHive Studio.
2. Open the Playground and open the MCP server selector.
3. Toggle the server on to enable its tools.
4. See the "Failed to detect tools for <server>" toast.

### Expected Behavior

The server's tools are discovered and enabled, the same as for Fetch.

### Actual Behavior

Tool discovery hangs and the toggle fails with "Failed to detect tools". The log records `Error invoking remote method 'chat:get-mcp-server-tools': reply was never sent`.

### Priority

High

### Environment

- OS: macOS (darwin/arm64)
- App Version: ToolHive Studio desktop 0.36.4
- Bundled `thv`: v0.31.0

### Additional Context

Evidence, all against the same live ToolHive servers:

| Client | Result |
| --- | --- |
| `thv mcp list tools` | works |
| `curl` POST initialize | works |
| raw `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport` | connects + lists tools (~40ms) |
| `@ai-sdk/mcp` vs stdio-proxied server (data-only SSE) | **hangs in `init()`** |
| `@ai-sdk/mcp` vs native server (`event: message`) | works (~30ms) |

Confirmed the AI SDK's `EventSourceParserStream` yields `event: undefined` for a `data:`-only frame and `"message"` only when the `event:` line is explicit.

**Proposed fix:** use the raw `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport` for HTTP discovery (as `buildRawTransport` already does for MCP Apps) instead of the AI SDK's HTTP transport, with a different workaround for the original `protocolVersion` crash. Alternatively, report upstream to `@ai-sdk/mcp` that its HTTP transport should treat event-less SSE frames as `message` per spec.

jstar0:
## Summary

Fixes #2333.

Fixes Playground tool discovery for stdio workloads that are exposed through the Streamable HTTP proxy.

The existing discovery path uses the AI SDK MCP HTTP client for all streamable-http workloads. For stdio-backed workloads proxied by ToolHive, the proxy can return event-less SSE `data:` frames. The raw MCP SDK transport handles that response shape, but the AI SDK discovery path can hang before returning tools.

## Changes

This change keeps the existing AI SDK discovery path for normal streamable-http workloads, and only switches stdio + streamable-http proxy discovery to the raw MCP SDK client through the existing `buildRawTransport(workload)` helper.

## Verification

```text
pnpm exec vitest run --dir main/src/utils main/src/utils/__tests__/mcp-tools.test.ts
pnpm run type-check
pnpm exec prettier --check main/src/utils/mcp-tools.ts main/src/utils/__tests__/mcp-tools.test.ts
pnpm exec eslint --max-warnings 0 main/src/utils/mcp-tools.ts main/src/utils/__tests__/mcp-tools.test.ts
```

### DIFF
--- a/main/src/utils/__tests__/mcp-tools.test.ts
+++ b/main/src/utils/__tests__/mcp-tools.test.ts
@@ -1,7 +1,40 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest'
-import { createTransport } from '../mcp-tools'
+import {
+  createTransport,
+  getWorkloadAvailableTools,
+  buildRawTransport,
+} from '../mcp-tools'
 import type { GithubComStacklokToolhivePkgCoreWorkload as CoreWorkload } from '@common/api/generated/types.gen'
 import { Experimental_StdioMCPTransport } from '@ai-sdk/mcp/mcp-stdio'
+
+const mockAiMcpClient = vi.hoisted(() => ({
+  tools: vi.fn().mockResolvedValue({}),
+  close: vi.fn().mockResolvedValue(undefined),
+}))
+
+const mockCreateMCPClient = vi.hoisted(() =>
+  vi.fn().mockResolvedValue(mockAiMcpClient)
+)
+
+const mockSdkClient = vi.hoisted(() => ({
+  connect: vi.fn().mockResolvedValue(undefined),
+  listTools: vi.fn().mockResolvedValue({ tools: [] }),
+  close: vi.fn().mockResolvedValue(undefined),
+}))
+
+vi.mock('@ai-sdk/mcp', async (importOriginal) => {
+  const original = await importOriginal<typeof import('@ai-sdk/mcp')>()
+  return {
+    ...original,
+    experimental_createMCPClient: mockCreateMCPClient,
+  }
+})
+
+vi.mock('@modelcontextprotocol/sdk/client/index.js', () => ({
+  Client: function ClientMock() {
+    return mockSdkClient
+  },
+}))
 
 vi.mock('../../logger', () => ({
   default: {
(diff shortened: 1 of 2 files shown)


peppescg:
@jstar0 Thanks for the fix 🙏 , we will take a look asap

samuv:
thank you @jstar0 the bug has resolved with the bump of the library and we plan a refactoring asap, so I will close this

PR #2425, opened 2026-07-06
fix: avoid downgrade prompts for newer server tags

### LINKED ISSUE #2383: [Bug] MCP update feature suggests a downgrade
### Bug Description

If you run a newer version of an MCP server than what's in the registry, the desktop app suggests a downgrade to match the registry. Seems it's doing a simple match and not a semver-aware check?

In my case, I edited my playwright server to bump it to v0.0.77, but the UI is prompting me to downgrade to v0.0.70.

### Steps to Reproduce

1. Run Playwright from the registry
2. Edit the workload to bump the image from the registry's v0.0.70 to v0.0.77
3. Note the update icon appears, suggesting a downgrade

Image

### Expected Behavior

The app should do a semver-aware check and only suggest an update if the registry version is newer than the running version.

### Actual Behavior

_No response_

### Priority

Low

### Environment

App version: 0.36.6

### Additional Context

_No response_

fengjikui:
## Summary

- make registry update prompts semver-aware when both image tags are valid semver
- keep the existing tag-drift fallback for non-semver tags like `latest` or custom labels
- add a regression test for the `v0.0.77` running image vs `v0.0.70` registry image downgrade case

Fixes #2383.

## Validation

- `pnpm exec vitest run renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `pnpm exec vitest run renderer/src/features/mcp-servers/components/__tests__/card-mcp-server.test.tsx renderer/src/features/mcp-servers/components/__tests__/table-mcp-servers.test.tsx`
- `pnpm run type-check`
- `pnpm exec eslint renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `git diff --check`


samuv (commented):
**Summary**

This is a focused, well-scoped fix: it swaps the naive tag-equality drift check for a semver-aware comparison so users aren’t prompted to “upgrade” to an older server tag, while keeping the previous string-based behavior for non-semver tags like `latest`.

**What works well**

- `hasRegistryTagUpdate` is small and has a clear fallback path for non-semver tags.
- The regression test for `v0.0.77` running vs `v0.0.70` registry directly covers the reported issue.
- The existing test with `v1.0.0` vs `v2.0.0` already proves that major bumps still trigger drift, so `semverGt` is the right comparison here.
- Using `semver.valid` normalizes `v` prefixes before comparing, which is a nice touch.

renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx
```diff
@@ -125,6 +125,34 @@ describe('useIsServerFromRegistry', () => {
       })
     })
 
+    it('does not report drift when the running semver tag is newer than the registry tag', async () => {
```
samuv, line 128: `praise` — Clear, focused regression test that captures the exact bug from #2383. The test name makes the expected behavior obvious.

renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx
```diff
@@ -125,6 +125,34 @@ describe('useIsServerFromRegistry', () => {
...
+        () => useIsServerFromRegistry('playwright'),
+        { wrapper }
+      )
+
+      await waitFor(() => {
+        expect(result.current.isFromRegistry).toBe(true)
+        expect(result.current.registryTools).toEqual(['browser-navigate'])
+        expect(result.current.drift).toBeNull()
+      })
+    })
```
samuv, line 154: `suggestion` — Would you mind adding a sibling test for the fallback path where one tag is semver and the other isn’t? For example, running `latest` against registry `v1.0.0` or vice versa. That would document the intended fallback behavior and protect it from a future refactor.

renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts
```diff
@@ -27,6 +28,19 @@ function getImageTag(image: string | undefined): string {
...
+  const localVersion = semverValid(localTag)
+
+  if (registryVersion && localVersion) {
+    return semverGt(registryVersion, localVersion)
+  }
+
+  return true
+}
+
 export function useIsServerFromRegistry(serverName: string) {
```
samuv, line 45: `suggestion` — This fallback preserves the old behavior, which is good. A short inline comment like `// Non-semver tags fall back to the previous "any difference is drift" behavior` would help the next reader understand the intent without reverse-engineering it.

renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts
```diff
@@ -27,6 +28,19 @@ function getImageTag(image: string | undefined): string {
   return lastColonIndex !== -1 ? image.slice(lastColonIndex + 1) : ''
 }
 
+function hasRegistryTagUpdate(registryTag: string, localTag: string): boolean {
+  if (registryTag === localTag) return false
+
+  const registryVersion = semverValid(registryTag)
+  const localVersion = semverValid(localTag)
+
+  if (registryVersion && localVersion) {
```
samuv, line 37: `question` — You chose `semver.valid` over `semver.coerce`. That means something like `v1.0` would fall back to string drift rather than being parsed as `1.0.0`. Is that intentional for this registry’s tagging conventions? Not a blocker, just confirming the choice was deliberate.

fengjikui:
Pushed `92efc9c1` addressing the review suggestions: added a sibling non-semver fallback regression (`latest` vs `v1.0.0`) and a short comment documenting that non-semver tags preserve the previous drift behavior.

Validation run locally:
- `pnpm exec vitest run renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx` (45 tests)
- `pnpm exec prettier --check renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `pnpm exec eslint renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `pnpm run type-check`
- `git diff --check`

I also checked the failing `label-pr` job; it is the fork token failing to add `size/XS` with `Resource not accessible by integration`, while the app/unit/static checks were green on the previous commit.

samuv:
@fengjikui thank you for the extra tests, could you rebase from main? That way I can verify the PR Size Labeler GitHub Action runs correctly now ( I made the change that should fix the run [here](https://github.com/stacklok/toolhive-studio/pull/2428))

fengjikui:
Rebased onto current `main` (`8b67315f`) and force-pushed the PR branch. The PR Size Labeler reran successfully and `label-pr` is now green on head `9ecb3196`.

Fresh local validation after the rebase:
- `pnpm exec vitest run renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx` (45 tests)
- `pnpm run type-check`
- `pnpm exec eslint --max-warnings 0 renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `pnpm exec prettier --check package.json pnpm-lock.yaml renderer/src/features/mcp-servers/hooks/use-is-server-from-registry.ts renderer/src/features/mcp-servers/hooks/__tests__/use-is-server-from-registry.test.tsx`
- `git diff --check`

Note: the repo's `test:nonInteractive` Electron runner could not start locally because the Electron binary install/download failed before tests executed, so I used direct Vitest for this hook test.

samuv (approved):
thank you, looks good!

PR #2469, opened 2026-07-15
fix: make network isolation mode explicit instead of inferred

### LINKED ISSUE #2326: Align workload network isolation with backend default (enable by default)

kantord:
https://github.com/user-attachments/assets/1bb0ee77-6a6f-4806-8702-d1e39b2326a8

## Summary

Aligns Studio with the ToolHive backend's hardened default (network isolation on for new workloads) and resolves the design flaw flagged in review on #2332.

- Replaces the boolean `networkIsolation` toggle with an explicit `networkAccess: 'none' | 'host' | 'proxy'` mode, plus an explicit `allowedDestinations: 'anywhere' | 'selected'` field that now directly drives `insecure_allow_all` — instead of inferring it from whether the allow-list arrays happened to be empty (the ambiguity Copilot and @kantord flagged on #2332: a registry entry that intentionally declares zero outbound permissions could previously get silently treated as "allow all").
- Wires up the previously-unused backend field `permission_profile.network.mode: "host"`, giving the new "Host networking (Advanced)" option a real effect.
- Consolidates two near-duplicate `NetworkIsolationTabContent` components (local + registry forms) into one shared, generic `NetworkAccessTabContent`.
- New UI follows the attached mockup: 3-way network access radio, nested "Allowed destinations" radio (Anywhere / Selected), and an "Allow host machine access" checkbox. That checkbox is UI-only for now — the backend has no REST-reachable `allow_docker_gateway` field on create/update yet (confirmed against `stacklok/toolhive` main and the latest release); a companion backend change exists in a separate worktree to close that gap, to be tracked as a follow-up once merged upstream.

Closes #2326

## Test plan

- [x] `pnpm run type-check` clean
- [x] `pnpm run lint` clean
- [x] `pnpm run test:nonInteractive` — 2527/2527 tests passing, including 17 new tests dedicated to `NetworkAccessTabContent`'s rendering/interactions
- [x] `pnpm run knip` — no new dead code
- [ ] Manual verification in a running app (blocked locally by an unrelated devcontainer toolchain issue — `better-sqlite3` fails to compile against Electron 43's V8 headers; not caused by this change)

Fully or partially written by an AI agent.


renderer/src/common/lib/form-schema-mcp.ts
```diff
@@ -179,22 +205,31 @@ const createProxyConfigSchema = () => {
...
     return
   }
 
   // Validate allowedHosts only when network isolation is enabled
   networkData.allowedHosts?.forEach((host, index) => {
     if (
       host.value.trim() !== '' &&
-      !/^\.?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/.test(host.value)
+      !/^\.?([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/.test(host.value) &&
+      !isValidIpv4Address(host.value)
```
eleftherias: From my understanding the backend won't actually work properly with ipv4 addresses. Was there a reason for this change?

eleftherias (commented):
The network mapping logic looks correct. I left one comment about ipv4 addresses.

renderer/src/common/lib/form-schema-mcp.ts
kantord: makes sense

kantord: it seems like the BE doesn't do any validation here though? is that a problem? it seems like this line write the strings into the rules without validation? https://github.com/stacklok/toolhive/blob/90539bbb40fe56c934b2c3a09a14fffa4aa95043/pkg/container/docker/squid.go#L326-L330

So probably the BE _should_ validate here? 

in any case, this validation doesn't need to live in FE so i'll just remove it

kantord: removed!

eleftherias (approved):

PR #2475, opened 2026-07-17
refactor(chat): migrate main-process chat to Effect domain services

### LINKED ISSUE #2474: refactor(chat): migrate main-process chat to Effect domain services
## Context

The main-process chat subsystem (`main/src/chat`) grew into a high-concurrency orchestration layer: streaming turns, MCP client lifecycles, persistence snapshots, pricing refresh, provider discovery, and app shutdown cleanup. The previous implementation solved this with **module-level mutable state**, **Promise chains**, and **ad-hoc cleanup branches**. That worked while the surface was small, but it made failure modes, cancellation, and resource ownership hard to reason about — and easy to get wrong (e.g. competing persistence writers, quit-time races after teardown).

This issue tracks migrating chat onto a **stable Effect-based architecture** while keeping the preload/renderer IPC contract unchanged.

## Why the previous approach was struggling

| Pain point | What it looked like before |
| --- | --- |
| **Scattered lifecycle** | Active streams lived in a module `Map`; `app.on('before-quit')` and WebContents `destroyed` handlers raced with teardown |
| **Competing persistence** | Stream registry *and* `toUIMessageStream.onEnd` both wrote thread messages — double-flush / ordering bugs |
| **Opaque failures** | Mixed thrown `Error`s, `{ success: false }`, and silent catches with no shared user-facing mapping |
| **Global caches** | Pricing and MCP UI metadata as module singletons — hard to reset in tests, hard to reason about freshness |
| **Partial MCP success** | Discovery failures were inconsistent: sometimes wipe metadata, sometimes continue, no clear “fail the turn if nothing usable” policy |
| **Test fragility** | Broad `vi.mock` graphs + module reset hooks instead of injectable Test Layers |

These are classic symptoms of effectful orchestration without a runtime: cancellation, scoped resources, and typed errors are bolted on rather than composed.

## Why Effect is a better fit

Effect gives us the primitives this subsystem actually needs — without rewriting the AI SDK chunk pipe (we keep Web `ReadableStream` as the data plane).

### 1. One managed runtime instead of import-time side effects
- App-scoped `ManagedRuntime` bootstraps after DB reconcile and disposes on quit.
- If init fails, the rest of the app still starts; chat returns a safe typed **unavailable** error instead of crashing into half-initialized globals.

### 2. Typed domain errors instead of stringly failures
- Expected failures are `Data.TaggedError`s (`StorageError`, `ProviderError`, `McpDiscoveryError`, `ChatUnavailableError`, …) with an internal cause **and** a safe `userMessage`.
- Adapters map those at the IPC boundary so the renderer never sees stack dumps or Fiber internals.

### 3. Scoped resources and Fibers instead of manual cleanup
- Each chat turn runs as a **scoped Fiber**; cancel/interrupt propagates to the AI SDK `AbortController`.
- MCP clients and built-in tool handles share one cleanup path for success, failure, interruption, **and** app shutdown — no duplicate `try/finally` forests.

### 4. Explicit ownership of stream persistence
- `StreamRegistry` / session is the **sole** message-persistence owner (throttled snapshots + exactly one final flush).
- Removes the competing `onEnd` write that previously raced with registry flushes.

### 5. Runtime-owned state instead of module globals
- Pricing cache and MCP UI metadata move to runtime `Ref`s (hydrated from disk/SQLite).
- Bounded retry/schedule only for **idempotent** reads (pricing/provider discovery) — never auto-retry model generation or tool execution.

### 6. Domain services with Live/Test Layers
- Chat is reorganized into `runtime`, `threads`, `settings`, `agents`, `providers`, `pricing`, `mcp`, `streaming`.
- Callers depend on services, not on import order. Tests install a test runtime instead of resetting module state.

### 7. Safer degradation policies
- MCP: continue with healthy enabled servers on partial discovery failure; **fail the turn** when enabled servers were requested but none can be used.
- Settings reads stay pure; reconciliation of stopped-server tool entries is an **explicit** effect after workload discovery / settings updates.
- Legacy `electron-store` construction is lazy so reconciliation and chat runtime init no longer fight at import time.

## What we implemented

Phased migration (IPC/preload/renderer unchanged):

1. **Foundation** — `effect` dependency, tagged errors, logging bridge, ManagedRuntime, lifecycle wiring (`when-ready` / `block-quit`), `chat:runtime:health` probe.
2. **Persistence** — Threads, settings, thread settings, agents as Effect services; lazy legacy-store access for one-time SQLite reconciliation.
3. **Providers / pricing / titles** — Injectable clients, Ref-backed pricing with deduped refresh, title generation as a domain service.
4. **MCP** — Discovery/proxy/metadata split; Ref-backed UI metadata cache; fail-closed when no enabled tools can be loaded.
5. **Streaming** — Registry + Fiber-owned turns; sole persistence owner; destroy/quit cleanup that does **not** require a live runtime (avoids uncaught `ChatUnavailableError` on quit).
6. **Cleanup** — Thin facades at old paths; barrel exports reduced to shared types/pure helpers; IPC imports explicit domain entry points.

## Non-goals

- No renderer/preload API changes.
- No `@effect/platform` (unstable surface) — stable `effect` core only.
- Pure helpers (`usage.ts`, `utils.ts`, constants, Zod IPC/tool schemas) stay plain TypeScript.

## Acceptance criteria

- [x] Chat IPC channel names, payloads, and fire-and-forget stream startup remain stable
- [x] Runtime init failure leaves the app usable; chat returns a typed unavailable error
- [x] StreamRegistry is the sole message-persistence owner (no competing `onEnd` write)
- [x] MCP: partial discovery degrades; empty usable tools with enabled servers fails the turn
- [x] Quit/destroy cleanup does not throw after runtime disposal
- [x] Focused Electron Vitest suites + `pnpm run type-check` pass for the chat surface
- [ ] Follow-up: remove remaining compatibility facades once all callers use domain entry points
- [ ] Follow-up: playground E2E smoke with Docker/ToolHive available

## Commits (branch `chat-effect`)

1. `chore(deps): add effect`
2. `refactor(chat): migrate chat subsystem to Effect domain services`
3. `feat(chat): expose Effect runtime health over IPC`
4. `test(chat): adapt unit tests for Effect chat runtime`

## References

- Effect docs: [Runtime](https://effect.website/docs/runtime/), [Scope](https://effect.website/docs/resource-management/scope/), [Cancellation](https://effect.website/docs/getting-started/creating-effects/)
- Related prior art in-repo: AI SDK v7 streaming migration (#2368) — this continues hardening the same orchestration boundary

samuv:
## Summary

Closes #2474

Migrates the main-process chat subsystem onto a stable Effect architecture.

- Add the `effect` package and an app-scoped `ManagedRuntime` with tagged domain errors and lifecycle wiring
- Reorganize chat into `runtime`, `threads`, `settings`, `agents`, `providers`, `pricing`, `mcp`, and `streaming` domain services behind thin compatibility facades
- Make `StreamRegistry` the sole message-persistence owner; run chat turns as scoped Fibers; keep quit/destroy cleanup safe after runtime disposal
- Move pricing and MCP UI metadata to runtime-owned `Ref` state with clearer MCP degradation / fail-closed policies
- Soft-fail unread IPC facades where safe; fail closed for thread message hydration (unavailable + StorageError) to avoid overwriting history
- `chat:agents:duplicate` now returns `{ success, agent?, error? }` (preload + renderer updated) so failures are not silent no-ops

## Test plan

- [x] `pnpm run type-check`
- [x] Focused Electron Vitest suites for `main/src/chat` and chat IPC handlers
- [x] Manual: quit the app without uncaught `ChatUnavailableError` from `purgeSender`
- [x] Automated (`active-streams.test.ts`): cancel mid-stream persists partial snapshot; shutdown flushes debounced write
- [x] Automated (`mcp-tools.test.ts`): partial discovery continues when one server fails; fail-closed when no usable tools remain
- [ ] Playground E2E smoke with Docker/ToolHive available


main/src/chat/streaming/chat-stream-service.ts
```diff
@@ -0,0 +1,52 @@
...
+                      sender
+                    ),
+                  catch: (cause) =>
+                    cause instanceof Error ? cause : new Error(String(cause)),
+                })
+              )
+
+              yield* Effect.addFinalizer(() => Fiber.interrupt(fiber))
+              return yield* Fiber.await(fiber)
+            })
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — switched to `Fiber.join(fiber)` so stream fiber failures propagate through `handleChatStreamRealtime` instead of resolving with an `Exit`.

main/src/chat/settings-storage.ts
```diff
@@ -1,407 +1,93 @@
...
-      `[CHAT] Failed to save settings for provider ${providerId}:`,
-      error
-    )
-    return {
-      success: false,
-      error: error instanceof Error ? error.message : 'Unknown error',
-    }
-  }
+  return runChatSync(SettingsService.getChatSettings(providerId))
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `getChatSettings` now uses `runChatSyncOr(..., defaultChatSettings(providerId))`, so unread/unavailable runtime returns the prior safe default instead of throwing into IPC.

main/src/chat/settings-storage.ts
```diff
@@ -1,407 +1,93 @@
...
 export function getSelectedModel(): ChatSettingsSelectedModel {
-  try {
-    const dbModel = readSelectedModelFromDb()
-    if (dbModel.provider && dbModel.model) return dbModel
-  } catch (err) {
-    log.error('[DB] SQLite read failed:', err)
-  }
-  return { provider: '', model: '' }
+  return runChatSync(SettingsService.getSelectedModel())
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `getSelectedModel` now uses `runChatSyncOr(..., EMPTY_SELECTED_MODEL)` so unread runtime returns `{ provider: '', model: '' }` instead of rejecting IPC.

main/src/chat/settings-storage.ts
```diff
@@ -1,407 +1,93 @@
...
-// Get the set of user-scope skills the user has enabled via the toolbar.
 export function getEnabledSkills(): string[] {
-  try {
-    return readEnabledSkillsFromDb()
-  } catch (error) {
-    log.error('[DB] Failed to read enabled skills:', error)
-    return []
-  }
+  return runChatSync(SettingsService.getEnabledSkills())
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `getEnabledSkills` now uses `runChatSyncOr(..., [])` so unread runtime returns an empty list instead of throwing.

main/src/chat/settings-storage.ts
```diff
@@ -1,407 +1,93 @@
...
-        return tools && tools.length > 0
-      }
-    )
-    return enabledServerNames
-  } catch (error) {
-    log.error('Failed to get enabled MCP servers from tools:', error)
-    return []
-  }
+  return runChatPromise(SettingsService.getEnabledMcpServersFromTools())
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — both paths now use `runChatPromiseOr` with `{}` / `[]` fallbacks when the runtime isn’t ready, matching the previous degrade-to-empty behavior.

main/src/chat/settings-storage.ts
```diff
@@ -1,407 +1,93 @@
...
   installedNames: readonly string[]
 ): number {
-  try {
-    return deleteEnabledSkillsNotIn(installedNames)
-  } catch (error) {
-    log.error('[DB] Failed to prune enabled skills:', error)
-    return 0
-  }
+  return runChatSync(SettingsService.pruneEnabledSkillsTo(installedNames))
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `pruneEnabledSkillsTo` now uses `runChatSyncOr(..., 0)`, so unread runtime returns `0` (same soft-fail as before) and doesn’t reject `chat:get-enabled-skills`.

main/src/chat/threads-storage.ts
```diff
@@ -1,164 +1,43 @@
...
 
 export function getThread(threadId: string): ChatSettingsThread | null {
-  try {
-    return readThreadFromDb(threadId)
-  } catch (err) {
-    log.error(`[THREADS] Failed to get thread ${threadId}:`, err)
-    return null
-  }
+  return runChatSync(ThreadsService.getThread(threadId))
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `getThread` / `getAllThreads` now use `runChatSyncOr` with `null` / `[]` fallbacks when the runtime isn’t ready.

main/src/chat/threads-storage.ts
```diff
@@ -172,139 +51,48 @@ export function addMessageToThread(
...
 
 export function getActiveThreadId(): string | undefined {
-  try {
-    return readActiveThreadIdFromDb() ?? undefined
-  } catch (error) {
-    log.error('[THREADS] Failed to get active thread ID:', error)
-    return undefined
-  }
+  return runChatSync(ThreadsService.getActiveThreadId())
 }
```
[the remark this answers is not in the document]
samuv: Fixed in `4bb714dd` — `getActiveThreadId` now uses `runChatSyncOr(..., undefined)` so unread runtime returns `undefined` instead of throwing into IPC.

main/src/chat/agents/registry.ts
```diff
@@ -1,232 +1,55 @@
...
-    }
-  } catch (err) {
-    log.error('[AGENTS] Failed to seed built-in agents:', err)
-  }
+  runChatSync(AgentsService.seedBuiltinAgents())
 }
 
 export function listAgents(): AgentConfig[] {
-  return readAllAgents()
+  return runChatSync(AgentsService.listAgents())
```
kantord: Same pattern Copilot flagged and that got fixed in `settings-storage.ts`/`threads-storage.ts` (commit `4bb714dd`) shows up here too: `listAgents()` (and `getAgent`, `resolveAgentForThread`, etc. below) call the bare `runChatSync`, which throws if the chat runtime isn't ready. `chat:agents:list` / `chat:agents:get` (`main/src/ipc-handlers/chat/agents.ts:15,17`) have no try/catch, so a runtime hiccup during startup or a failed bootstrap can reject these IPC calls — same regression class as the fixed files, just not covered by that fix.

main/src/chat/active-streams.ts
```diff
@@ -1,805 +1,65 @@
...
-  if (!stream || stream.status !== 'streaming') return null
-  if (sender.isDestroyed()) return null
-  stream.subscribers.add(sender)
-  return {
-    streamId: stream.streamId,
-    replayChunks: buildReplayChunks(stream),
-    toolUiMetadata: stream.toolUiMetadata,
-  }
+export function subscribeToStream(chatId: string, sender: WebContents) {
+  return runChatSync(StreamRegistryService.subscribeToStream(chatId, sender))
```
kantord: Same issue as the `agents/registry.ts` comment, but higher-impact: `subscribeToStream` here and `cancelStream` at line 32 back `chat:stream:resume` / `chat:stream:cancel` (`main/src/ipc-handlers/chat/streaming.ts:64-77`), which also have no try/catch. If these throw during a runtime-not-ready window, resuming or cancelling an active chat stream fails outright — not just a cosmetic empty-list default like the already-fixed settings reads.

main/src/chat/mcp-tools.ts
```diff
@@ -1,442 +1,33 @@
...
+} from './mcp/mcp-service-impl'
 
-/** Extracts the `_meta.ui` block from a raw tool definition. */
-function extractToolUiMeta(
-  toolDef: unknown
-): { resourceUri?: string; visibility?: string[] } | undefined {
-  return (toolDef as { _meta?: Record<string, unknown> })?._meta?.['ui'] as
-    { resourceUri?: string; visibility?: string[] } | undefined
+export function getCachedUiMetadata() {
+  return runChatSync(McpService.getCachedUiMetadata())
```
kantord: Same bare `runChatSync`/`runChatPromise` pattern as the two files fixed in `4bb714dd` (here and at lines 25, 32). Worth checking whether the corresponding IPC handlers catch, or whether these should switch to `runChatSyncOr`/`runChatPromiseOr` with a safe default the way the settings/threads facades now do.

main/src/chat/providers/providers-facade.ts
```diff
@@ -0,0 +1,34 @@
...
+      models: provider.models,
+    })),
+  }
+}
+
+export async function fetchProviderModelsHandler(
+  providerId: string,
+  tempCredential?: string
+): Promise<{ id: string; name: string; models: string[] } | null> {
+  return runChatPromise(
```
kantord: Same pattern again — bare `runChatPromise` (also at line 33), no fallback if the runtime isn't ready yet.

main/src/chat/thread-integration.ts
```diff
@@ -1,128 +1,30 @@
...
-  }
-
-  if (!thread.messages || thread.messages.length === 0) {
-    return []
-  }
-
-  // Return stored messages directly
-  return thread.messages
+export async function getThreadMessagesForTransport(threadId: string) {
+  return runChatPromise(ThreadsService.getThreadMessagesForTransport(threadId))
```
kantord, line 31: Same pattern here too — `getThreadMessagesForTransport` (bare `runChatPromise`) and `getThreadInfo` at line 22 (bare `runChatSync`). This one is used to build the message history sent to the model, so a runtime-not-ready throw here risks breaking message sending itself, not just a read.

main/src/chat/streaming/stream-registry-service.ts
```diff
@@ -0,0 +1,884 @@
...
+    | { kind: 'text'; id: string }
+    | { kind: 'reasoning'; id: string }
+    | { kind: 'tool'; toolCallId: string }
+  >
+  textParts: Map<string, TextPartReplay>
+  reasoningParts: Map<string, ReasoningPartReplay>
+  toolParts: Map<string, ToolPartReplay>
+}
+
+const streams = new Map<string, ActiveStream>()
```
kantord: This is the file where Effect's structured-concurrency primitives would earn their cost the most, but the actual engine (`runManagedStream`, `pumpChunks`, this `streams` Map, the `persistTimer` debounce) is plain imperative code — Effect only wraps the edges (`Effect.sync`/`Effect.tryPromise` around lines 858-880).

Sibling services added in this same PR already lean on `Ref` for shared state (`pricing/pricing-service.ts:67-68`, `mcp/mcp-ui-metadata-cache.ts:20-21`) — applying that pattern here, plus forking the stream as a real `Fiber` with `Effect.addFinalizer` guarding `flushPersist`, would make "the debounced write always flushes on every exit path" a structural guarantee instead of three separate manual call sites (`finalizeSuccess`, `finalizeError`, `shutdownAllActiveStreams`) that each have to independently remember to call it. It's also exactly what the still-unchecked "cancel mid-stream, confirm persistence" test-plan item is trying to verify by hand — worth having the type system guarantee it instead.

kantord (commented):
**Disclosure:** This review comment (and the inline comments below) was fully or partially written by an AI agent, reading through the migration commits and cross-checking IPC handler wiring against the pattern Copilot already flagged. This is a **partial pass**, not a complete/exhaustive review of the PR — please don't treat an absence of a comment elsewhere as a clean bill of health.

What's covered below:

1. **The `runChatSyncOr`/`runChatPromiseOr` fix in `4bb714dd` doesn't appear to have been applied consistently.** It fixed exactly the 8 spots Copilot commented on (`settings-storage.ts`, `threads-storage.ts`), but the same bare-`runChatSync`/`runChatPromise` pattern — which throws/rejects when the chat runtime isn't ready — still exists in `agents/registry.ts`, `active-streams.ts`, `mcp-tools.ts`, `providers/providers-facade.ts`, and `thread-integration.ts`, and their IPC handlers (e.g. `chat:agents:list`, `chat:stream:resume`, `chat:stream:cancel`) have no try/catch either. Flagged individually below.
2. **Two of three manual test-plan checkboxes are still unchecked** ("cancel mid-stream, confirm cleanup and persistence" and "MCP partial discovery / fail-closed") — worth resolving before merge, especially since (1) touches exactly those code paths.
3. **A design suggestion on `stream-registry-service.ts`** — flagged inline.

Happy to expand this into a fuller pass if useful, just flagging what surfaced so far.

main/src/chat/agents/registry.ts
samuv: Fixed in `b241f0bf` — read paths (`listAgents`, `getAgent`, `resolveAgentForThread`, `getThreadAgentId`, `duplicateAgent`, `deleteAgent`) now use `runChatSyncOr` with safe defaults (`[]` / `null` / builtin stub / `{ success: false, error: unavailable }`). Mutations that should fail closed (`create` / `update` / `setThreadAgent` / seed) still use `runChatSync`.

main/src/chat/active-streams.ts
samuv: Fixed in `b241f0bf` — `subscribeToStream` / `cancelStream` / other control+read helpers now use `runChatSyncOr` (`null` / `false` / `[]`). `runManagedStream` still uses `runChatPromise` so starting a turn fails loudly if the runtime isn’t ready.

main/src/chat/mcp-tools.ts
samuv: Fixed in `b241f0bf` — `getCachedUiMetadata` → `{}`, `getMcpServerTools` → `null`, `createMcpTools` → empty `{ tools, clients, enabledTools }` via `runChatSyncOr` / `runChatPromiseOr`.

main/src/chat/providers/providers-facade.ts
samuv: Fixed in `b241f0bf` — `fetchProviderModelsHandler` falls back to `null`; `getAllProvidersHandler` falls back to the static catalog from `discoverToolSupportedModels()`.

main/src/chat/thread-integration.ts
samuv: Fixed in `b241f0bf` — `getThreadMessagesForTransport` → `[]`, `getThreadInfo` → empty info object via `runChatPromiseOr` / `runChatSyncOr`. Writes still go through `runChatToResultSync`.

main/src/chat/streaming/stream-registry-service.ts
samuv: Acknowledged — soft-fail consistency is in `b241f0bf`. The stream-registry Effect/`Ref`/Fiber-finalizer redesign is larger; we’ll treat that as a follow-up plan rather than folding it into this soft-fail pass.

samuv: Follow-up done in `3eed5105` (+ readability split in `17b133b5`).

What we changed for this comment:

1. **Persist-on-exit is structural** — `runManagedStream` is now `Effect.scoped` + `forkScoped` + `Fiber.join`, with `Effect.addFinalizer(() => flushPersist(stream))`, so cancel / error / success / scope teardown all flush. Explicit flushes on success/error remain for broadcast ordering; the finalizer is the safety net.

2. **Removed the module-level persist callback** — no more `initStreamRegistryPersistence` / `persistMessagesImpl`. The service closes over `ThreadsService` and persists via `getManagedRuntime()?.runSyncExit(...)`.

3. **Service-owned registry state** — streams map is created in `StreamRegistryService` (with an Effect `Ref`) and published through a small registry holder so `purgeSender` / shutdown can still run after runtime dispose without going through adapters.

4. **Coverage for the hand-test concern** — Vitest now covers cancel mid-stream persistence, shutdown flushing a pending debounce, and MCP partial discovery (one server fails, another succeeds).

Still out of scope for this PR (as noted earlier): rewriting `pumpChunks` / replay maps into pure Effect streams, and putting `WebContents` subscriber sets into a fully Effect concurrent structure.

main/src/chat/agents/registry.ts
kantord: Confirmed fixed in `b241f0bf` — read paths (`listAgents`, `getAgent`, `resolveAgentForThread`, `getThreadAgentId`, `duplicateAgent`, `deleteAgent`) now degrade via `runChatSyncOr` with safe defaults, and mutations (`createCustomAgent`, `updateAgent`, `setThreadAgent`) deliberately still fail loud — the right call for writes. Verified against `origin/chat-effect`. Resolving.

_Disclosure: written by an AI agent._

main/src/chat/active-streams.ts
kantord: Confirmed fixed in `b241f0bf` — `subscribeToStream`/`cancelStream`/other read+control helpers now use `runChatSyncOr` with safe defaults; `runManagedStream` intentionally still fails loud so starting a turn doesn't silently no-op. Verified against `origin/chat-effect`. Resolving.

_Disclosure: written by an AI agent._

main/src/chat/mcp-tools.ts
kantord: Confirmed fixed in `b241f0bf` — `getCachedUiMetadata`/`getMcpServerTools`/`createMcpTools` now degrade to safe defaults instead of throwing/rejecting when the runtime isn't ready. Verified against `origin/chat-effect`. Resolving.

_Disclosure: written by an AI agent._

main/src/chat/providers/providers-facade.ts
kantord: Confirmed fixed in `b241f0bf` — both facades now fall back (`null` / static catalog from `discoverToolSupportedModels()`) instead of throwing. Verified against `origin/chat-effect`. Resolving.

_Disclosure: written by an AI agent._

main/src/chat/thread-integration.ts
kantord: Confirmed fixed in `b241f0bf` — reads degrade to `[]`/empty info via `runChatPromiseOr`/`runChatSyncOr`; writes still go through `runChatToResultSync`, which is appropriate. Verified against `origin/chat-effect`. Resolving.

_Disclosure: written by an AI agent._

main/src/chat/runtime/lifecycle.ts
```diff
@@ -0,0 +1,55 @@
...
+      error instanceof Error ? error.message : 'unknown initialization error'
+    )
+    log.error('[CHAT] Effect runtime initialization failed:', error)
+  }
+}
+
+export async function shutdownChatRuntime(): Promise<void> {
+  // Mark unavailable before dispose so IPC adapters never observe ready=true
+  // with a null ManagedRuntime mid-teardown.
+  markChatRuntimeUnavailable('runtime_disposing')
```
kantord: **Confirmed bug.** `markChatRuntimeUnavailable('runtime_disposing')` runs *before* `shutdownAllActiveStreams()`. `shutdownAllActiveStreams()` calls `flushPersist(stream)` for every active stream, which goes through the `makePersistMessages` closure calling `getManagedRuntime()` — which now returns `null` purely because health just flipped (the underlying `ManagedRuntime` instance is still alive; `disposeChatRuntime()` hasn't run yet). So every shutdown-time flush returns `{success: false}`, and the pending debounced snapshot (up to ~250ms of assistant text/tool state) is never persisted on quit-during-stream — silently defeating the exact guarantee `3eed5105` ("harden stream registry persistence lifecycle") was written to add.

The new regression test ("shutdownAllActiveStreams flushes a pending debounced snapshot") calls `shutdownAllActiveStreams()` directly without going through `shutdownChatRuntime()`/`markChatRuntimeUnavailable()` first, so runtime health stays `'ready'` for the whole test and it passes despite this ordering bug. Suggest flushing pending streams *before* flipping health to unavailable, or making the flush path check `getManagedRuntimeInstance()` (the raw instance) rather than the health-gated `getManagedRuntime()`.

main/src/chat/thread-integration.ts
```diff
@@ -1,128 +1,44 @@
...
-  }
-
-  if (!thread.messages || thread.messages.length === 0) {
-    return []
-  }
-
-  // Return stored messages directly
-  return thread.messages
+export async function getThreadMessagesForTransport(threadId: string) {
+  return runChatPromiseOr(
```
kantord: **Confirmed, real data-loss path.** `getThreadMessagesForTransport` now falls back to `[]` via `runChatPromiseOr` instead of throwing when the runtime is unavailable. The renderer (`chatThreadQueryOptions`) forwards `messages ?? []` with no success/failure discriminator, so a user opening an existing thread during a transient runtime-unavailable window (e.g. right after launch) hydrates as if the thread has zero messages, with no error shown. The write path is confirmed **overwrite, not merge** — `ThreadsRepository.updateMessages` does `{...existing, messages, ...}`, a full replace of the `messages` field. If the user sends a message from this falsely-empty state, the first persistence flush overwrites the thread's real stored history in SQLite with just the new exchange — permanent loss of prior conversation history, not just a transient UI glitch.

main/src/chat/agents/registry.ts
```diff
@@ -121,112 +26,56 @@ export function resolveAgentForThread(
...
-  }
-  deleteAgentFromDb(id)
-  return { success: true }
+  return runChatSyncOr(AgentsService.deleteAgent(id), {
+    success: false,
+    error: CHAT_UNAVAILABLE_USER_MESSAGE,
+  })
 }
 
 export function duplicateAgent(id: string): AgentConfig | null {
```
kantord: **Confirmed, traced end-to-end.** `duplicateAgent` now resolves to `null` (via `runChatSyncOr(..., null)`) instead of throwing when the runtime is unavailable. Both renderer call sites (`agent-detail-page.tsx`, `agents-page.tsx`) only branch on truthy `copy` (`if (copy) { toast.success(...); navigate(...) }`) with no `else`, and the surrounding `try/catch` never fires since nothing rejected. Clicking "Duplicate" while the runtime is down produces a completely silent no-op — no toast, no navigation, no error, nothing visible. Worth either giving this one a typed `{success, error}` result like `deleteAgent` already has, or adding an explicit empty-branch toast on the renderer side.

kantord (commented):
**Disclosure:** This review was fully or partially written by an AI agent, run as an 8-angle structured code review (line-by-line scan, removed-behavior audit, cross-file trace, reuse/simplification/efficiency/altitude/conventions passes) with each candidate independently re-verified against the actual code before being reported. This covers the 3 newest commits (`b241f0bf`, `3eed5105`, `17b133b5`) on top of the state our last review round covered — it is a focused pass, not a guaranteed-complete audit of the whole PR.

8 findings survived verification (6 CONFIRMED, 2 PLAUSIBLE-but-weakened), most severe first — flagged individually below. Two are worth resolving before merge:

1. **Shutdown-flush ordering bug** (`runtime/lifecycle.ts`) — defeats the exact guarantee the persistence-hardening commit added; the new regression test doesn't catch it because it bypasses the real shutdown path.
2. **Thread-hydration data-loss path** (`thread-integration.ts`) — a transient runtime-unavailable window during thread load can lead to real conversation history being overwritten.

The rest are real but lower-impact (a dead `Ref` that doesn't provide the atomicity its own comment claims, a silent no-op on agent duplication, a latent registry bug that's currently unreachable in production, a duplicated/weaker error-classification helper, a replay-state edge case with no visible UI impact today, and a fallback-value inconsistency that's already mitigated by a downstream gate).

main/src/chat/streaming/stream-registry-service.ts
```diff
@@ -0,0 +1,99 @@
...
+export class StreamRegistryService extends Effect.Service<StreamRegistryService>()(
+  'chat/StreamRegistryService',
+  {
+    accessors: true,
+    effect: Effect.gen(function* () {
+      const threads = yield* ThreadsService
+      const streamsMap = new Map<string, ActiveStream>()
+      // Keep an Effect Ref so the service owns the map; imperative helpers
+      // read the same Map instance via setActiveRegistry.
+      yield* Ref.make(streamsMap)
```
kantord: **Confirmed still present.** `yield* Ref.make(streamsMap)` creates a `Ref` whose result is discarded — never bound, never read via `Ref.get`, never written via `Ref.set`/`Ref.update` anywhere in the codebase. The commit message for `3eed5105` states the goal as "Own stream state in StreamRegistryService with Ref-backed map," and the comment right above this line says "Keep an Effect Ref so the service owns the map" — but all real reads/writes still go through the plain mutable `Map` shared via `setActiveRegistry`/`getStreams()`. This was flagged in our last review round and the reply said it was addressed in this commit, but the dead line survived the module split in `17b133b5` unchanged. Suggest either actually threading the `Ref` through `getStreams()`/`setActiveRegistry`, or dropping the `Ref.make` call and the misleading comment — a `Ref` nobody reads is worse than no `Ref`, since it signals a guarantee that doesn't exist.

main/src/chat/streaming/stream-registry-state.ts
```diff
@@ -0,0 +1,59 @@
+import type { WebContents } from 'electron'
+import type {
+  ActiveStream,
+  PersistMessagesSync,
+  StreamRegistryRuntime,
+} from './stream-registry-types'
+
+let activeRegistry: StreamRegistryRuntime | null = null
```
kantord: **Plausible, currently latent.** The active-streams registry `Map` is recreated fresh on every `StreamRegistryService` construction (i.e. on chat-runtime restart) via `setActiveRegistry`, rather than being one process-lifetime-stable object like the pre-split code had. If the runtime is ever restarted mid-process and a user immediately reuses the same `chatId`, a still-finishing old stream's `teardownStream()` calls `getStreams().delete(stream.chatId)` — but `getStreams()` by then resolves to the swapped-in new registry, deleting the *new* stream's live entry and making it invisible to cancel/status IPC calls while it keeps running in the background.

Verified this is not reachable via any production code path today — `initializeChatRuntime()`/`shutdownChatRuntime()` are each called exactly once in the app lifecycle (`app.whenReady()` / `blockQuit()` immediately followed by `app.quit()`); only Vitest's `beforeEach`/`afterEach` hooks restart the runtime repeatedly. So this is a latent landmine, not an active bug — but it would activate the moment any feature (e.g. a settings-triggered runtime restart) reuses this shutdown/reinit pair without addressing the stale-teardown race.

main/src/chat/streaming/stream-registry-persist.ts
```diff
@@ -0,0 +1,159 @@
...
+    flushPersist(stream)
+    try {
+      stream.abortController.abort()
+    } catch {
+      // already aborted
+    }
+  }
+}
+
+export function makePersistMessages(
```
kantord, line 124: **Confirmed duplication, currently latent risk.** `makePersistMessages` hand-rolls an `Effect`-`Exit`-to-`{success,error}` conversion (get runtime, `runSyncExit`, unwrap `Cause.failureOption`, duck-type check for a `userMessage` property) instead of reusing `runChatToResultSync` from `runtime/adapters.ts`, which already implements this exact conversion with proper domain-error detection (`isDomainError` checking `_tag` against `DOMAIN_ERROR_TAGS`). Notably, `threads-storage.ts`'s `updateThreadMessages` wrapper already uses the shared helper correctly for the same underlying operation — this call site bypasses it and reimplements a weaker version by hand. Every error reachable here today is already a properly tagged domain error, so there's no live misclassification yet, but the duplicated logic will drift from the shared helper's classification rules over time. Worth switching to `runChatToResultSync` directly.

main/src/chat/streaming/stream-registry-replay.ts
```diff
@@ -0,0 +1,311 @@
...
+        part.providerExecuted = chunk.providerExecuted ?? part.providerExecuted
+        part.title = chunk.title ?? part.title
+        if (chunk.providerMetadata) {
+          part.callProviderMetadata = chunk.providerMetadata
+        }
+        part.state = 'input-available'
+      }
+      return
+    }
+    case 'tool-input-error': {
```
kantord, line 120: **Confirmed, low current impact.** `recordChunkForReplay`'s `tool-input-error` handling sets `rawInput`/`errorText` and `state: 'output-error'`, but `emitToolReplay` has no branch that ever emits a `tool-input-error`-typed chunk or reads `part.rawInput` — a late-subscribing renderer for a tool call that failed during input parsing gets `tool-input-start` -> `tool-output-error` (skipping the input-error step entirely), losing the model's actual malformed arguments, whereas the live subscriber saw them. Traced the AI SDK's `processUIMessageStream` — it doesn't crash on this sequence, and no component in this repo currently reads `rawInput` from tool parts, so there's no visible UI regression today. But the state genuinely diverges between live and reconnected clients, and would surface the moment anything starts surfacing `rawInput` for debugging/support purposes.

main/src/chat/providers/providers-facade.ts
```diff
@@ -0,0 +1,38 @@
...
+  providerId: string,
+  tempCredential?: string
+): Promise<{ id: string; name: string; models: string[] } | null> {
+  return runChatPromiseOr(
+    ProvidersService.fetchProviderModels(providerId, tempCredential),
+    null
+  )
+}
+
+export async function getAllProvidersHandler(): Promise<
```
kantord, line 31: **Plausible, already mitigated downstream.** `getAllProvidersHandler` falls back to the full static `discoverToolSupportedModels().providers` catalog when the runtime is unavailable, inconsistent with every sibling facade in this same commit choosing an honest empty default (`getPricingMap` -> `{}`, `getThreadInfo` -> empty object, `getMcpServerTools` -> `null`). Verified this doesn't currently cause the severe harm it looks like it would (a healthy-looking model picker during an outage) — `use-available-models.ts`'s `providersWithCredentials` independently filters via `getChatSettings`, which itself soft-fails to an empty `apiKey`, so the picker still ends up correctly empty. Still worth aligning for consistency, since any future consumer of `getAllProviders` that doesn't route through that same credential gate would be misled into thinking providers are live when they aren't.

main/src/chat/runtime/lifecycle.ts
samuv: Fixed in `f6e1a6d8`.

1. `shutdownChatRuntime()` now calls `shutdownAllActiveStreams()` **before** `markChatRuntimeUnavailable('runtime_disposing')`.
2. `makePersistMessages` uses `getManagedRuntimeInstance()` (raw instance) so a flush still works if health is already flipped.
3. Added regression tests that go through `shutdownChatRuntime()` and through `markChatRuntimeUnavailable` → `shutdownAllActiveStreams`.

main/src/chat/thread-integration.ts
samuv: Fixed in `f6e1a6d8` — `getThreadMessagesForTransport` now uses `runChatPromise` (fail closed) instead of soft-failing to `[]`. A runtime-unavailable window rejects the IPC/query so the renderer cannot hydrate as an empty thread and overwrite real SQLite history on the next send. Covered by a unit test that asserts rejection with the unavailable message.

main/src/chat/streaming/stream-registry-service.ts
samuv: Fixed in `bdc04aa1`.

Dropped the unused `Ref.make(streamsMap)` and the misleading ownership comment. The service configures a process-lifetime streams `Map` via `configureStreamRegistry` — no dead `Ref` signaling a guarantee we don’t actually hold.

main/src/chat/agents/registry.ts
samuv: Fixed in `bdc04aa1`.

`duplicateAgent` now returns `{ success, agent?, error? }` (same shape as `deleteAgent`). Unavailable runtime and missing agents both surface an error string; the agents list/detail pages toast on `!result.success` instead of silently no-oping.

main/src/chat/streaming/stream-registry-persist.ts
samuv: Fixed in `bdc04aa1`.

`makePersistMessages` now uses shared `operationResultFromExit` (extracted from `runChatToResultSync`) for Exit → `{ success, error }` conversion with proper domain-error tagging. It still runs against `getManagedRuntimeInstance()` so shutdown flush works while health is `runtime_disposing` (health-gated `runChatToResultSync` would soft-fail there).

main/src/chat/streaming/stream-registry-replay.ts
samuv: Fixed in `bdc04aa1`.

`tool-input-error` now sets `state: 'input-error'` and `emitToolReplay` emits a real `tool-input-error` chunk (including `rawInput`) for late subscribers. Covered by a new active-streams replay unit test.

main/src/chat/streaming/stream-registry-state.ts
samuv: Fixed in `bdc04aa1`.

The streams `Map` is now process-lifetime-stable in `stream-registry-state.ts`. Service (re)init only swaps the persist callback via `configureStreamRegistry`; it no longer replaces the `Map`. Stale `teardownStream()` after a future mid-process restart can’t delete a newly registered same-`chatId` entry.

main/src/chat/providers/providers-facade.ts
samuv: Fixed in `bdc04aa1`.

`getAllProvidersHandler` now soft-fails to `[]` when the runtime is unavailable, aligned with the other unread facades — no more static catalog that can look “live” during an outage.

main/src/chat/streaming/stream-registry-service.ts
samuv: Follow-up on the Ref/ownership piece: in `bdc04aa1` we dropped the unused `Ref.make(streamsMap)` entirely and made the streams `Map` process-lifetime-stable (service re-init only rebinds the persist callback). Persist-on-exit via scoped Fiber + finalizer from `6ff13b00` remains.

Still intentionally out of scope: rewriting `pumpChunks`/replay into pure Effect streams.

main/src/chat/thread-integration.ts
samuv: Also note: after rebasing onto `origin/main`, this fail-closed change lives at `5decc2d9` (same change as the earlier `f6e1a6d8` reply).

main/src/chat/runtime/lifecycle.ts
samuv: Also note: after rebasing onto `origin/main`, this fix lives at `5decc2d9` (same change as the earlier `f6e1a6d8` reply).

kantord (dismissed):

main/src/chat/mcp/mcp-ui-metadata-cache.ts
```diff
@@ -0,0 +1,62 @@
...
+        try: () => readAllMcpAppUiMetadata(),
+        catch: (cause) => cause,
+      }).pipe(
+        Effect.tap((fromDb) => Ref.set(metadata, fromDb)),
+        Effect.tap(() => Ref.set(loaded, true)),
+        Effect.catchAll((error) =>
+          Effect.sync(() => {
+            log.error('[MCP Apps] Failed to load UI metadata from DB:', error)
+          })
+        )
```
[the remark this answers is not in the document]
samuv: Agreed. On read failure we now mark `loaded = true` and leave the in-memory cache empty, so a transient DB error doesn’t retry/spam on every subsequent `get()` during streaming startup.

main/src/chat/streaming/chat-stream-service-impl.ts
```diff
@@ -0,0 +1,336 @@
...
+              cleanupError
+            )
+          }
+
+          throw new Error(toUserFacingProviderMessage(error))
+        }
+      } catch (error) {
+        log.error('[CHAT] Chat stream error:', error)
+        throw error
+      }
```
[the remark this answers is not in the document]
samuv: Good catch — this is the same empty-`.message` class we fixed on the inner path. Outer catch now also throws `new Error(toUserFacingProviderMessage(error))` so setup failures (agent / MCP / builtin tools) surface a stable user string.

kantord (approved):

PR #2493, opened 2026-07-21
fix(renovate): skip cooling-off period for toolhive, fix deprecated fileMatch

danbarr:
## Summary
- Set `minimumReleaseAge: null` for the `stacklok/toolhive` package rule — we control those releases directly, so the 3-day cooling-off period just adds lag between backend releases and the UI picking them up
- Replace deprecated `fileMatch` with `managerFilePatterns` in the `customManagers` block (regex now wrapped in `/…/` delimiters as required)

## Test plan
- [x] `renovate-config-validator renovate.json` passes cleanly with no deprecation warnings

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -12,7 +12,8 @@
       "matchDatasources": ["github-releases"],
       "matchPackageNames": ["stacklok/toolhive"],
       "prPriority": 10,
-      "schedule": []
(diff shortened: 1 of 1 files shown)


samuv (approved):
it makes sense

PR #2551, opened 2026-08-03
feat(mcp): native Modern MCP client with SDK v2 auto-negotiation

samuv:
## Summary

Adds native MCP **2026-07-28 / 2025-11-25** support in the Playground by replacing `@ai-sdk/mcp` with the official TypeScript SDK v2 client. Studio now auto-negotiates Modern with ToolHive v0.41 while still talking to Legacy backends.

### Client & transport
- Centralize MCP connectivity in `connectWorkloadMcpClient()` with `versionNegotiation: { mode: 'auto' }` and streamable-HTTP/SSE transports via the ToolHive proxy
- Remove `@ai-sdk/mcp`; add `@modelcontextprotocol/client` (runtime) and `@modelcontextprotocol/server` (E2E fixtures)
- Drop the direct `@modelcontextprotocol/sdk` dependency — v1 remains only as a **transitive** dep of `@modelcontextprotocol/ext-apps` (MCP Apps app-bridge)

### Playground integration
- New `mcp-ai-tool.ts` adapter maps native MCP tools to AI SDK `dynamicTool` (schema sanitization, metadata, `isError`, abort signals)
- MCP Apps use the same native client for `readResource` / `callTool`
- Session-scoped `{ tools, enabledTools, close }` replaces leaking client arrays; streaming closes MCP + builtin tools in `finally`

### Deferred by design
- **Interactive `input_required`**: `autoFulfill: false` with a clear unsupported error — no pause/resume UX yet

### Tests
- Unit/integration: auto-negotiation config, tool adaptation, cleanup, strict Modern-only fixture (`legacy: 'reject'`)
- E2E: Legacy + Modern Playground paths through ToolHive proxy (SDK v2 test servers accept Legacy startup probes from thv)

## Test plan

- [x] `pnpm run type-check`
- [x] `pnpm run lint`
- [x] `pnpm run knip`
- [x] `pnpm run test:nonInteractive`
- [x] `pnpm run package`
- [x] CI Playground E2E (Legacy + Modern; Docker + Ollama)

### DIFF
--- a/e2e-tests/helpers/test-mcp-server.ts
+++ b/e2e-tests/helpers/test-mcp-server.ts
@@ -33,11 +33,24 @@ export interface TestMcpServer {
 async function readRequestBody(
   req: express.Request
 ): Promise<Uint8Array | undefined> {
-  if (req.method === 'GET' || req.method === 'HEAD') return undefined
-  if (req.body && Buffer.isBuffer(req.body)) return req.body
+  if (
+    req.method === 'GET' ||
+    req.method === 'HEAD' ||
+    req.method === 'DELETE'
+  ) {
+    return undefined
+  }
+
+  if (Buffer.isBuffer(req.body)) {
+    return req.body.length > 0 ? req.body : undefined
+  }
+
+  // express.raw() leaves `{}` for empty bodies — do not forward that as "{}".
   if (req.body && typeof req.body === 'object') {
+    if (Object.keys(req.body).length === 0) return undefined
     return Buffer.from(JSON.stringify(req.body))
   }
+
   return undefined
 }
 
@@ -45,6 +58,10 @@ function createTestMcpHandler(secretCode: string, serverName: string) {
   // ToolHive still probes remote streamable-HTTP endpoints with a Legacy
   // client during workload startup. SDK v2 handlers must accept Legacy there
   // while Studio negotiates Modern through the proxy.
+  //
+  // createMcpHandler invokes this factory once per HTTP request, so concurrent
+  // clients each get an isolated McpServer instance (same isolation intent as
+  // the old per-request McpServer + transport pattern / GHSA-345p-7cg4-v4c7).
   return createMcpHandler(() => {
     const server = new McpServer({
       name: serverName,
@@ -105,7 +122,7 @@ async function startStreamableHttpMcpServer(options: {
         new Request(url, {
           method: req.method,
           headers: req.headers as HeadersInit,
-          body: body ? Buffer.from(body) : undefined,
+          body: body && body.length > 0 ? Buffer.from(body) : undefined,
         })
       )
 
--- a/e2e-tests/helpers/test-mcp-server.ts
+++ b/e2e-tests/helpers/test-mcp-server.ts
@@ -1,6 +1,11 @@
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js'
+import {
+  createMcpHandler,
+  McpServer as ModernMcpServer,
+} from '@modelcontextprotocol/server'
 import express from 'express'
+import { z } from 'zod'
 import type { Server } from 'http'
 
 // Simple word+number format (e.g. "apple42") - hard to guess randomly but simple enough
@@ -28,6 +33,114 @@ export interface TestMcpServer {
   bearerToken: string
   url: string
   stop: () => Promise<void>
+}
+
+async function readRequestBody(
+  req: express.Request
+): Promise<Uint8Array | undefined> {
+  if (req.method === 'GET' || req.method === 'HEAD') return undefined
+  if (req.body && Buffer.isBuffer(req.body)) return req.body
+  if (req.body && typeof req.body === 'object') {
+    return Buffer.from(JSON.stringify(req.body))
+  }
+  return undefined
+}
+
+function createModernMcpHandler(secretCode: string) {
+  return createMcpHandler(
+    () => {
+      const server = new ModernMcpServer({
+        name: 'e2e-modern-test-server',
+        version: '1.0.0',
+      })
+
+      server.registerTool(
+        'get_secret_code',
+        {
+          description: 'Returns a secret code for testing',
+          inputSchema: z.object({}),
+        },
+        async () => ({
+          content: [{ type: 'text', text: secretCode }],
+        })
+      )
+
+      return server
+    },
+    { legacy: 'reject' }
+  )
+}
+
+export async function startModernTestMcpServer(): Promise<TestMcpServer> {
+  const secretCode = generateSimpleCode()
+  const bearerToken = `token-${generateSimpleCode()}`
+  const handler = createModernMcpHandler(secretCode)
+
+  const app = express()
+  app.use(express.raw({ type: '*/*' }))
+
+  app.get('/.well-known/oauth-protected-resource', (_req, res) => {
+    res.status(404).send('Not found')
+  })
+  app.get('/.well-known/oauth-protected-resource/mcp', (_req, res) => {
+    res.status(404).send('Not found')
+  })
+
+  app.get('/', (_req, res) => {
+    res.json({ status: 'ok', protocol: 'modern' })
+  })
+
+  app.all('/mcp', async (req, res) => {
+    if (req.method !== 'OPTIONS') {
+      const authHeader = req.headers.authorization
+      if (authHeader !== `Bearer ${bearerToken}`) {
+        res.status(401).send('Unauthorized')
+        return
+      }
+    }
+
+    try {
+      const host = req.headers.host ?? '127.0.0.1'
+      const url = new URL(req.url ?? '/mcp', `http://${host}`)
+      const body = await readRequestBody(req)
+      const response = await handler.fetch(
+        new Request(url, {
+          method: req.method,
+          headers: req.headers as HeadersInit,
+          body: body ? Buffer.from(body) : undefined,
+        })
+      )
+
+      res.statusCode = response.status
+      response.headers.forEach((value, key) => {
+        res.setHeader(key, value)
+      })
+      const responseBody = Buffer.from(await response.arrayBuffer())
+      res.end(responseBody)
+    } catch (error) {
+      res.statusCode = 500
+      res.end(String(error))
+    }
+  })
+
+  return new Promise((resolve) => {
+    const httpServer: Server = app.listen(0, () => {
+      const address = httpServer.address()
+      const port = typeof address === 'object' && address ? address.port : 0
+
+      resolve({
+        port,
+        secretCode,
+        bearerToken,
+        url: `http://127.0.0.1:${port}/mcp`,
+        stop: async () => {
+          await new Promise<void>((res) => {
+            httpServer.close(() => res())
+          })
+        },
+      })
+    })
+  })
 }
 
 export async function startTestMcpServer(): Promise<TestMcpServer> {
--- a/e2e-tests/helpers/test-mcp-server.ts
+++ b/e2e-tests/helpers/test-mcp-server.ts
@@ -47,28 +47,28 @@ async function readRequestBody(
 }
 
 function createModernMcpHandler(secretCode: string) {
-  return createMcpHandler(
-    () => {
-      const server = new ModernMcpServer({
-        name: 'e2e-modern-test-server',
-        version: '1.0.0',
+  // ToolHive still probes remote streamable-HTTP endpoints with a Legacy
+  // client during workload startup. Rejecting Legacy here prevents the
+  // server card from reaching Running even though Studio negotiates Modern.
+  return createMcpHandler(() => {
+    const server = new ModernMcpServer({
+      name: 'e2e-modern-test-server',
+      version: '1.0.0',
+    })
+
+    server.registerTool(
+      'get_secret_code',
+      {
+        description: 'Returns a secret code for testing',
+        inputSchema: z.object({}),
+      },
+      async () => ({
+        content: [{ type: 'text', text: secretCode }],
       })
+    )
 
-      server.registerTool(
-        'get_secret_code',
-        {
-          description: 'Returns a secret code for testing',
-          inputSchema: z.object({}),
-        },
-        async () => ({
-          content: [{ type: 'text', text: secretCode }],
-        })
-      )
-
-      return server
-    },
-    { legacy: 'reject' }
-  )
+    return server
+  })
 }
 
 export async function startModernTestMcpServer(): Promise<TestMcpServer> {
--- a/e2e-tests/helpers/test-mcp-server.ts
+++ b/e2e-tests/helpers/test-mcp-server.ts
@@ -1,9 +1,4 @@
-import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
-import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js'
-import {
-  createMcpHandler,
-  McpServer as ModernMcpServer,
-} from '@modelcontextprotocol/server'
+import { createMcpHandler, McpServer } from '@modelcontextprotocol/server'
 import express from 'express'
 import { z } from 'zod'
 import type { Server } from 'http'
@@ -46,13 +41,13 @@ async function readRequestBody(
   return undefined
 }
 
-function createModernMcpHandler(secretCode: string) {
+function createTestMcpHandler(secretCode: string, serverName: string) {
   // ToolHive still probes remote streamable-HTTP endpoints with a Legacy
-  // client during workload startup. Rejecting Legacy here prevents the
-  // server card from reaching Running even though Studio negotiates Modern.
+  // client during workload startup. SDK v2 handlers must accept Legacy there
+  // while Studio negotiates Modern through the proxy.
   return createMcpHandler(() => {
-    const server = new ModernMcpServer({
-      name: 'e2e-modern-test-server',
+    const server = new McpServer({
+      name: serverName,
       version: '1.0.0',
     })
 
@@ -71,10 +66,13 @@ function createModernMcpHandler(secretCode: string) {
   })
 }
 
-export async function startModernTestMcpServer(): Promise<TestMcpServer> {
+async function startStreamableHttpMcpServer(options: {
+  serverName: string
+  healthPayload: Record<string, unknown>
+}): Promise<TestMcpServer> {
   const secretCode = generateSimpleCode()
   const bearerToken = `token-${generateSimpleCode()}`
-  const handler = createModernMcpHandler(secretCode)
+  const handler = createTestMcpHandler(secretCode, options.serverName)
 
   const app = express()
   app.use(express.raw({ type: '*/*' }))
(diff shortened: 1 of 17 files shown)


main/src/chat/mcp/mcp-ai-tool.ts
```diff
@@ -0,0 +1,95 @@
+import { dynamicTool, jsonSchema, type JSONSchema7, type JSONValue } from 'ai'
```
kantord: **Praise:** Nice extraction — pulling the MCP↔AI SDK adaptation logic (`normalizeInputSchema`, `mcpCallToolResultToModelOutput`, `createAiMcpTool`) into its own module gives it a single, testable seam instead of scattering it inline in `mcp-service-impl.ts`. Much easier to unit test and reason about in isolation.

main/src/chat/mcp/mcp-ai-tool.ts
```diff
@@ -0,0 +1,95 @@
...
+function mcpCallToolResultToModelOutput({
+  output,
+}: {
+  toolCallId: string
+  input: unknown
+  output: unknown
+}) {
+  const result = output as CallToolResult
+
+  if (!('content' in result) || !Array.isArray(result.content)) {
```
kantord, line 46: **Critique:** This never checks `CallToolResult.isError`, so a normal (non-throwing) MCP error response — e.g. `{ content: [...], isError: true }` — gets routed through the success branch and handed to the model as if the tool call succeeded. Would it make sense to check `result.isError` here and surface it as a tool error (or at least flag it) before falling through to the content mapping below?

main/src/utils/mcp-tools.ts
```diff
@@ -1,178 +1,147 @@
...
-        url: `http://localhost:${workload.port}/sse#${workload.name}`,
-      },
-    }),
-  }
+type ResolvedTransportType = 'streamable-http' | 'sse' | 'unsupported'
 
-  // For stdio transport, ToolHive exposes the server via a proxy (SSE or streamable-http)
-  // Check proxy_mode or URL pattern to determine the actual transport to use
-  let transportType = workload.transport_type as keyof typeof transportConfigs
+function resolveTransportType(workload: CoreWorkload): ResolvedTransportType {
```
kantord, line 37: **Praise:** `resolveTransportType` is a nice simplification over the old mutable-cast dispatch table (`transportConfigs[transportType] || transportConfigs.default`) — a pure function returning a 3-value union is much easier to test and reason about.

main/src/utils/mcp-tools.ts
```diff
@@ -1,178 +1,147 @@
...
-    type?: string
-    url?: string | URL
-  }
-  if (cfgTransport.type === 'http' && cfgTransport.url) {
-    const url =
-      cfgTransport.url instanceof URL
-        ? cfgTransport.url
-        : new URL(cfgTransport.url)
-    return new StreamableHTTPClientTransport(url)
+export function buildMcpClientTransport(workload: CoreWorkload): Transport {
```
kantord, line 55: **Question:** Unlike the old `createTransport`, this now throws for stdio workloads that don't yet have a resolved `proxy_mode`/URL match (the `'unsupported'` branch below), where the old code silently built a transport instead. Naming the failure explicitly is the right instinct, but it does change behavior — is there a caller that needs a friendlier fallback/retry (e.g. while ToolHive is still populating `proxy_mode` right after workload start) instead of a hard throw propagating up as a discovery error?

kantord (commented):
## Review summary

This is a solid migration to the native Modern MCP client with SDK v2 auto-negotiation — worth calling out the wins alongside a few asks.

**What stands out (praise):**
- Clean extraction of the MCP↔AI SDK adaptation logic into `mcp-ai-tool.ts`
- `resolveTransportType` replaces a mutable-cast dispatch table with a pure, testable function
- Moving to the native SDK's `versionNegotiation: { mode: 'auto' }` removes the *cause* of the old `protocolVersion` getter workaround rather than papering over it
- Idempotent `close()` guards in both `mcp-tools.ts` and `mcp-service-impl.ts`
- The new `mcp-client-modern.test.ts` spins up a real HTTP server end-to-end — meaningfully stronger than mocking the SDK

**Worth a look before merge (see inline comments):**
- `isError` on MCP tool results isn't checked, so failed tool calls can look successful to the model
- The dropped `isMcpToolDefinition` validation leaves discovery unguarded against malformed tool definitions
- `buildMcpClientTransport` now throws for stdio workloads without a resolved proxy — flagging to confirm this behavior change is intentional
- Two independent `additionalProperties` normalizations disagree with each other
- A few efficiency/simplification notes (sequential connects, duplicated request-bridging logic, redundant `fetchWorkloads` calls)

None of this blocks the overall direction — flagging for discussion and possible fixups.

_Fully or partially written by an AI agent._

main/src/utils/mcp-tools.ts
```diff
@@ -1,178 +1,147 @@
...
+    {
+      capabilities: options.capabilities,
+      versionNegotiation: { mode: 'auto' },
+      inputRequired: { autoFulfill: false },
+    }
   )
+
+  await client.connect(buildMcpClientTransport(workload))
+
+  let closed = false
```
kantord, line 96: **Praise:** Good attention to detail guarding `close()` with a `closed` flag here — easy to skip, but double-close bugs are exactly the kind of thing that's painful to track down later.

main/src/utils/mcp-tools.ts
```diff
@@ -1,178 +1,147 @@
...
+      serverMcpTools[tool.name] = {
+        description: tool.description,
+        inputSchema: jsonSchema({
+          ...(schema as Record<string, unknown>),
+          type: 'object',
+          properties:
+            'properties' in schema && schema.properties
+              ? schema.properties
+              : {},
+          additionalProperties: false,
```
kantord: **Critique:** This unconditionally forces `additionalProperties: false`, but `mcp-ai-tool.ts`'s `normalizeInputSchema` uses `raw.additionalProperties ?? false` and preserves an explicit `true`. If a server ever declares `additionalProperties: true` on purpose, the tool would show up here (customize-tools UI) with a schema that contradicts what the runtime call path in `mcp-ai-tool.ts` actually honors. Worth aligning the two normalizations, maybe by sharing one helper?

main/src/chat/mcp/mcp-service-impl.ts
```diff
@@ -106,20 +96,19 @@ function shouldSkipAppOnlyTool(
...
-  const clientInfo = { name: 'toolhive-studio-mcp-apps', version: '1.0.0' }
-  const clientOptions = {
-    capabilities: { extensions: MCP_UI_EXTENSION_CAPABILITY },
-  }
-
+async function createRawMcpClientForServer(serverName: string): Promise<{
+  client: Awaited<ReturnType<typeof connectWorkloadMcpClient>>['client']
+  close: () => Promise<void>
+}> {
   const workload = (await fetchWorkloads()).find((w) => w.name === serverName)
```
kantord, line 120: **Question:** `createRawMcpClientForServer` calls `fetchWorkloads()` here independently of `createMcpTools`'s own call to it further down. Since this hits the ToolHive API fresh each time, could this share a short-lived cache (or accept the workload list as a param) to avoid duplicate round-trips when a chat turn fetches a UI resource and also discovers tools?

main/src/chat/mcp/mcp-service-impl.ts
```diff
@@ -334,34 +253,29 @@ export async function createMcpTools(
...
   }
-): Promise<{
-  tools: ToolSet
-  clients: Awaited<ReturnType<typeof createMCPClient>>[]
-  enabledTools: Record<string, string[]>
-}> {
+): Promise<McpToolSession> {
   const mcpTools: ToolSet = {}
-  const mcpClients: Awaited<ReturnType<typeof createMCPClient>>[] = []
+  const closeCallbacks: Array<() => Promise<void>> = []
```
kantord: **Critique:** The cleanup bookkeeping on `closeCallbacks` mixes two strategies — `.pop()` on the success path below (assumes the last-pushed callback is always the one just added) and `indexOf`/`splice` on the error path. A `Map` keyed by server name (or storing `{ serverName, close }` pairs) would make the removal logic explicit rather than relying on array position, and would hold up better if the connect loop below is ever parallelized.

main/src/chat/mcp/mcp-service-impl.ts
```diff
@@ -334,34 +253,29 @@ export async function createMcpTools(
...
-  // and only swap it into the module-level cache (and persist it to SQLite)
-  // once discovery has completed successfully. A transient fetchWorkloads()
-  // or getEnabledMcpTools() failure must not wipe the previously persisted
-  // snapshot — historical MCP App iframes rely on it.
   const nextCachedUiMetadata: Record<string, ToolUiMetadataEntry> = {}
   let discoverySucceeded = false
 
-  /** Registers a validated tool and caches its UI metadata if present. */
-  const registerTool = (
+  let closePromise: Promise<void> | null = null
```
kantord, line 381: **Praise:** Memoizing `closePromise` so repeated `close()` calls return the same in-flight promise is a nice, easy-to-miss correctness detail.

main/src/chat/mcp/mcp-service-impl.ts
```diff
@@ -383,8 +297,6 @@ export async function createMcpTools(
       resolveEnabled(),
     ])
     enabledTools = resolvedEnabledTools
-    // Flag is set here, AFTER the two required calls resolved, so any
-    // rejection above propagates into the catch below without flipping it.
     discoverySucceeded = true
 
     for (const [serverName, toolNames] of Object.entries(enabledTools)) {
```
kantord: **Question:** This connects to each enabled server sequentially (connect + `listTools()` per iteration). Since servers are independent, would `Promise.allSettled` over the servers make sense here to bound total discovery latency by the slowest server rather than the sum of all of them? Could matter for users with several enabled servers, especially any slower remote ones.

main/src/chat/mcp/mcp-service-impl.ts
```diff
@@ -399,31 +311,62 @@ export async function createMcpTools(
...
+
+          const ui = extractToolUiMeta(nativeTool)
+          if (shouldSkipAppOnlyTool(ui)) {
             log.debug(`Skipping app-only tool ${toolName} from ${serverName}`)
-          } else {
-            log.warn(`Tool ${toolName} from ${serverName} failed validation`)
+            continue
           }
+
+          const aiTool = createAiMcpTool({
```
kantord: **Critique:** The old `isMcpToolDefinition` type guard used to filter out malformed tool definitions before they were registered. That's gone now, and the only validation left on the native `listTools()` path is the `typeof schema === 'object'` check inside `normalizeInputSchema`. A misbehaving server returning e.g. a non-string `description` would previously get silently skipped and now flows straight through to `createAiMcpTool`. Intentional, or worth restoring a lighter equivalent check?

e2e-tests/helpers/test-mcp-server.ts
```diff
@@ -30,24 +30,62 @@ export interface TestMcpServer {
   stop: () => Promise<void>
 }
 
-export async function startTestMcpServer(): Promise<TestMcpServer> {
+async function readRequestBody(
```
kantord, line 33: **Critique:** For a request with no body (e.g. a session-terminate `DELETE`), `express.raw()` leaves `req.body` as `{}` (a plain object, not a `Buffer`), so this falls into the `typeof req.body === 'object'` branch and forwards the literal 2-byte string `"{}"` as the body instead of leaving it empty. Could cause spurious e2e failures if a method that expects no body gets rejected by the SDK v2 handler for invalid JSON-RPC. Might be worth special-casing an empty/`{}` object here to return `undefined`.

Side note: this logic is nearly duplicated in `main/src/utils/__tests__/mcp-client-modern.test.ts`'s own `readRequestBody` (raw `IncomingMessage` stream reader instead of `express.raw`+`Buffer` check) — could be worth extracting a shared helper so a fix to one doesn't get missed in the other.

e2e-tests/helpers/test-mcp-server.ts
```diff
@@ -30,24 +30,62 @@ export interface TestMcpServer {
...
+    return Buffer.from(JSON.stringify(req.body))
+  }
+  return undefined
+}
+
+function createTestMcpHandler(secretCode: string, serverName: string) {
+  // ToolHive still probes remote streamable-HTTP endpoints with a Legacy
+  // client during workload startup. SDK v2 handlers must accept Legacy there
+  // while Studio negotiates Modern through the proxy.
+  return createMcpHandler(() => {
```
kantord, line 65: **Question:** The old server code (per the deleted comment) explicitly created a fresh `McpServer` + transport per request specifically to avoid cross-client state leaking (citing GHSA-345p-7cg4-v4c7). Here `createMcpHandler` is given one shared factory closure for all requests — does `createMcpHandler` guarantee per-client isolation internally, or could concurrent e2e clients hitting `/mcp` end up sharing state? Not blocking since this is test-only code, but flagging since the removed comment called out a real security consideration.

main/src/utils/__tests__/mcp-client-modern.test.ts
```diff
@@ -0,0 +1,194 @@
+import { describe, it, expect, afterAll, vi } from 'vitest'
```
kantord, line 1: **Praise:** This is a solid addition — spinning up a real HTTP server and driving `connectWorkloadMcpClient` against it end-to-end (including a UI resource read) is meaningfully stronger than mocking the SDK and would actually catch wire-protocol regressions.

main/src/chat/mcp/mcp-ai-tool.ts
samuv: Great catch, thank you. Fixed in edcd10f3 — `toModelOutput` now checks `isError` and maps to AI SDK `error-text` / `error-json` so the model sees a failed tool call instead of a success. Added a unit test for that path too.

main/src/chat/mcp/mcp-ai-tool.ts
samuv: Thanks! That was exactly the goal of the extraction — keeping the adapter small and easy to test on its own.

main/src/utils/mcp-tools.ts
samuv: Appreciate it — glad that landed cleanly.

main/src/utils/mcp-tools.ts
samuv: Thanks — those double-close bugs are annoying to chase, so good to hear that stood out.

main/src/utils/mcp-tools.ts
samuv: Intentional, and thanks for double-checking. Studio only speaks to ToolHive’s HTTP/SSE proxy, so unresolved stdio without `proxy_mode` is a real misconfig/race rather than something we want to paper over. I left a short comment in the code to make that explicit. If we later see flaky post-start discovery, we can add a targeted retry around proxy readiness, but I would rather keep the failure loud for now.

main/src/utils/mcp-tools.ts
samuv: Agreed. Fixed by sharing `normalizeMcpInputSchema()` between discovery and the AI adapter, so an explicit `additionalProperties: true` is preserved in both places.

main/src/chat/mcp/mcp-service-impl.ts
samuv: Yep, that bookkeeping was getting brittle. Switched to a `Map` keyed by server name so close registration/removal no longer depends on array position.

main/src/chat/mcp/mcp-service-impl.ts
samuv: Good call. I added in-flight dedupe for `fetchWorkloads()` so concurrent callers in the same turn share one request, without a TTL cache that would go stale across sequential calls.

main/src/chat/mcp/mcp-service-impl.ts
samuv: Thanks — happy that detail made it through review.

main/src/chat/mcp/mcp-service-impl.ts
samuv: Fair point. Restored a lightweight `isUsableMcpTool()` guard before AI adaptation, so malformed `description` / `inputSchema` entries get skipped with a warning again.

e2e-tests/helpers/test-mcp-server.ts
samuv: Good question. `createMcpHandler` invokes the factory once per HTTP request, so each client gets an isolated `McpServer` instance — same isolation intent as the old per-request pattern. I added a short comment in the fixture calling that out.

main/src/utils/__tests__/mcp-client-modern.test.ts
samuv: Thanks — that wire-level coverage was the main reason for adding it.

main/src/chat/mcp/mcp-service-impl.ts
samuv: Makes sense. Discovery now runs enabled servers through `Promise.allSettled`, so total latency tracks the slowest server instead of the sum.

e2e-tests/helpers/test-mcp-server.ts
samuv: Nice find. Fixed empty/`{}` bodies (and DELETE) to forward `undefined` instead of `"{}"`. Also tightened the integration test’s stream reader the same way. Left the two helpers separate for now since one is Express/`express.raw` and the other is a raw `IncomingMessage` stream, but happy to extract a shared util later if they keep drifting.

kantord (approved):
## Re-review

All findings from the previous review are addressed in 85950dc8 ("fix(mcp): address review feedback on native Modern client"):

- `isError` is now mapped to AI SDK `error-text`/`error-json` outputs instead of being treated as success
- Tool-shape validation restored as `isUsableMcpTool()` (new `normalize-mcp-input-schema.ts`), applied during discovery
- Schema normalization unified into one shared `normalizeMcpInputSchema()` used by both the discovery UI path and the runtime tool path — the `additionalProperties` disagreement is gone
- The `'unsupported'` transport throw now has a comment explaining it's an intentional fail-loud on a config error, not a silent fallback
- `closeCallbacks` array replaced with a `Map<serverName, close>` — no more position-based bookkeeping
- Per-server discovery now runs in parallel via `Promise.allSettled` instead of a sequential loop
- `fetchWorkloads()` calls are deduped via an in-flight promise cache
- The e2e empty-body `"{}"` bug is fixed in both copies of `readRequestBody`, with explicit DELETE handling
- Added a comment confirming `createMcpHandler` isolates state per request, addressing the earlier isolation question

Verified independently rather than trusting the commit message alone: installed the updated lockfile and ran `type-check` (clean), `lint` (clean), the full unit suite (2623 tests passing), and `knip` (no new findings).

Two very minor, non-blocking nitpicks for a future pass if useful:
- `readRequestBody` in `test-mcp-server.ts` and `mcp-client-modern.test.ts` was fixed in both places but never deduplicated into a shared helper.
- `ServerDiscoveryResult.keepOpen` is always `true` when a result is returned (the `null`-return path already covers "don't keep open"), so the field is effectively dead.

Approving — nice turnaround on the feedback.

_Fully or partially written by an AI agent._

PR #2567, opened 2026-08-04
fix(registry): stop matching reverse-DNS namespace in search

samuv:
## Summary
- Registry search matched the full server name, including reverse-DNS namespaces like `io.github.stacklok/…`. Searching for common terms such as `github` therefore matched (almost) the entire catalog and looked broken.
- Filter on the short name (segment after the last `/`), plus title, description, and tags — the fields users actually see.
- Add regression tests for the namespace false-positive, short-name matching, and tag matching (existing title-search coverage kept).

## Test plan
- [x] Open Registry and type `github` — only GitHub-related servers remain (not the whole catalog)
- [x] Type `fetch` — finds `io.github.stacklok/fetch` via short name
- [x] Type a tag (e.g. a known catalog tag) — matching servers remain
- [x] `pnpm exec vitest run renderer/src/routes/__tests__/registry.test.tsx`

### DIFF
--- a/renderer/src/common/hooks/use-filter-sort.ts
+++ b/renderer/src/common/hooks/use-filter-sort.ts
@@ -2,7 +2,8 @@ import { useState, useMemo } from 'react'
 
 interface UseFilterSortOptions<T> {
   data: T[]
-  filterFields: (item: T) => string[]
+  /** Fields to match against. Receives the lowercased search term for query-aware matching. */
+  filterFields: (item: T, searchTerm: string) => string[]
   sortBy?: (item: T) => string
   sortOrder?: 'asc' | 'desc'
 }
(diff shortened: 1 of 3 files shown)


samuv:
Addressed the follow-up review points in 15683961:

1. **Fully qualified names** — conscious trade-off initially, but agreed it's worth supporting. `filterFields` now receives the search term; when the query contains `/`, we also match the full namespaced name so pasting e.g. `io.github.stacklok/fetch` works, without reintroducing the `github` false-positive for bare terms. Covered by a regression test.

2. **Comment** — softened to "title/description (what users see) and tags (common search terms)".

3. **Tags** — spread as separate fields instead of `join(' ')`.

Left the pre-existing `filterFields` inline-closure / `useMemo` point alone as out of scope.

rdimitrov (approved):

PR #2664, opened 2026-09-02
chore(ci): route Claude workflows through Stacklok LLM gateway

eleftherias:
## Summary

Routes all Claude-powered workflows through the Stacklok LLM gateway and splits the single shared API key into three per-pipeline keys, so spend can be attributed to each group of workflows.

## Changes

| Workflow | Secret | Sites |
|---|---|---|
| `claude.yml` | `STACKLOK_GATEWAY_KEY_CLAUDE_TAG` | 1 |
| `bug-triage-cron.yml` | `STACKLOK_GATEWAY_KEY_BUG_FIX` | 1 |
| `_bug-fix-agent.yml` | `STACKLOK_GATEWAY_KEY_BUG_FIX` | 3 |
| `_security-fix-agent.yml` | `STACKLOK_GATEWAY_KEY_SECURITY_FIX` | 2 |

`ANTHROPIC_BASE_URL: https://llm-gateway.stacklok.dev/anthropic` is set at **job** level rather than per step, so any `claude-code-action` step added to these jobs later routes through the gateway automatically instead of silently going direct and escaping cost tracking.

No caller changes were needed — `on-main.yml`, `security-fix-cron.yml`, and `bug-fix-on-label.yml` all use `secrets: inherit`, which passes the new secret names through to the two reusable workflows.

## Verification

- All 7 `claude-code-action` call sites updated; `secrets.ANTHROPIC_API_KEY` no longer appears anywhere in the repo.
- All four workflow files parse as valid YAML, with the base URL resolving on every job that contains a Claude step.
- The two steps in `_bug-fix-agent.yml` with existing `env:` blocks (`BUG_TEST_FILE`, `BUG_ISSUE_NUMBER`) are unaffected — step-level `env` merges with job-level and the keys don't collide.
- `triage.yml` calls an external reusable workflow but does not pass `secrets: inherit`, so it never received the old key.

The now-unreferenced `ANTHROPIC_API_KEY` repo secret is deliberately left in place as a rollback path.

### DIFF
--- a/.github/workflows/_bug-fix-agent.yml
+++ b/.github/workflows/_bug-fix-agent.yml
@@ -21,6 +21,8 @@ jobs:
       pull-requests: write
       issues: write
       id-token: write
+    env:
+      ANTHROPIC_BASE_URL: https://llm-gateway.stacklok.dev/anthropic
     steps:
       - name: Generate GitHub App token
         id: app-token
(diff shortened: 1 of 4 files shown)


samuv (approved):
👌🏼

kantord (approved):

